PR #15, opened 2015-01-03
Added \di and \dv meta-commands.

j-bennet:


amjith:
Looks good!

Thank you for adding the commands.  :japanese_goblin:

PR #30, opened 2015-01-06
Add entry point

svenstaro:
I'm a distro packager and the generated /usr/bin/pgcli doesn't work for me and I need more control. I need my own entry point so I can execute the module using python -m pgcli.main.


amjith:
:+1: Thank you for the PR. 

:rabbit2:

Issue #24, opened 2015-01-06
Unicode chars in results cause query to return nothing


atombender:
OS X Terminal.app. My environment:

```
# env | grep "LANG|LC" 
LANG=en_US.UTF-8
```

pgcli examples:

```
test> create table unicodechars (t text);
CREATE TABLE
test> insert into unicodechars (t) values ('e');
INSERT 0 1
test> insert into unicodechars (t) values ('é');
INSERT 0 1
test> select * from unicodechars limit 1;
+-----+
| t   |
|-----|
| e   |
+-----+
SELECT 1
test> select * from unicodechars;

test>
```

The last query returns nothing, not even an error. Also:

```
test> select 'e';
+------------+
| ?column?   |
|------------|
| e          |
+------------+
SELECT 1
test> select 'é';

test> select '😁';

test>
```

The prompt is also screwed up by  UTF-8 chars. For example, using the up-arrow to go back leads to:

```
test> selecté';
```

amjith:
That's awful. Sorry about that. 

I'll take a look at it. Thank you for the detailed report.

amjith:
This is fixed by pr #40. 

The fix will be released in next ver.

atombender:
Awesome, thanks!

amjith:
Version 0.10.1 released to PyPI. https://pypi.python.org/pypi/pgcli/0.10.1

PR #40, opened 2015-01-06
Fix UnicodeDecodeError in python2

macobo:
Great tool, thanks for developing!

I ran into some trouble getting it to work with a databases containing unicode data and this pull request aims to mend that. See also #24

I also added some extra debugging information to ensure that the proper error message is shown and always logged for ease of debugging.

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -3,6 +3,7 @@ from __future__ import unicode_literals
 from __future__ import print_function
 
 import os
+import traceback
 import logging
 
 import click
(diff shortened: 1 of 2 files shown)


macobo:
Not quite sure why the build is failing, but I'm guessing it doesn't have to do with this PR.

amjith:
Thank you for taking the time to fix this issue.

I'll test this change manually and merge it in soon.

amjith:
Yay! This totally fixed the issue reported by #24. 

Thanks a lot! :shaved_ice:

PR #42, opened 2015-01-06
Use standard postgres environmental tags

darikg:
so the pgcli command behaves more similarly to the psql command


amjith:
This is going to need more work to add reasonable defaults when these env vars don't exist. 

For example, I don't have any of those env vars defined on my box (I'm on OS X). 

When I launch `pgcli` without any args I get the following error: 

```
$ pgcli                                               
Usage: pgcli [OPTIONS] DATABASE

Error: Missing argument "database".
```

I'll leave this PR open for now. Either you can update it with some defaults or I can take a stab at it later this week.

darikg:
Hmm... If connection arguments are supplied neither from the command line arguments nor from the environmental variables, what possible default action is there? psql apparently resorts to compile-time-specified defaults. I guess I could hardcode in defaults postgres@localhost/postgres ?

amjith:
I think psql falls back to using the current username as the database name. At least that's what it does in my installation.

darikg:
You're right. It's pretty easy to grab the username to replace empty defaults, but it's trickier to match psql behavior with the password:
* if no password is supplied, attempt to connect without one (which is where pgpass comes into play)
    * unless -W is specified, in which case prompt for password immediately
        * so we need to parse URIs immediately to see if the password was supplied or not
* if authentication fails without a password, prompt for it, then attempt to connect again
  * unless -w is specified 

Anyway, I think I got this all working as it should, but I'm used to mercurial and git is pretty bewildering to me right now. I realize now that I submitted this PR as a commit directly on master. Should I close the PR and submit a new one on a feature branch instead?

amjith:
No need to close this PR. If you push your changes to master (in your fork) this PR will be automatically updated to reflect the changes. 

I'll take a look at the diff and test it out. Thanks!

pgcli/main.py
```diff
@@ -27,23 +26,44 @@
...
+from urlparse import urlparse
+from getpass import getuser
+from psycopg2 import OperationalError
+
 _logger = logging.getLogger(__name__)
 
 @click.command()
-@click.option('-h', '--host', default='', help='Host address of the '
-        'postgres database.')
+@click.option('-h', '--host', default='localhost', envvar='PGHOST',
```
amjith: I originally had the default for host set to `localhost` but this was broken on IPv6 machines since localhost points to ::1 and people don't generally run their postgres instance on ::1 but instead using a unix socket. 

So I changed it to an empty string since psycopg2 will do the right thing if you pass in an empty string. So I'd recommend changing this to an empty string.

pgcli/main.py
```diff
@@ -27,23 +26,44 @@
...
+        port = parsed.port
+        host = parsed.hostname
+
+    if prompt_passwd and not passwd:
         passwd = click.prompt('Password', hide_input=True, show_default=False,
                 type=str)
-    else:
-        passwd = ''
+
+    auto_passwd_prompt = not passwd and not never_prompt
```
amjith: I think I understand the logic, but I'd prefer if there was a comment here explaining what this does. Feel free to write a paragraph if you'd like. I don't believe in the idea that a "Well-written code would never need comments". I like to be generous with my commenting. 

A detailed comment today will save me 30mins when I come back to it 2 months later. :)

pgcli/main.py
```diff
@@ -71,7 +91,16 @@ def cli(database, user, password, host, port):
 
     # Connect to the database.
     try:
-        pgexecute = PGExecute(database, user, passwd, host, port)
+        try:
+            pgexecute = PGExecute(database, user, passwd, host, port)
+        except OperationalError as e:
+            if 'no password supplied' in e.message and auto_passwd_prompt:
```
amjith: Once again a comment here would be preferable, to say that a connection retry is attempted since a password was required by wasn't provided.

amjith:
Thanks a lot for making it robust. I really appreciate the time you've put into making this behave just like psql. 

I'll be happy to merge it as soon as you address the comments I've left. 

:palm_tree:

darikg:
thanks for your help!

amjith:
Awesome! Thank you for making the edits. Looks great!

:tangerine:

Issue #32, opened 2015-01-06
More graceful error management


svenstaro:
Currently, on syntax errors in psycopg2 and connection error, the error isn't caught and pgcli fails and exits. Perhaps you should handle that more gracefully.

amjith:
Can you give me a sample error that you're getting?

I tried `pgcli non-existent-db` and I get a `FATAL:  database "abc" does not exist`. 

I presume you're triggering a different scenario. If you can tell me how you're triggering the error that'll be useful as well.

Walms:
Syntax errors in the sql will cause an exit for me. 

For instance if my sql is just 'SELECT' I get :

    amorini> SELECT                        
    Traceback (most recent call last):
      File "/usr/local/lib/python3.3/dist-packages/pgcli/packages/pgspecial.py", line 746, in execute
        command_executor = CASE_SENSITIVE_COMMANDS[command][0]
    KeyError: 'SELECT'

    During handling of the above exception, another exception occurred:

    Traceback (most recent call last):
      File "/usr/local/lib/python3.3/dist-packages/pgcli/pgexecute.py", line 105, in run
        return pgspecial.execute(cur, sql)
      File "/usr/local/lib/python3.3/dist-packages/pgcli/packages/pgspecial.py", line 748, in execute
        command_executor = NON_CASE_SENSITIVE_COMMANDS[command.lower()][0]
    KeyError: 'select'

    During handling of the above exception, another exception occurred:

    Traceback (most recent call last):
      File "/usr/local/lib/python3.3/dist-packages/pgcli/main.py", line 104, in cli
        res = pgexecute.run(document.text)
      File "/usr/local/lib/python3.3/dist-packages/pgcli/pgexecute.py", line 108, in run
        cur.execute(sql)
    psycopg2.ProgrammingError: syntax error at end of input
    LINE 1: SELECT
                  ^


    During handling of the above exception, another exception occurred:

    Traceback (most recent call last):
      File "/usr/local/bin/pgcli", line 9, in <module>
        load_entry_point('pgcli==0.10.0', 'console_scripts', 'pgcli')()
      File "/usr/local/lib/python3.3/dist-packages/click/core.py", line 610, in __call__
        return self.main(*args, **kwargs)
      File "/usr/local/lib/python3.3/dist-packages/click/core.py", line 590, in main
        rv = self.invoke(ctx)
      File "/usr/local/lib/python3.3/dist-packages/click/core.py", line 782, in invoke
[...]

PR #49, opened 2015-01-07
ISSUE #32 More graceful error management

Walms:
Catching specific errors from psycopg2 and using
    the pgerror prop instead of message to avoid app
    crashing.


amjith:
Thank you for the PR. 

:camel:

PR #50, opened 2015-01-07
Fix #45

### LINKED ISSUE #45: Add doc string to the back-slash commands
Add in a doc to display when typing a pg special command that starts with back-slash.

snahor:
Check PEP 352. To put it simple, from 2.6 the message attribute is deprecated and removed in 3.x.


amjith:
I didn't know about the deprecation of e.message attribute.

Thank you for the fix. :racehorse:

Issue #59, opened 2015-01-07
install page


wodabr:
Hi,

there is an error on install page, it should be 

sudo easy_install pip not sudo easy_install install pip

jmagnusson:
I guess you meant the `sudo easy_install install pgcli` error and not `sudo easy_install install pip`, right?

wodabr:
nope :) there's no install command in easy_install

jmagnusson:
Re-read my question :smile:

wodabr:
ok, I ment every occurence of easy_install on http://pgcli.com/install :)

easy_install install pgcli
sudo easy_install install pip

jmagnusson:
Aha then you should submit your issue here https://github.com/amjith/pgcli.com/issues

amjith:
Merged and updated. Thank you for catching that.

PR #61, opened 2015-01-07
README fix for #59

jmagnusson:


amjith:
Thanks for fixing it in Readme. 

I've also fixed the site. Thanks for reporting. 

:+1:

PR #62, opened 2015-01-07
Add support of \dn [PATTERN] command

CyberDem0n:
This command is used in standard psql for listing of schemas


pgcli/packages/pgspecial.py
```diff
@@ -25,6 +25,26 @@ def parse_special_command(sql):
...
+def list_schemas(cur, pattern, verbose):
+    """
+    Returns (rows, headers, status)
+    """
+    sql = """SELECT n.nspname AS "Name",
+    pg_catalog.pg_get_userbyid(n.nspowner) AS "Owner"
+    FROM pg_catalog.pg_namespace n WHERE n.nspname """
+
+    if pattern:
+        sql += "~ '^(" + pattern.replace('*', '.*') + ")$'"
```
amjith: You might have to use [sql_name_pattern](https://github.com/amjith/pgcli/blob/master/pgcli/packages/pgspecial.py#L656) to do this.

amjith:
Apart from the pattern conversion stuff, things look good. If you can update that, I'll test it locally and merge it in. 

Thanks for doing this.

pgcli/packages/pgspecial.py
CyberDem0n: sql_name_pattern is not really doing what we need.

First of all \dn accepts only schema name, so we don't need to extract schema-table pair from it.
Then, any object may contain "." in it's name. For example: "my.schema"."my.table". Moreover, object name could also contain even " (double quote) character.
I think we should reimplement sql_name_pattern by using the same approach as in original processSQLNamePattern function: http://git.postgresql.org/gitweb/?p=postgresql.git;a=blob;f=src/bin/pg_dump/dumputils.c;h=55509866ff5c692288e8b31074a6912020b381ca;hb=master#l956

amjith: I did base the sql_name_pattern() on the original from psql, but I guess I didn't fully cover all the cases and chose to go with a simplified impl. 

Would you like to take a stab at it or should I file an issue for it and get back to it at a later time?

CyberDem0n: Ok, I will take it over.

CyberDem0n: Done with sql_name_patter and related stuff. Please check.

amjith:
:+1: Thank you!!

Looks good. :soccer:

PR #63, opened 2015-01-07
Add expanded output (\x) support

stuartquin:
Hi @amjith
I've added an expanded output flag, it attempts to emulate the psql version as much as possible.
Unfortunately I couldn't find a nice way of doing this with tabulate so added my own `expanded.py`.
Let me know if anything needs modified!

Thanks

### DIFF
--- a/pgcli/packages/expanded.py
+++ b/pgcli/packages/expanded.py
@@ -1,0 +1,41 @@
+def pad(field, total, char=u" "):
+    return field + (char * (total - len(field)))
+
+def get_separator(num, header_len, data_len):
+    total_len = header_len + data_len + 1
+
+    sep = u"-[ RECORD {0} ]".format(unicode(num))
+    if len(sep) < header_len:
+        sep = pad(sep, header_len - 1, u"-") + u"+"
+
+    if len(sep) < total_len:
+        sep = pad(sep, total_len, u"-")
+
+    return sep + u"\n"
+
+def expanded_table(rows, headers):
+    header_len = max([len(x) for x in headers])
+    max_row_len = 0
+    results = []
+
+    padded_headers = [pad(x, header_len) + u" |" for x in headers]
+    header_len += 2
+
+    for row in rows:
+        row_len = max([len(str(x)) for x in row])
+        row_result = ""
+        if row_len > max_row_len:
+            max_row_len = row_len
+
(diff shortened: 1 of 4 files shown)


amjith:
Thank you for adding this, I appreciate it very much. I'll have to spend a bit more time on this. 

In the meantime, it looks like the tests are failing on Python 3. https://travis-ci.org/amjith/pgcli/builds/46178430

pgcli/packages/expanded.py
```diff
@@ -0,0 +1,41 @@
...
+    header_len += 2
+
+    for row in rows:
+        row_len = max([len(str(x)) for x in row])
+        row_result = ""
+        if row_len > max_row_len:
+            max_row_len = row_len
+
+        for item in zip(padded_headers, row):
+            row_result += item[0] + u" " + unicode(item[1]) + u"\n"
```
amjith: You might need to do something like this https://github.com/amjith/pgcli/blob/master/pgcli/packages/tabulate.py#L18 to make the code compatible between Py2 and Py3.

pgcli/packages/expanded.py
```diff
@@ -0,0 +1,41 @@
...
+def expanded_table(rows, headers):
+    header_len = max([len(x) for x in headers])
+    max_row_len = 0
+    results = []
+
+    padded_headers = [pad(x, header_len) + u" |" for x in headers]
+    header_len += 2
+
+    for row in rows:
+        row_len = max([len(str(x)) for x in row])
```
amjith: I'd prefer to do 

```
row_len = len(max(row, key=lambda x: len(x)))
```

This prevents creating a new list for each iteration. It might not be a huge deal for small tables, but if someone does this on a table with thousands of rows, it'll probably result in a lot of GC thrash.

amjith:
I pulled it locally and tested it out. It works beautifully. Thank you!!

If you can address the two comments mentioned above, I'll have no concerns with merging this in. 

:evergreen_tree:

amjith:
Oh and thanks for adding tests!! Very much appreciated.

amjith:
Merging this in, so I can make it work for Python 3. 

Thanks again @stuartquin. 

:cocktail:

amjith:
Merged it via command line to resolve some conflicts. 

Thank you!

Issue #58, opened 2015-01-07
Commands with + at end fails


pythonhacker:
searchdb> \d+
ERROR:  syntax error at or near "\"
LINE 1: \d+
        ^

searchdb> \l+
ERROR:  syntax error at or near "\"
LINE 1: \l+
        ^
Looks like they are not properly parsed (Are you using Python "cmd" based REPL ?)
These are Python syntax errors clearly.

amjith:
This is a bug. I'll take a look. 

I'm not using the 'cmd' module for the REPL. I'm using https://github.com/jonathanslenders/python-prompt-toolkit to build the repl.

amjith:
Fixed in 0.10.3

https://pypi.python.org/pypi/pgcli/0.10.3

PR #69, opened 2015-01-07
Fix syntaxerror for special commands like \d+

macobo:
See #58


amjith:
Great!

Thank you for fixing that. :ice_cream:

PR #71, opened 2015-01-07
Travis - run tox only once per python version

macobo:
Current version of pgcli runs tests too many times in travis, total 12 times instead of 4.

Example: https://travis-ci.org/amjith/pgcli/builds/46215230
Idea for the solution was taken from https://github.com/eventlet/eventlet/blob/master/.travis.yml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,10 @@
 language: python
-python:
-  - "3.4"
-  - "3.3"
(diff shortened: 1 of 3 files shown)


amjith:
:+1:

Thank you! Thanks for adding 3.4 as well. 

:cake:

Issue #68, opened 2015-01-07
Database access in tests


macobo:
It's currently quite hard to make sure that no new regressions are introduced while adding new functionality or fixing bugs.

I would propose adding support for a tests that have access to a running postgres instance and can use it to run commands against. If postgres is not running, such tests would be skipped.

Travis [does have support for postgres](http://docs.travis-ci.com/user/using-postgresql/) and with some experimentation I'm sure turning it on in wouldn't be a problem.

@amjith, how do you feel about this? Would it be something that agrees with your plans for the project?

amjith:
Yes. That is indeed in the grand plan. https://github.com/amjith/pgcli/blob/master/tests/test_plan.wiki

Just haven't gotten enough hours in a day. :)

macobo:
Woot. We'll see, maybe I'll throw something together. :)

Cheers!

PR #75, opened 2015-01-07
Simplistic db tests 

macobo:
Note that this is not meant to be a complete solution, rather an experiment to see what tests that use postgres could look like. Any and all feedback is welcome, @amjith.

See also #68

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -103,15 +103,13 @@ def cli(database, user, password, host, port):
             try:
                 _logger.debug('sql: %r', document.text)
                 res = pgexecute.run(document.text)
+
                 output = []
                 for rows, headers, status in res:
                     _logger.debug("headers: %r", headers)
                     _logger.debug("rows: %r", rows)
-                    if rows:
-                        output.append(tabulate(rows, headers, tablefmt='psql'))
-                    if status:  # Only print the status if it's not None.
-                        output.append(status)
                     _logger.debug("status: %r", status)
+                    output.extend(format_output(rows, headers, status))
                 click.echo_via_pager('\n'.join(output))
             except Exception as e:
                 _logger.error("sql: %r, error: %r", document.text, e)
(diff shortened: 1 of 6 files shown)


pgcli/main.py
```diff
@@ -103,16 +103,13 @@ def cli(database, user, password, host, port):
             try:
                 _logger.debug('sql: %r', document.text)
                 res = pgexecute.run(document.text)
-                output = []
```
macobo: The old code here seems buggy and confusing. Currently the list returned by pgexecute.run always contains only 1 element so why is the for loop even needed?

amjith: That exists because https://github.com/amjith/pgcli/blob/master/pgcli/pgexecute.py#L111 can return results for more than one table. 

eg: This command will return the table details for multiple tables. 

```
django_app> \d django*
```

Can you elaborate why you say it is buggy?

macobo: Ok, obviously missed that case.

It still seems suspicious to me if the `click.echo_via_pager('\n'.join(output))` line is correct, as old data is never flushed from the output but the pager is called for each such list.

Will test tomorrow and reset the code here if needed.

amjith: I haven't had a chance to review your diff yet, but you're right it is a bug. 

The click.echo_via_pager() call should be outside the for loop. Good catch.

amjith:
Overall I like the structure of the test. I'd recommend adding the first test and then refactoring the pieces such as moving the ''.join(output) into a format_output() function. But it's not an absolute requirement. I just prefer small, atomic commits. :)

If you can address the `output` being a list, I'm happy with the new addition.

macobo:
Thanks for the comments - fixed and rebased!

amjith:
I can't get the test to pass. I created a user called `postgres` and tried it. 

Traceback:
```

        if dsn is None:
            if not items:
                raise TypeError('missing dsn and no parameters')
            else:
                dsn = " ".join(["%s=%s" % (k, _param_escape(str(v)))
                    for (k, v) in items])

>       conn = _connect(dsn, connection_factory=connection_factory, async=async)
E       OperationalError: FATAL:  database "_test_db" does not exist

async      = False
connection_factory = None
cursor_factory = None
database   = None
dsn        = 'user=postgres host=localhost dbname=_test_db'
host       = 'localhost'
items      = [('user', 'postgres'), ('host', 'localhost'), ('dbname', '_test_db')]
k          = 'dbname'
kwargs     = {'dbname': '_test_db'}
password   = None
port       = None
user       = 'postgres'
v          = '_test_db'

.tox/py27/lib/python2.7/site-packages/psycopg2/__init__.py:164: OperationalError
================================================================================ 50 passe
```

amjith:
My bad. The `postgres` user I created didn't have privileges to create a db. It is fixed now.

I'll merge this in by hand since it won't merge cleanly. 

Thanks a lot!! :wine_glass:

amjith:
The changes in this PR are now merged in by the subsequent PR #78. 

Thank you! :clap:

PR #78, opened 2015-01-08
Faster startup

### LINKED ISSUE #25: Extremely Slow Startup
It takes it well over a minute to start up on my database - I suspect it is due to the inefficiency of iterating over all the tables and then issuing separate selects to pull in column names. (I am seeing boatloads of queries like SELECT column_name FROM information_schema.columns WHERE\n    table_name ='')

Would likely be faster to do a single select table_name, column_name from information_schema.columns.

macobo:
Do only one query for all columns, instead of len(tables) queries. Should fix #25.

Note that as I wanted to have tests, you should probably merge in #75 first. For an accurate diff, see the last commit!

### DIFF
--- a/pgcli/pgexecute.py
+++ b/pgcli/pgexecute.py
@@ -1,6 +1,7 @@
 import logging
 import psycopg2
 import psycopg2.extensions
+from collections import defaultdict
 from .packages import pgspecial
 
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


pgcli/pgexecute.py
```diff
@@ -123,21 +123,19 @@ def run(self, sql):
                 return [(None, None, cur.statusmessage)]
 
     def tables(self):
```
amjith: Let's rename this to def tables_columns(self) to reflect what it's actually doing.

amjith:
Looks good. I appreciate you taking the initiative to address #25. 

I can't get the tests to pass. I'm having a hard time tracking down why that's the case, since I see the create_database call. I'll dig in a little bit and see if I can find out what's going on.

Thanks!

amjith:
Merged via command line (caveman style). 

Thank you for the performance boost. :apple:

PR #81, opened 2015-01-08
setup.py: license='BSD'

msabramo:
More useful to see on PyPI.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,7 @@ setup(
         author='Amjith Ramanujam',
         author_email='amjith[dot]r[at]gmail.com',
         version=version,
-        license='LICENSE.txt',
+        license='BSD',
         url='http://pgcli.com',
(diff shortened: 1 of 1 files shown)


amjith:
I have vendored in the Tabulate package, which is licensed under MIT. This might change in the future when a new version of Tabulate is released that has the changes I need.

So I wouldn't want to blanket it under BSD without the author's consent.

msabramo:
Gotcha.

amjith:
I don't think we'll do this for a while, since we've vendored in OrderedDict now. I'm going to close this for now.

amjith:
@msabramo It took us a while, but we're finally ready to call pgcli 'BSD' licensed. We're no longer packaging in tabulate library. 

Looks like the PR has grown outdated, but rest assured I'll be updating the setup.py with your change. 

Thank you!

PR #83, opened 2015-01-08
support for table/column names which should be escaped

qwesda:
this patch supports table and column names that have space, special chars, non-ascii word characters and names that conflict with build in function names and keywords.

It also fixes a minor issue where the table columns where not refreshed because of a case-sensitive comparison of the tokens.

For example the following tables/columns can be autocompleted with this patch:
```sql
CREATE TABLE "CamelCase" (
  "has space" text,
  "has %^$&" text,
  "φρόνησις" text,
  "MIN" text,
  "DROP" text,
  "CamelCase" text
);
CREATE TABLE "foo bar" (
  "has space" text,
  "has %^$&" text,
  "φρόνησις" text,
  "MIN" text,
  "DROP" text,
  "CamelCase" text
);
CREATE TABLE "foo *@#^!$" (
  "has space" text,
  "has %^$&" text,
  "φρόνησις" text,
  "MIN" text,
  "DROP" text,
  "CamelCase" text
);
CREATE TABLE "εὐδαιμονία" (
  "has space" text,
  "has %^$&" text,
  "φρόνησις" text,
  "MIN" text,
  "DROP" text
);
CREATE TABLE "MIN" (
  "has space" text,
  "has %^$&" text,
  "φρόνησις" text,
  "MIN" text,
  "DROP" text,
  "CamelCase" text
);
```

### DIFF
--- a/pgcli/pgcompleter.py
+++ b/pgcli/pgcompleter.py
@@ -4,6 +4,7 @@ from collections import defaultdict
 from prompt_toolkit.completion import Completer, Completion
 from .packages.sqlcompletion import suggest_type
 from .packages.parseutils import last_word
+from re import compile
 
 _logger = logging.getLogger(__name__)
 
@@ -45,12 +46,34 @@ class PGCompleter(Completer):
         super(self.__class__, self).__init__()
         self.smart_completion = smart_completion
 
+        self.name_pattern = compile("^[_a-zA-Z][_a-zA-Z0-9\$]*$")
+
+    def extend_escape_name(self, name):
+        if not self.name_pattern.match(name) or name in self.keywords or name in self.functions:
+            name = '"%s"' % name
+
+        return name
+
+    def extend_unescape_name(self, name):
+        if name[0] == '"' and name[-1] == '"':
+            name = name[1:-1]
+
+        return name
+
+    def extend_escaped_names(self, names):
+        for i, name in enumerate(names):
+            names[i] = self.extend_escape_name(name)
+
+        return names
+
     def extend_special_commands(self, special_commands):
         # Special commands are not part of all_completions since they can only
         # be at the beginning of a line.
         self.special_commands.extend(special_commands)
 
     def extend_database_names(self, databases):
+        databases = self.extend_escaped_names(databases)
+
         self.databases.extend(databases)
 
     def extend_keywords(self, additional_keywords):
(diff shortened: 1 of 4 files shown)


amjith:
Thank you for adding the auto-escaping. 

I appreciate it very much. 

I'm going to merge this in, but I think there are still a few things that need to be done to make it more robust. 

eg: Auto-completion isn't working when I start a quote. 

```
SELECT * FROM "IN
```

But I'll create a new issue for that and track it separately.

amjith:
:mega:

pgcli/pgcompleter.py
```diff
@@ -45,12 +46,31 @@ def __init__(self, smart_completion=True):
         super(self.__class__, self).__init__()
         self.smart_completion = smart_completion
 
+        self.name_pattern = compile("^[_a-z][_a-z0-9\$]*$")
+
+    def extend_escape_name(self, name):
```
amjith: This should just be called `escape_name()`.

pgcli/pgcompleter.py
```diff
@@ -45,12 +46,31 @@ def __init__(self, smart_completion=True):
...
+        if not self.name_pattern.match(name) or name in self.keywords or name in self.functions:
+            name = '"%s"' % name
+
+        return name
+
+    def extend_unescape_name(self, name):
```
amjith: This should be called `unescape_name()`.

pgcli/pgcompleter.py
```diff
@@ -45,12 +46,31 @@ def __init__(self, smart_completion=True):
         super(self.__class__, self).__init__()
         self.smart_completion = smart_completion
 
+        self.name_pattern = compile("^[_a-z][_a-z0-9\$]*$")
+
+    def extend_escape_name(self, name):
+        if not self.name_pattern.match(name) or name in self.keywords or name in self.functions:
```
amjith: The lookups inside keywords and functions has to be case-insensitive. Because someone could name their columns `select` all lower case, which should be escaped, but it won't be because the keywords is a collection of uppercased words.

So change this check to: 
if not self.name_pattern.match(name) or name.lower() in self.keywords or name.lower() in self.functions:

amjith: `upper()`/`lower()` -> `potato`/`tomato`. It's all the same. 

But seriously you're right it should be upper(). I was just a sleepy kitten when I left that comment. But I see you've sent a new PR with the fixes. :+1:

PR #87, opened 2015-01-08
README.rst: small tweaks

msabramo:
See at https://github.com/msabramo/pgcli/blob/patch-5/README.rst


amjith:
:+1: 

Thank you for the finer touch. :)

:tractor:

PR #89, opened 2015-01-08
Support CTRL+arrow shortcuts, bump prompt_toolkit to 0.26

macobo:
See #9.

Note that the new version has a slightly different API, which is why this PR is the size it is.

### DIFF
--- a/pgcli/key_bindings.py
+++ b/pgcli/key_bindings.py
@@ -1,14 +1,18 @@
 import logging
+from prompt_toolkit import filters
 from prompt_toolkit.keys import Keys
-from prompt_toolkit.enums import InputMode
+from prompt_toolkit.key_binding.registry import Registry
+from prompt_toolkit.key_binding.bindings.emacs import load_emacs_bindings
 
 _logger = logging.getLogger(__name__)
 
-def pgcli_bindings(registry, cli_ref):
+def pgcli_bindings():
     """
     Custom key bindings for pgcli.
     """
-    line = cli_ref().line
+    registry = Registry()
+    load_emacs_bindings(registry)
+
     handle = registry.add_binding
 
     @handle(Keys.F2)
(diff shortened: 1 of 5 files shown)


pgcli/key_bindings.py
```diff
@@ -25,12 +30,15 @@ def _(event):
         Enable/Disable Multiline Mode.
         """
         _logger.debug('Detected F3 key.')
-        line.always_multiline = not line.always_multiline
+        buf = event.cli.current_buffer
+        buf.always_multiline = not buf.always_multiline
 
-    @handle(Keys.ControlSpace, in_mode=InputMode.INSERT)
+    @handle(Keys.ControlSpace, filter=~filters.HasSelection())
```
macobo: Not 100% confident that this is the same thing, but casual testing seemed to indicate so.

amjith:
Damn! You're fast. I was going to spend some time understanding the changes in API and work on upgrading. But this makes life so much easier. 

I'll test it out when I get home and merge it. Thanks!!

pgcli/key_bindings.py
amjith: I'm finally caught up on reviewing the new api changes in prompt_toolkit (0.26). 

The filters portion is not strictly necessary, but it doesn't hurt. What you're checking to see is whether we're in the selection mode. The only way to get into a selection mode with emacs keybindings is by hitting Ctrl+Space. But since we're overriding ControlSpace itself, I don't think we'll ever be in a selection mode. That's why I think having that arg is redundant.

/cc @jonathanslenders for a sanity check of my explanation here. No hurry, but I'd appreciate your input when you have some time. 

 But it's a non-blocker. So I'll merge this in.

amjith:
Thank you for taking the time to do the necessary changes to make it compatible with the new version of prompt_toolkit. 

Your contributions are appreciated very much. 

:bike:

pgcli/key_bindings.py
jonathanslenders: You are right. It's fine like this.

PR #90, opened 2015-01-08
Update environment variables to those used by psql.

barisbalic:
psql uses `PGHOST`, `PGPORT`, `PGUSER`, and `PGDATABASE` as it's environment variables, this is my humble suggestion that pgcli behaves the same way.

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -30,14 +30,14 @@ from .key_bindings import pgcli_bindings
 _logger = logging.getLogger(__name__)
 
 @click.command()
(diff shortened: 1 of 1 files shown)


barisbalic:
Didn't notice that a similar PR was in play, my bad.  This one doesn't change any other existing behaviour so should work out of the box.  But if you want to discard it, I have no problem with that either.

amjith:
Thank you for the PR. 

The other PR #42 has been merged which covers the same ground. 

Cheers! :bee:

barisbalic:
No problem at all.

Issue #23, opened 2015-01-06
Investigate if it's possible to embed pgcli inside emacs 


amjith:
http://www.emacswiki.org/emacs/SqlMode

amjith:
Closing it since it's a rare use case.

gtrak:
I would like this..

tninja:
+1

bandagriss:
+1

agzam:
pgcli would be awesome as a back-end for ob-sql, so then you could query from org-mode source blocks and have autocompletion, etc.

PR #91, opened 2015-01-09
ipython magic for faster exploratory data analysis

darikg:
I really like @catherinedevlin's [ipython-sql](https://github.com/catherinedevlin/ipython-sql) for getting data from postgresql into python for data munging and visualization, but it's inconvenient writing queries in ipython without autocompletion and syntax highlighting. Since this is obviously pgcli's forte, it seems like the two projects really complement each other. This is my naive attempt to combine them. The general idea is you can drop into a pgcli session from ipython, iterate a query, then quit pgcli to find the query results in your ipython workspace. 

## Requirements
ipython-sql (`pip install ipython-sql`) 

## Usage
From an ipython console (not the notebook though, it'll hang), run `%load_ext pgcli.magic` to load the extension. This also loads the ipython-sql magic if it hasn't been already. Then the command `%pgcli postgresql://yourdbhere` should start a pgcli session. Quit the session with `\q` to return to ipython. The resultset from the last successful query should be available in the local variable `_`. See ipython-sql documentation for controlling the return type.

## Example
```
In [1]: %load_ext pgcli.magic

In [2]: %pgcli postgresql://dg@localhost/temp
temp> select generate_series(5, 10), random()
+-------------------+----------+
|   generate_series |   random |
|-------------------+----------|
|                 5 | 0.546024 |
|                 6 | 0.261947 |
|                 7 | 0.19185  |
|                 8 | 0.186082 |
|                 9 | 0.944421 |
|                10 | 0.963471 |
+-------------------+----------+
SELECT 6
temp> \q
GoodBye!
6 rows affected.
Out[2]:
[(5, 0.385186035186052),
 (6, 0.114940960425884),
 (7, 0.547451738733798),
 (8, 0.333624354097992),
 (9, 0.0919555663131177),
 (10, 0.453994856681675)]

In [3]: _[0]
Out[3]: (5, 0.385186035186052)
```

## TODO
- [x] Refactor most of pgcli.main into a standalone PGCli class. This should dovetail with #23 and other editor plugins
  - [x] Support querying the PGCli object for the last successful query after leaving a cli session
  - [ ] Persist the database connection between multiple %pgcli calls. 
- [ ] Better ipython-sql integration
  - [ ] %sql magic supports working with multiple database connections, and referring to previous connections with a username@hostname alias. %sql without any connection string simply reuses the previous connection. %pgcli should either attempt to emulate this, or even get database connections directly from %sql
  - [ ] Possibly pass a resultset directly to %sql instead of having it re-evaluate the query, which is not only inefficient but also dangerous if the query is mutating
- [ ] Plotting directly in pgcli. I'm envisioning something that lets you call matplotlib and treat the column names from the most recent query as local variables, like
```
> select foo from bar;
> \pylab hist(foo)
```
which should open a figure window with a histogram of the `foo` values

### DIFF
--- a/pgcli/magic.py
+++ b/pgcli/magic.py
@@ -1,0 +1,34 @@
+from pgcli.main import PGCli
+
+def load_ipython_extension(ipython):
+
+    #This is called via the ipython command '%load_ext pgcli.magic'
+
+    #first, load the sql magic if it isn't already loaded
+    if not ipython.find_line_magic('sql'):
+        ipython.run_line_magic('load_ext', 'sql')
+
+    #register our own magic
+    ipython.register_magic_function(pgcli_line_magic, 'line','pgcli')
+
+def pgcli_line_magic(line):
+
+    #for now, assume line is connection string e.g. postgres://localhost
+    uri = line
+
+    pgcli = PGCli()
+    pgcli.connect_uri(uri)
+
+    try:
+        pgcli.run_cli()
+    except SystemExit:
+        pass
+
+    if not pgcli.query_history:
+        return
+
+    q = pgcli.query_history[-1]
+    if q.successful:
+        ipython = get_ipython()
+        return ipython.run_cell_magic('sql', uri, q.query)
+
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -34,6 +34,10 @@ except ImportError:
 from getpass import getuser
 from psycopg2 import OperationalError
 
+from collections import deque, namedtuple
+
+#Query tuples are used for maintaining history
+Query = namedtuple('Query', ['query', 'successful'])
 
 class PGCli(object):
     def __init__(self, force_passwd_prompt=False, never_passwd_prompt=False,
@@ -56,6 +60,8 @@ class PGCli(object):
 
         self.logger = logging.getLogger(__name__)
         self.initialize_logging()
+
+        self.query_history = []
 
     def initialize_logging(self):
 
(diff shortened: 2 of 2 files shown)


amjith:
Holy cow!! This would be awesome!! :D 

I had no idea pgcli could be used this way, but I'm glad someone does.

Would you prefer to have this PR open until all the items in the TODO list are done? 

I'd prefer to merge this as we go along to make sure the branches don't fall out of sync too much. Also it'll encourage us to do this iteratively instead of merging it all in one big chunk. Thoughts? 

BTW, it is confusing to see some of the commits from PR #42 show up in this PR. I'm not sure what to do about it.

darikg:
Sure, merging early makes sense to me. Sorry about that errant commit showing up. Can I rebase on master and push --force or will that blow something up?

amjith:
You should be able to rebase and push force.  I'd say make a copy of the
files first.

darikg:
it worked!

darikg:
I *think* that worked

amjith:
> I think that worked

LOL! 

I'll give it a shot when I get back home. Thanks for doing this. This is quite brilliant.

amjith:
I had to manually merge it in via command line since auto merge had some conflicts. 

It is now up to date and available in master. 

Thank you very much for doing this. This is quite a nice feature that I had not envisioned for pgcli. I'm happy you took the initiative to start this. 

I'm quite excited to see it progress through the rest of the items in the TODO. 

Especially item: 

> "Possibly pass a resultset directly to %sql instead of having it re-evaluate the query, which is not only inefficient but also dangerous if the query is mutating".

Thanks again for the contribution. 

Cheers!
:snail:

Issue #77, opened 2015-01-08
Canceling slow query


karlb:
When using psql, I can cancel a slow query (e.g. when I accidentally forgot a LIMIT clause) by pressing ctrl-c. This would be useful for pgcli, too.

amjith:
Yes. That would be useful. I'll look into this. 

Thank you!

amjith:
Note to dev:

Check out how to how to hook up [on_abort](https://github.com/jonathanslenders/python-prompt-toolkit/blob/master/prompt_toolkit/__init__.py#L233) to cancel a query. 

[Psycopg2.cancel](http://initd.org/psycopg/docs/connection.html#connection.cancel)

amjith:
This is now available via the pr #92 by @macobo. 

It will be available in the next release. 

If you're itching to try it out you can install from master using the following command: 

```
pip install -e git+git@github.com:amjith/pgcli.git#egg=pgcli
```

karlb:
That was quick! Thanks!

amjith:
This is now released in 0.11.0. 

https://pypi.python.org/pypi?:action=display&name=pgcli&version=0.11.0

PR #92, opened 2015-01-09
Support for cancelling queries

macobo:
See #77

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -157,7 +157,8 @@ def cli(database, user, host, port, prompt_passwd, never_prompt):
                     output.extend(format_output(rows, headers, status))
                 click.echo_via_pager('\n'.join(output))
             except KeyboardInterrupt:
-                pgexecute.reconnect()
+                # Restart connection to the database
+                pgexecute.connect()
                 _logger.debug("cancelled query, sql: %r", document.text)
                 click.secho("cancelled query", err=True, fg='red')
             except Exception as e:
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -156,6 +156,10 @@ def cli(database, user, host, port, prompt_passwd, never_prompt):
                     _logger.debug("status: %r", status)
                     output.extend(format_output(rows, headers, status))
                 click.echo_via_pager('\n'.join(output))
+            except KeyboardInterrupt:
+                pgexecute.reconnect()
+                _logger.debug("cancelled query, sql: %r", document.text)
+                click.secho("cancelled query", err=True, fg='red')
             except Exception as e:
                 _logger.error("sql: %r, error: %r", document.text, e)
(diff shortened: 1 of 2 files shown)


macobo:
Note that the test failures are the same as in the current master!

pgcli/pgexecute.py
```diff
@@ -11,6 +12,9 @@
 psycopg2.extensions.register_type(psycopg2.extensions.UNICODE)
 psycopg2.extensions.register_type(psycopg2.extensions.UNICODEARRAY)
 
+# When running a query, make pressing CTRL+C raise a KeyboardInterrupt
+psycopg2.extensions.set_wait_callback(psycopg2.extras.wait_select)
```
amjith: This is interesting. It works as expected but I don't understand how. Reading the documentation says this waits for data to be available, but during the wait time does it automatically capture Ctrl+C and convert it to KeyboardInterrupt? Is this a behavior of `select()` or `wait_select()`?

pgcli/main.py
```diff
@@ -156,6 +156,10 @@ def cli(database, user, host, port, prompt_passwd, never_prompt):
                     _logger.debug("status: %r", status)
                     output.extend(format_output(rows, headers, status))
                 click.echo_via_pager('\n'.join(output))
+            except KeyboardInterrupt:
+                pgexecute.reconnect()
```
amjith: Would it be prudent to call [psycopg2.connection.cancel()](http://initd.org/psycopg/docs/connection.html#connection.cancel)? Just to make sure we're not leaving a long running query on the server and just abandoning a connection.

macobo: Good catch!

My first stab at this showed that the connection was automatically cancelled by `psycopg2.extras.wait_select`.

Looking at the implementation of `wait_select` though, I am not 100% sure WHY it closes the connection. Will test and give you an update.

pgcli/pgexecute.py
macobo: This was explained here: http://initd.org/psycopg/articles/2014/07/20/cancelling-postgresql-statements-python/

Note that wait_select is supposed to carry out the same role as the function defined in that article.

amjith: Yeah I noticed that in the article after I left the comment. Now I'm trying to find the source for wait_select() to check the difference between wait_select() and wait_select_inter().

macobo: Updated the PR with link to the article.

The function can be found [here](https://github.com/psycopg/psycopg2/blob/df9fbc515d6f586e84d7853965f2bc544177ac6a/lib/extras.py#L566). It is the same function as the one from the article, minus handling KeyboardInterrupt.

pgcli/main.py
macobo: Updated to explicitly close the connection.

amjith:
Thanks for doing this. 

:beetle:

PR #94, opened 2015-01-09
Fix travis and build errors under python 3.

macobo:
PR #71 introduced a bug, where the python version that was used by travis did not match the version intended.

Also fix the problems plaguing current master builds failing under python 3, with tests! :)

@amjith

### DIFF
--- a/pgcli/packages/expanded.py
+++ b/pgcli/packages/expanded.py
@@ -1,10 +1,12 @@
+from .tabulate import _text_type
+
 def pad(field, total, char=u" "):
     return field + (char * (total - len(field)))
 
 def get_separator(num, header_len, data_len):
     total_len = header_len + data_len + 1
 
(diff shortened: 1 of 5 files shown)


pgcli/packages/expanded.py
```diff
@@ -1,10 +1,12 @@
+from .tabulate import _text_type
```
amjith: Just an FYI, tabulate is a thrid party package (https://pypi.python.org/pypi/tabulate). I've copied the file physically since I added a new format that wasn't available. 

I've sent a PR to the original project and it's been merged. Waiting for their next release to start depending on it. 

So in the future we might not be able to do `from .tabulate import _text_type`. But this will do for now. :)

amjith:
Thank you for fixing Travis as well as the breaking tests. 

Cheers! :gem:

PR #95, opened 2015-01-09
Renamed escape_* functions and fixed case-insensitive compare

qwesda:
I fixed the remarks for the pull request #83


amjith:
Looks good. Thanks for addressing those. :)

:tada:

Issue #93, opened 2015-01-09
Multiple queries on the same line


macobo:
Only the last query result is output.

Got:
```
$ pgcli -U postgres -h localhost _test_db
_test_db> select 1; select 2
+------------+
|   ?column? |
|------------|
|          2 |
+------------+
SELECT 1
```

Expected: PSQL
```
$ psql -U postgres -h localhost _test_db
psql (9.3.5)
Type "help" for help.

_test_db=# select 1; select 2;
 ?column? 
----------
        1
(1 row)

 ?column? 
----------
        2
(1 row)
```

amjith:
This is a tricky one (or maybe not). 

We'll have to split by semi-colons and run them one at a time using pyscopg2 and print the results.

But the splitting has to take into account that a semi-colon could be inside a quoted string.

OT: I noticed in the pasted output, that the headers have ? around the column name. Is there a problem with displaying unicode chars?

macobo:
I *think* sqlparse has a function for doing the splitting, I'll dig a bit and do a PR.

Re OT: as psql does the same thing I think it's the intended behavior.

amjith:
This is now released in v0.12.0.

https://pypi.python.org/pypi/pgcli/0.12.0

PR #98, opened 2015-01-10
Support having multiple queries on the same line.

macobo:
See #93.

Note that if any of the queries raises an error, nothing besides the error is output. This is different behavior from psql which outputs separately for each query:

```
_test_db=# select 1; invalid syntax; select 2;
 ?column?
----------
        1
(1 row)

ERROR:  syntax error at or near "invalid"
LINE 1: invalid syntax;
        ^
 ?column?
----------
        2
(1 row)
```

Let me know if we want to be fully psql-compatible with this or if this acceptable behavior.

### DIFF
--- a/pgcli/pgexecute.py
+++ b/pgcli/pgexecute.py
@@ -2,6 +2,7 @@ import logging
 import psycopg2
 import psycopg2.extras
 import psycopg2.extensions
+import sqlparse
 from collections import defaultdict
 from .packages import pgspecial
 
@@ -89,11 +90,12 @@ class PGExecute(object):
         are a list of tuples. Each tuple has 3 values (rows, headers, status).
         """
 
+        # Remove spaces and EOL
+        sql = sql.strip()
         if not sql:  # Empty string
             return [(None, None, None)]
 
         # Remove spaces, eol and semi-colons.
-        sql = sql.strip()
         sql = sql.rstrip(';')
 
         # Check if the command is a \c or 'use'. This is a special exception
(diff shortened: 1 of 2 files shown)


pgcli/pgexecute.py
```diff
@@ -106,31 +108,35 @@ def run(self, sql):
             except:
                 _logger.debug('Database name missing.')
                 raise RuntimeError('Database name missing.')
-            self.conn = psycopg2.connect(database=dbname,
-                    user=self.user, password=self.password, host=self.host,
-                    port=self.port)
             self.dbname = dbname
-            self.conn.autocommit = True
+            self.connect()
```
macobo: oof, just realized that this might break everything if connecting to the db failed. Fixing tomorrow.

tests/test_pgexecute.py
```diff
@@ -69,16 +69,43 @@ def test_database_list(executor):
...
-def test_unicode_support_in_output(executor, expanded):
-    if expanded:
+def test_invalid_column_name(executor):
+    with pytest.raises(psycopg2.ProgrammingError) as excinfo:
+        run(executor, 'select invalid command')
+    assert 'column "invalid" does not exist' in str(excinfo.value)
+
+@pytest.yield_fixture(params=[True, False])
+def expanded(request, executor):
+    if request.param:
```
amjith: Took me a while to recognize this is a setup and teardown technique to toggle the expanded output. I'll probably add a comment after merge. :)

amjith:
Thank you for implementing this. 

I'm perfectly fine with quitting early if one of the sql statements fail. I think failing early is a desirable behavior and I'm ok with deviating from `psql` on that. 

Thank you! :zap:

PR #103, opened 2015-01-11
fix logging bug: log statements outside pgcli.main not logged to file

darikg:
file handler was attached to the pgcli.main logger instead of pgcli

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -82,11 +82,12 @@ class PGCli(object):
 
         handler.setFormatter(formatter)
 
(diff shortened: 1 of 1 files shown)


amjith:
Damn, that was quick. :)

I'll try it out. Thanks!

darikg:
hah, i actually figured it out yesterday and then forgot about it until i
saw the issue

amjith:
Thank you! Just tested it and works great. 

Cheers! :bulb:

PR #104, opened 2015-01-11
ipython magic improvements

darikg:
`%pgcli` magic now persists pgcli objects by stashing them in sql-magic connection objects. after first connecting with `%pgcli postgresql://blahblahblah` you can return to that same connection with simply `%pgcli`. if you have multuple pgcli connections open, you can refer to them with an alias like `%pgcli username@dbname`

also, mutating queries aren't rerun in sql magic

### DIFF
--- a/pgcli/magic.py
+++ b/pgcli/magic.py
@@ -1,4 +1,9 @@
-from pgcli.main import PGCli
+from .main import PGCli
+import sql.parse
+import sql.connection
+import logging
+
+_logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


amjith:
The changes look good. Let me know if you can address the comments I left inline. 

Thanks you!

:green_apple:

darikg:
thanks for the careful comments :) i made the changes you suggested, squashed them, and rebased on master

amjith:
Appreciate the quick turn around on addressing the comments. 

Merged. Cheers!

:seedling:

PR #105, opened 2015-01-11
allows pgcli to connect to specified user's database

### LINKED ISSUE #101: Connecting to server only specifying `postgres` user
When I usually connect to a postgres server I just type the following:

    psql -U postgres

This automatically connects me to postgres database. Here, `pgcli` is connecting me to my current user database.
I don't know if this is a Debian specificity, but I coudn't see anything about it in my config files nor on postgresql documentation.
Isn't it a difference of behavior?

fpietka:
... when database name is omitted

Fix #101

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -98,8 +98,11 @@ class PGCli(object):
         # Connect to the database.
 
         if not database:
-            # default to current OS username just like psql
-            database = user = getuser()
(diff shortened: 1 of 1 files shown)


amjith:
Awesome! Tested it locally and it works as expected. Thank you!

:sunflower:

PR #119, opened 2015-01-14
don't try to spilt None - fixes AttributeError: 'NoneType' object has no...

darikg:
I broke the special commands :( sorry!!


amjith:
No worries. Thanks for noticing it and fixing it. :)

Cheers!
:train:

PR #120, opened 2015-01-14
fix bug where show_help command failed in python 3 due to iteritems() de...

darikg:
Not sure if there's more standard way to get a dict iterator in both 2 & 3.

### DIFF
--- a/pgcli/packages/pgspecial.py
+++ b/pgcli/packages/pgspecial.py
@@ -762,7 +762,8 @@ def sql_name_pattern(pattern):
 def show_help(cur, arg, verbose):  # All the parameters are ignored.
     headers = ['Command', 'Description']
     result = []
-    for command, value in sorted(CASE_SENSITIVE_COMMANDS.iteritems()):
+
+    for command, value in sorted(CASE_SENSITIVE_COMMANDS.items()):
         if value[1]:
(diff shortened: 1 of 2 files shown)


pgcli/packages/pgspecial.py
```diff
@@ -762,7 +762,14 @@ def sql_name_pattern(pattern):
...
     result = []
-    for command, value in sorted(CASE_SENSITIVE_COMMANDS.iteritems()):
+
+    try:
+        itr = CASE_SENSITIVE_COMMANDS.iteritems()
+    except AttributeError:
+        #Python 3
+        itr = CASE_SENSITIVE_COMMANDS.items()
+
+    for command, value in sorted(itr):
```
amjith: You can just replace this line with:

```
for command, value in sorted(CASE_SENSITIVE_COMMANDS.items()):
```

In Python 2 it will return a list and Python 3 it'll be an iterator, but both of them are acceptable for `sorted()`.

amjith:
Thank you for fixing that!!

:deciduous_tree: 

I wish there was a way to get notified when a PR gets updated.

darikg:
ah sorry, I just assumed you'd get notified. thanks for merging!

Issue #39, opened 2015-01-06
Auto-complete view names


rgm3:
Much like table names, it would be nice to have the names of views (shown via \dv) included in the auto-complete list.

toupeira:
Also schema names, as in `SELECT * FROM schema_name.table_name`

contradictioned:
:+1:

toupeira:
Thanks, works like a charm.

PR #127, opened 2015-01-18
Make autocomplete schema-aware

### LINKED ISSUE #22: Add auto-completion for tables other than public schema
Right now the tables are only populated from the public schema. Change this to include tables from all schemas.

darikg:
This PR would fix #22, and also I think #39 as a side-effect

Issues:
  * I used pandas DataFrames for working with database metadata -- I found them a lot more flexible than a pure standard lib dict/list based approach. But I'm not sure how you'd feel about adding pandas as a requirement.
  * We don't have any tests for "DoubleQuoted" identifiers -- I'm not confident I didn't break that.

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -332,11 +332,9 @@ def quit_command(sql):
             or sql.strip() == ':q')
 
 def refresh_completions(pgexecute, completer):
-    tables, columns = pgexecute.tables()
-    completer.extend_table_names(tables)
-    for table in tables:
-        table = table[1:-1] if table[0] == '"' and table[-1] == '"' else table
-        completer.extend_column_names(table, columns[table])
+    tables, columns = pgexecute.get_metadata()
+    completer.extend_tables(tables)
+    completer.extend_columns(columns)
     completer.extend_database_names(pgexecute.databases())
 
 if __name__ == "__main__":
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -214,6 +214,12 @@ class PGCli(object):
                         end = time()
                         total += end - start
                         mutating = mutating or is_mutating(status)
+
+                        if need_search_path_refresh(document.text, status):
+                            logger.debug('Refreshing search path')
+                            completer.set_search_path(pgexecute.search_path())
+                            logger.debug('Search path: %r', completer.search_path)
+
                 except KeyboardInterrupt:
                     # Restart connection to the database
                     pgexecute.connect()
@@ -329,6 +335,22 @@ def need_completion_refresh(sql):
     try:
         first_token = sql.split()[0]
         return first_token.lower() in ('alter', 'create', 'use', '\c', 'drop')
+    except Exception:
+        return False
+
+def need_search_path_refresh(sql, status):
+    # note that sql may be a multi-command query, but status belongs to an
+    # individual query, since pgexecute handles splitting up multi-commands
+    try:
+        status = status.split()[0]
+        if status.lower() == 'set':
+            # Since sql could be a multi-line query, it's hard to robustly
+            # pick out the variable name that's been set. Err on the side of
+            # false positives here, since the worst case is we refresh the
+            # search path when it's not necessary
+            return 'search_path' in sql.lower()
+        else:
+            return False
     except Exception:
         return False
 
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -267,10 +267,10 @@ class PGCli(object):
 
         pgexecute = self.pgexecute
 
-        schemata, tables, columns = pgexecute.get_metadata()
-        completer.extend_schemata(schemata)
-        completer.extend_tables(tables)
-        completer.extend_columns(columns)
+        completer.set_search_path(pgexecute.search_path())
+        completer.extend_schemata(pgexecute.schemata())
+        completer.extend_tables(pgexecute.tables())
+        completer.extend_columns(pgexecute.columns())
         completer.extend_database_names(pgexecute.databases())
 
     def get_completions(self, text, cursor_positition):
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -352,13 +352,5 @@ def quit_command(sql):
             or sql.strip() == '\q'
             or sql.strip() == ':q')
 
-def refresh_completions(pgexecute, completer):
-    tables, columns = pgexecute.tables()
-    completer.extend_table_names(tables)
-    for table in tables:
-        table = table[1:-1] if table[0] == '"' and table[-1] == '"' else table
-        completer.extend_column_names(table, columns[table])
-    completer.extend_database_names(pgexecute.databases())
-
 if __name__ == "__main__":
     cli()
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -332,7 +332,8 @@ def quit_command(sql):
             or sql.strip() == ':q')
 
 def refresh_completions(pgexecute, completer):
-    tables, columns = pgexecute.get_metadata()
+    schemata, tables, columns = pgexecute.get_metadata()
+    completer.extend_schemata(schemata)
     completer.extend_tables(tables)
     completer.extend_columns(columns)
     completer.extend_database_names(pgexecute.databases())
--- a/pgcli/packages/sqlcompletion.py
+++ b/pgcli/packages/sqlcompletion.py
@@ -46,6 +46,12 @@ def suggest_type(full_text, text_before_cursor):
     return suggest_based_on_last_token(last_token, text_before_cursor, full_text)
 
 def suggest_based_on_last_token(token, text_before_cursor, full_text):
+    """Returns a list of suggestion dicts
+
+        A suggestion dict is a dict with a mandatory field "type", and optional
+        additional fields supplying additional scope information.
+    """
+
     if isinstance(token, string_types):
         token_v = token
     else:
@@ -66,28 +72,67 @@ def suggest_based_on_last_token(token, text_before_cursor, full_text):
             # If the lparen is preceeded by a space chances are we're about to
             # do a sub-select.
             if last_word(text_before_cursor, 'all_punctuations').startswith('('):
-                return 'keywords', []
-        return 'columns', extract_tables(full_text)
+                return [{'type': 'keyword'}]
+
+        return [{'type': 'column', 'tables': extract_tables(full_text)}]
+
     if token_v.lower() in ('set', 'by', 'distinct'):
-        return 'columns', extract_tables(full_text)
+        return [{'type': 'column', 'tables': extract_tables(full_text)}]
     elif token_v.lower() in ('select', 'where', 'having'):
-        return 'columns-and-functions', extract_tables(full_text)
+        return [{'type': 'column', 'tables': extract_tables(full_text)},
+                {'type': 'function'}]
     elif token_v.lower() in ('from', 'update', 'into', 'describe', 'join', 'table'):
-        return 'tables', []
+        return [{'type': 'schema'}, {'type': 'table', 'schema': []}]
     elif token_v.lower() == 'on':
-        tables = extract_tables(full_text, include_alias=True)
-        return 'tables-or-aliases', tables.keys()
+        tables = extract_tables(full_text)
+
+        # Use table alias if there is one, otherwise the table name
+        alias = tables['alias'].where(tables['alias'].notnull(), tables['table'])
+
+        return [{'type': 'alias', 'aliases': list(alias)}]
+
     elif token_v in ('d',):  # \d
-        return 'tables', []
+        return [{'type': 'schema'}, {'type': 'table', 'schema': []}]
     elif token_v.lower() in ('c', 'use'):  # \c
-        return 'databases', []
+        return [{'type': 'database'}]
     elif token_v.endswith(','):
         prev_keyword = find_prev_keyword(text_before_cursor)
         if prev_keyword:
-            return suggest_based_on_last_token(prev_keyword, text_before_cursor, full_text)
+            return suggest_based_on_last_token(
+                prev_keyword, text_before_cursor, full_text)
     elif token_v.endswith('.'):
-        current_alias = last_word(token_v[:-1])
-        tables = extract_tables(full_text, include_alias=True)
-        return 'columns', [tables.get(current_alias) or current_alias]
 
-    return 'keywords', []
+        suggestions = []
+
+        identifier = last_word(token_v[:-1], 'all_punctuations')
+
+        # TABLE.<suggestion> or SCHEMA.TABLE.<suggestion>
+        tables = extract_tables(full_text)
+        tables = get_matching_tables(tables, identifier)
+        suggestions.append({'type': 'column',
+                            'tables': tables[['schema', 'table', 'alias']]})
+
+        # SCHEMA.<suggestion>
+        suggestions.append({'type': 'table', 'schema': identifier})
+
+        return suggestions
+
+    return [{'type': 'keyword'}]
+
+def get_matching_tables(tables, identifier):
+    """
+    :param tables: DataFrame with columns [schema, table, alias]
+    :param identifier: a table name, table alias, or fully qualified schema.tablename
+    :return: a row or rows from tables that match the indentifier
+    """
+    tables['full_name'] = tables.apply(qualify_table_name, axis=1)
+
+    #match a table to an identifier if the identifer is equal to any one of the
+    #table name, table alias, or schema-qualified table name
+    matches = tables[['table', 'alias', 'full_name']] == identifier
+    is_match = matches.any(axis=1)
+
+    return tables[is_match]
+
+def qualify_table_name(row):
+    return row['schema'] + '.' + row['table'] if row['schema'] else row['table']
--- a/pgcli/packages/sqlcompletion.py
+++ b/pgcli/packages/sqlcompletion.py
@@ -78,12 +78,12 @@ def suggest_based_on_last_token(token, text_before_cursor, full_text):
     elif token_v.lower() in ('from', 'update', 'into', 'describe', 'join', 'table'):
         return [{'type': 'schema'}, {'type': 'table', 'schema': []}]
     elif token_v.lower() == 'on':
-        tables = extract_tables(full_text)
+        tables = extract_tables(full_text)  # [(schema, table, alias), ...]
 
         # Use table alias if there is one, otherwise the table name
-        alias = tables['alias'].where(tables['alias'].notnull(), tables['table'])
+        alias = [t[2] or t[1] for t in tables]
 
-        return [{'type': 'alias', 'aliases': list(alias)}]
+        return [{'type': 'alias', 'aliases': alias}]
 
     elif token_v in ('d',):  # \d
         return [{'type': 'schema'}, {'type': 'table', 'schema': []}]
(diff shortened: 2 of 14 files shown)


amjith:
Taking a look now. It might take me a while to fully review. 

I'm a bit hesitant to add Pandas as a requirement since it's not quite straight-forward to install. But if it really does make it easy to do certain things, then I'm open to adding it. 

I've been meaning to learn Pandas for a while now, this will be a great opportunity. :)

tests/test_parseutils.py
```diff
@@ -3,52 +3,114 @@
...
+            'table': ['tabl1', 'tabl2'],
+            'alias': ['t1', 't2']}
 
 def test_simple_insert_single_table():
     tables = extract_tables('insert into abc (id, name) values (1, "def")')
-    assert tables == ['abc']
+    assert tables.to_dict('list') == \
+           {'schema': [None], 'table': ['abc'], 'alias': ['abc']}
+
+@pytest.mark.xfail
```
amjith: This needs an `import pytest`. That's why the Travis tests are failing.

darikg: oh that makes sense. Pycharm was apparently just autoimporting pytest for me. I'll fix it in a minute

darikg:
I'm really sorry about that. Right now I only know how to squash commits in
git, I haven't figured out how to split them apart yet, haha. Mostly I
wanted to find out what you thought about the overall approach before I
spent too much more time on it. What I should've had as the first commit
was refactoring packages.sqlcompletion.suggest_type to return an array of
suggestion objects (right now they're just dicts, but they could be
namedtuples or actual custom classes or whatever) instead of the type as a
string like 'columns-and-functions'  or whatever, since that doesn't scale
too well as the number of different possibilities expands.

I'm totally cool with not merging right away, since I think this would
provide the foundation for more complicated autocompletion (CTEs, user
defined functions (especially set-returning functions, which can act as
tables), arbitrary-depth-nested composite types...) so we might as well get
it right.

I also totally understand your hesitations regarding pandas (even if we
don't end up using it, you should still learn it, it's awesome!) it's
totally feasible to rewrite in a more straightforward way -- it just saved
me some time in development.

amjith:
No worries!!

I like the refactoring done to suggest_type method.

I'm pulling down your repo to play around a little. I might send PRs to
your repo to make some corrections. Then we can finally merge it back into
mainline.

pgcli/main.py
```diff
@@ -332,11 +332,10 @@ def quit_command(sql):
             or sql.strip() == ':q')
 
 def refresh_completions(pgexecute, completer):
-    tables, columns = pgexecute.tables()
-    completer.extend_table_names(tables)
-    for table in tables:
-        table = table[1:-1] if table[0] == '"' and table[-1] == '"' else table
-        completer.extend_column_names(table, columns[table])
+    schemata, tables, columns = pgexecute.get_metadata()
+    completer.extend_schemata(schemata)
```
amjith: extend_schemata() is missing in PGCompleter() class.

darikg: dang it

amjith:
@darikg You're right about the auto-escaped tables and columns being broken. 

I've added new tests to the master branch to properly test the auto-escaped identifiers.

I also pulled in your PR into a branch in the mainline. https://github.com/amjith/pgcli/tree/darikg/schema-completion and merged it in with the master. 

The tests are failing. 

My attempts to fix them have been unsuccessful because of my inexperience with pandas. But I've narrowed it down to this line. https://github.com/amjith/pgcli/blob/darikg/schema-completion/pgcli/pgcompleter.py#L94

I think your intention is to apply the escaping to all the relevant values in the DataFrame, but it doesn't apply the transformation in place, it actually applies the transformation and returns the transformed copy. 

So I tried changing it to this: 

```  
      # data is a DataFrame with columns [schema, table, is_visible]
        tmp = data[['schema', 'table']].apply(self.escaped_names)
        self.tables = self.tables.append(tmp)
```

But that didn't work because we want `tmp` to have the `is_visible` col but that's not present in the transformed `tmp` var. 

I'm also confused by how the `apply()` method works. Because I would have expected it to work with `apply(self.escape_name)` but that gives an error. 

I might start updating that branch https://github.com/amjith/pgcli/tree/darikg/schema-completion with a non-pandas version of the change you've made, if you don't mind.

darikg:
Thanks for the auto-escaped identifiers tests, they made it a lot easier to find all my mistakes (there were a lot). 

You're right about DataFrame.apply operating in-place, so it should be
```data[['schema', 'table']] = data[['schema', 'table']].apply(self.escaped_names)```

Looks like the tests are passing in python 2, but failing in 3 because of the dict-sorting in the test files. I'll have to figure out a better way to do that. I should probably write a lot more schema-autocompletion tests in general too -- I assume there's a whole bunch of weird edge cases

**Comment by amjith** [comment]
@darikg Thanks for fixing that. I'll take a look at the failing Py3 tests. 

I'll dig into this a little more tonight. It's been fun seeing how simplified the code can be when pandas is used. But I'm still not sure why `apply()` didn't take `self.escape_name` as an arg. 

Also if you have any docs for quick start on Pandas, I'd appreciate some pointers. :)

**Comment by darikg** [comment]
In a DataFrame, each column is a Series object, and DataFrame.apply(func)
calls func on each Series in turn. Series are duck-typed like lists, so you
need basically func to accept a list-like object. You can do
DataFrame.applymap() instead, which applies func on individual elements in
the frame.

The pandas docs are actually really good. Also the guy who wrote pandas has
a book (python for data analysis) with a bunch of chapters on pandas.

**Comment by amjith** [comment]
I tried it out last night. It works well. 

I did run into one issue. When you type:

```
SELECT * FROM django_migration WHERE <tab>
```

The column names are listed twice. I was debugging it for a while and I think it's coming from `populate_scoped_cols()` when you do an inner join merge, I haven't confirmed it yet. 

But the weird thing is, I can't reproduce it in the unit-tests. At first I thought I couldn't reproduce it in tests because the tests call `set()` on the list of suggestions which removes duplicates, but even if I replaced the set() with list() it still doesn't trigger that behavior in tests. Which makes it harder to debug. 

I've been adding a ton of _logger.debug() statements to debug this  (caveman-style) while running the repl.

**Comment by darikg** [comment]
Wow, nice find. I was able to replicate something like that by creating two identical tables, one in the public schema and one inside another (non-visible) schema. Then get_scoped_cols was ultimately combining rows from unqualed_tables and columns dataframes and on just the table column, instead of schema and table. So two different tables of the same name ended up contributing columns to the suggestion list. I just pushed a commit that should fix that. Is this the case in your database by any chance?

I was going to add a test for this issue as well, but I was thinking, does it make sense to make the mock database structure in test_smart_completion more and more complicated, or should it be split up into a simple mock database with only a public schema, and then a separate set of tests for a more complicated db design?

**Comment by amjith** [comment]
Nope. That wasn't it. 

It was actually this line: 

```
return list(qualed['column']) + list(unqualed['column'])
```

Screenshot: 

screen shot 2015-01-20 at 10 34 29 pm

It's picking the columns from a qualed table `public.django_migrations` and an unqualed table `django_migrations` and appending them together and returning it. 

But both of them are the same table so the columns appear twice. 

I think the simple solution is make those lists into sets an do a union operation. That will get rid of duplicates, but that's really treating the symptoms. The root cause fix would be excluding the qual table name from the unqualed table name list. 

BTW, I'm totally fine with separating the more complicated tests to a different file.

**Comment by darikg** [comment]
Hmm... if the table in the query isn't qualifed, then the corresponding value of `scoped_tbls['schema']` should be null, and `django_migrations` shouldn't show up in `qualed` (because of the missing inner join condition on `schema`. Right?

Anyway, I was thinking about how to do this without pandas, and I think I would define namedtuples `Table('schema_name', 'table_name', 'is_visible')` and `Column('schema_name', 'table_name', 'column_name')' and store the database metadata as lists of namedtuples instead of dataframes. 

I think I should also have `suggest_type` return tuples of namedtuple suggestion types, instead of my current dict-based approach. Doing everything with tuples would make everything immutable, so I could get rid of my stupid assert_equals kludge in the tests, and the tests would then start passing again in python 3.

What do you think?

**Comment by amjith** [comment]
> Hmm... if the table in the query isn't qualifed, then the corresponding value of scoped_tbls['schema'] should be null, and django_migrations shouldn't show up in qualed (because of the missing inner join condition on schema. Right?

It doesn't happen on an unqualed table name, only on qualed table name and the schema has to be 'public' and the table needs to be visible. 

So try this: 

```
SELECT * FROM public.users WHERE 
```

The table 'users' has to be visible and it should be qualed with a schema name. 

I have some thoughts on the panda-less implementation. I'll write up my thoughts later tonight.

**Comment by amjith** [comment]
About the data-structures to use in place of Panda.

I don't think using a namedtuple for Tables and Columns is going to work for a couple of reasons. 

Having `Table(schema, table_name, is_visible)` will result in a lot of duplication of data. For eg: All the tables in `public` schema will have ('public', 'table1', True), ('public', 'table2', True), ('public', 'table3', True) etc. Where the schema and is_visible columns are redundant. When you have to lookup all the tables in the 'custom' schema, you have to iterate over a list of namedtuples to filter them by the schema column that matches 'custom'. 

So here's my proposal. 

The `parseutils.extract_tables()` should return a tuple (or a namedtuple) with (schema, table, alias). 

* 'schemata' should be a list of schemas.
* 'schema_tables' should be a dictionary with schema as key and a list of table names in the schema. eg: {'public' : ['table1', 'table2'], 'custom' :['table1', 'table2']}
* 'visible_tables' should be a `set()` of all the visible table names from schemas. This is used for unqaled completion. If we don't have this, then we would have to iterate through the schema_tables dictionary to gather all the values and build a set of all the table name on the fly (which is inefficient). 
* 'columns' should be a dictionary with keys as 'tables' and values as a list of column names that are in that table. 

What do you think?

**Comment by darikg** [comment]
Good call on setting up the metadata storage for finding stuff more efficiently.

> 'columns' should be a dictionary with keys as 'tables' and values as a list of column names that are in that table.

What about when two tables in different schemas have the same name? I guess we could have two (redundant) dictionaries, one indexed by 'table' for unqualed references, and another indexed by the tuple ('schema', 'table') for qualified references?

**Comment by amjith** [comment]
Yes. There are two structures with tables in them. One is `schema_tables` for qualified references and the other is `visible_tables` which is just a set() of all tables from all the schemas without a qualified reference. 

So when someone types `SELECT * FROM custom.u`, we can look up all the tables in the schema named `custom` from the `schema_tables` dictionary and find the tables that start with 'u'. Or if the user types `SELECT * FROM u` without a schema name, we'll look up all the tables that start with 'u' in the `visible_tables` + all the `schemas`. 

I'm not sure if I'm making this clear.

If you're constrained for time or have other priorities, let me know and I can build on top of the work you've done. I have touched it so far (apart from testing it locally) because I didn't want to duplicate any work you might have done. 

BTW, thank you for talking through the design with me. It's been very helpful. :)

**Comment by darikg** [comment]
I'm thinking of a case like this:

```
> CREATE SCHEMA s1;
> CREATE TABLE s1.tbl (a INT, b INT);
> CREATE SCHEMA s2;
> CREATE TABLE s2.tbl (c INT, d INT);
> SELECT <autocomplete> FROM s1.tbl;
```

Now, you need to look up the columns so you can suggest (a, b) and not (c, d). There's a ```columns``` dictionary, but how is it indexed? It can't just be ```columns['tbl']```, but needs to be something like ```columns[('s1', 'tbl')]```... right?

**Comment by amjith** [comment]
Oh good catch. 

Here's my current thinking. Instead of having multiple data-structures for schemas, tables and columns we have one data-structure. Let's call it `metadata`. 

`metadata` is a nested dictionary like this:

```
{'schema': {'table': [columns,....] } }
```

We always maintain a state called default schema (initially set to 'public'). 

Now 

```
> SELECT * FROM <tab>            # will autocomplete all the tables from the default schema.
> SELECT * FROM custom.          # will autocomplete all tables from 'custom' schema and columns from 'public.custom' table if it exists. 
> SELECT <tab> FROM table        # will autocomplete all columns from 'table' in the default schema. 
> SELECT <tab> FROM custom.table # all columns from 'custom.table'
> SELECT * FROM blah WHERE <tab> # all columns from a table 'blah' in the default schema.
```

The way to change the default schema is to use `set search_path TO custom`. That will make 'custom' to be the new default. 

What do you think?

**Comment by darikg** [comment]
That would definitely work for the way **I** use postgresql :). The question is how many other people's weird usage patterns do you want to support?

1. search_path can be a list of schemas, not just one.
  * ```set search_path to custom, public, other_schema```
2. Given (1), tables of the same name in different schemas on the search_path can shadow each other.
  * E.g. given tables s1.tbl and s2.tbl, ```SET search_path TO s1, s2; SELECT * FROM tbl``` will select all from s1.tbl, not s2.tbl. (Or maybe the other way around? I can't remember)
2. The default search_path can be set at the database or user level, and doesn't have to default to ```public```. 
  * Actually, at least on my local server, it defaults to "$user", public

I think these issues won't be a problem as long as we grab the search path as a list of schema names while grabbing the rest of the metadata, and search for unqualified tables by iterating over the search_path one schema at a time (so correct shadowing behavior is exhibited.)

One extra thing to keep in mind for the future is eventually I'd like to be able to support user-defined functions, especially table-returning functions, and even more especially, somehow hinting function argument names. I think that's a bunch of extra metadata that would probably be stored in separately from the table and column metadata, but I thought it was at least worth bringing up.

**Comment by amjith** [comment]
I don't use schemas very much. So I'm glad you're thinking of all these possibilities. 

If I understand your correctly, if we make the default schema as a list, then we can stick with the nested-dict as the data-structure for metadata, right?

About the default search_path, do you happen to know how it's set at the user level? If that comes from the database server then can we read that in while starting up? 

I think adding support for user-defined functions could be handled separately. We currently have built-in functions and keywords that are statically initialized, similarly we can populate a user-defined functions at startup time. 

Hinting the args is possible but that'll require a bit more plumbing. We'll let future @darikg and @amjith  worry about that. :P

**Comment by amjith** [comment]
You do have a point about various ways of setting the default, but I can't think of another way to handle it. Even if we decide to go with a different data-structure we'd still need to support a way to complete unqualed references, which means we'll either show way more table names for auto-completion or none at all.

I'm sure there is a way to get the search_path from the server itself, so we can merely refresh the default path whenever we encounter a `set` command instead of parsing the path from the sql statement itself.

**Comment by darikg** [comment]
>If I understand your correctly, if we make the default schema as a list, then we can stick with the nested-dict as the data-structure for metadata, right?

Yeah, exactly. 

> About the default search_path, do you happen to know how it's set at the user level? If that comes from the database server then can we read that in while starting up?

I know the query ```SHOW search_path``` returns a string containing a comma separated list of schema names. We could parse that (I guess taking special care for the edge case where a schema is double-quoted and contains a comma) in to a list of schema names. There might be a function in postgresql somewhere that returns a column of schema names instead, which would be nicer.

> You do have a point about various ways of setting the default, but I can't think of another way to handle it. Even if we decide to go with a different data-structure we'd still need to support a way to complete unqualed references

Well, it's [handled currently](https://github.com/darikg/pgcli/blob/78288101c20972b4749c04ab49bfba6dfd68717a/pgcli/pgexecute.py#L60) with the pg_catalog.pg_table_is_visible function, which does all the work for us. But then you'd either need to store the is_visible tag for each table (as in the current PR), or keep two redundant data structures (as we discussed upthread).

**Comment by amjith** [comment]
I think we can settle for using `SHOW search_path` and parsing the output right now using a new method called `refresh_search_path()`. 

We can trigger a refresh_search_path() when a `set` statement is entered. Let's punt on maintaining two different data-structures unless we run into a situation that demands it.

**Comment by darikg** [comment]
Sounds good. Turns out there's a slightly better way to get the search path:
```select unnest(current_schemas(false))``` which returns a column of schema names, instead of a single string.

Anyway, I can start implementing this, or if you'd prefer to, let me know.

**Comment by amjith** [comment]
If you can implement it, I'd very much appreciate it. But if you're swamped
let me know.

**Comment by darikg** [comment]
Sure, I can start working on it. Does it make sense to keep working on this branch, or make a new pr from master?

**Comment by amjith** [comment]
Same branch. But merge in master to bring this branch up to date.

**Comment by darikg** [comment]
Done!

todo: 
  [ ] Figure out a way to compare lists of dictionaries in test_sqlcompletion so the tests stop failing in python 3
  [ ] Use itertools.groupby to avoid unnecessary name escaping on the same schema & table names over and over
  [ ] Split test_smart_completion into two files, on for simpler tests on a public-schema-only database, and another for a more complicated structure
  [ ] More schema-related tests in general

**Comment by darikg** [comment]
Done!

todo: 
  - [ ] Figure out a way to compare lists of dictionaries in test_sqlcompletion so the tests stop failing in python 3
  - [ ] Use itertools.groupby to avoid unnecessary name escaping on the same schema & table names over and over
  - [ ] Split test_smart_completion into two files, on for simpler tests on a public-schema-only database, and another for a more complicated structure
  - [ ] More schema-related tests in general

**Comment by amjith** [comment]
Just tried it out. The completion works great. I haven't tested the path stuff yet, but I'll check it soon. 

One tiny issue found so far is the completion for special commands need to be adjusted. 

```
\d custom.<tab> 
```

That should show only tables from the custom schema. But that could be a future enhancement. 

Thank you for working on this feature. I'll read through the code later today. 

Cheers!

File: pgcli/main.py [hunk 3]
```diff
@@ -349,6 +352,13 @@ def quit_command(sql):
             or sql.strip() == '\q'
             or sql.strip() == ':q')
 
+def refresh_completions(pgexecute, completer):
```
amjith: Is this function still necessary?

darikg: Nope! I must've fat-fingered it back in during the merge

darikg:
The way I handled the \d command looks a little fragile, have a look and tell what you think

amjith:
You've been very productive!

I pulled the changes and tried it out it does work as expected. 

I looked at the code and you're right it's a wee bit fragile. I'm wondering why it wasn't handled by the elif condition below `elif token_v.endswith('.'):`. 

I still haven't had time (stuck on a unicode issue #134) to review the full change-set but the initial perusal looks good. I'll give it a good look tonight. 

Thanks!

darikg:
I think it's something wonky with sending non-sql through sqlparse:

```
In [1]: from sqlparse import parse

In [2]: p = parse(r'\d xxx')[0]

In [3]: p.tokens
Out[3]: [<Error '\' at 0x345ec28>, <Identifier 'd xxx' at 0x4457f48>]

In [4]: p.tokens[-1].value
Out[4]: u'd'
```

pgcli/packages/sqlcompletion.py
```diff
@@ -66,28 +66,65 @@ def suggest_based_on_last_token(token, text_before_cursor, full_text):
...
+
+        identifier = last_word(token_v[:-1], 'all_punctuations')
+
+        # TABLE.<suggestion> or SCHEMA.TABLE.<suggestion>
+        tables = extract_tables(full_text)
+        tables = [t for t in tables if identifies(identifier, *t)]
+        suggestions.append({'type': 'column', 'tables': tables})
+
+        # SCHEMA.<suggestion>
+        suggestions.append({'type': 'table', 'schema': identifier})
```
amjith: I'm not sure I fully understand the logic here. But it seems to work. I'll spend more time on this to understand it better. But for now, I'm happy to merge. :)

amjith:
This is quite the PR. :sparkles:

I appreciate the time you've put into this. Thank you very much!!

I've finally given it a thorough reading and I'm quite satisfied with all the changes. I'll come back to the suggestion for '.' at a later time. 

This is good for merge. :shaved_ice:

PR #129, opened 2015-01-19
Documentation and an addition to smart completion for ON keyword.

j-bennet:
Added a section about meta-commands to developer documentation.
ON keyword now suggests tables and aliases on the right side of equal sign also.


amjith:
Thank you for adding more docs. :headphones: 

I like the new addition to the completion. :+1:

PR #133, opened 2015-01-22
Cast bytea fields to text for output.

drocco007:
In the current version of pgcli, `bytea` fields trigger Unicode decode errors when queried:

    _test_db> CREATE TABLE binarydata(c bytea) 
    CREATE TABLE
    _test_db> INSERT INTO binarydata VALUES (decode('DEADBEEF', 'hex'))
    INSERT 0 1
    _test_db> SELECT * FROM binarydata
    'ascii' codec can't decode byte 0xde in position 0: ordinal not in range(128)
    _test_db> 

Casting the field to a string for display allows tables with `bytea` fields to be queried as expected. This pull request adds a type converter to have psycopg2 automatically perform the conversion on all `bytea` fields:

    _test_db> SELECT * FROM binarydata
    +------------+
    | c          |
    |------------|
    | \xdeadbeef |
    +------------+
    SELECT 1
    _test_db>

### DIFF
--- a/pgcli/pgexecute.py
+++ b/pgcli/pgexecute.py
@@ -13,6 +13,11 @@ _logger = logging.getLogger(__name__)
 psycopg2.extensions.register_type(psycopg2.extensions.UNICODE)
 psycopg2.extensions.register_type(psycopg2.extensions.UNICODEARRAY)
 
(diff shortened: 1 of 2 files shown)


amjith:
Awesome! Thank you for the PR. 

:tulip:

PR #137, opened 2015-01-24
Minor PGCli API improvements

darikg:
I've been slowly working on a pgcli autocompletion plugin for Sublime Text 3, and found myself importing and copy/pasting a little more than I'd like. This PR simply moves pgcompleter construction to PGCli.__init__, and adds a convenience method PGCli.get_completions(text, cursor_position) which delegates to self.completer.get_completions.


amjith:
:+1:

PR #142, opened 2015-01-28
Fix #106 - autocompletion in multiple statements

### LINKED ISSUE #106: Auto-completion is broken for multiple queries in the same line
eg: Let's assume $ is the cursor position.

```
misago_testforum> SELECT * FROM django_migrations; SELECT * FROM $
```

That should suggest table names, but instead I'm getting keywords.

darikg:


pgcli/packages/sqlcompletion.py
```diff
@@ -38,10 +38,30 @@ def suggest_type(full_text, text_before_cursor):
...
-    last_token = p and p.token_prev(len(p.tokens)) or ''
+    if len(parsed) > 1:
+        # Multiple statements being edited -- isolate the current one by
+        # cumulatively summing statement lengths to find the one that bounds the
+        # current position
+        current_pos = len(text_before_cursor)
+        stmt_start, stmt_end = 0, 0
+
+        for statement in parsed:
+            stmt_len = len(statement.to_unicode())
```
darikg: this could be `str(statement)` instead of `statement.to_unicode()` -- i don't know the difference

amjith: It's a deprecated method in sqlparse lib. But we need it. Because in Python 2 we need to use unicode(statement) in order to handle unicode characters, but Python 3 doesn't have the unicode() method. So instead of doing an if-check and switching to calling str() if python 3 and calling unicode() if python 2, we can just use the deprecated method which does this for us automatically.

amjith:
Looks good. 

:melon:

Issue #82, opened 2015-01-08
\dt schema.*


fzerorubigd:
`\dt *.*` and `\dt schema.*` has no effect at all. its always show public schema tables.

fzerorubigd:
I don't know if compatibility with psql is a goal or not? since I found the `\dn` is not implemented also.

amjith:
Yes `schema` support is still pending #22. 

Compatibility with psql is indeed a goal. But it takes time to catch up with years of work. :)

`\dn` is currently pending in a PR #62. I'll merge it in once I get a chance to review it.

fzerorubigd:
Thank you for your fast response!
So this is duplicate of them, I close this one.

amjith:
Schema support is added to master. Sql statements with schema names now have auto-completion (thanks @darikg). 

But \dt schema is still broken. So I'm reopening the issue. 

\dn is available in master.

darikg:
Looking into this a bit, the problem is that the \dt special command currently ignores the following argument, so no filtering of any sort is done, it just shows all public tables no matter what.

psql seems to work like the following pseudo-python:
```
# Accept an argument of the form [schema_filter.][table_filter]
if schema_filter:
    if schema_filter == '*':
        pass
    else:
        sql += " AND n.nspname ~ '^(schema_filter)$' "
else:
    sql += " AND pg_catalog.pg_table_is_visible(c.oid) "

if table_filter:
    if table_filter == '*':
        pass
    else:
        sql += " AND c.relname ~ '^(table_filter)$' "
```

So the biggest impediment to fixing this issue is figuring out how to parse the argument into a schema filter and table filter, converting wildcards '*' and '?' into regex syntax, and respecting double quotes

amjith:
There's already functions in pgspecial.py to do the wildcard conversion and double quotes escaping. 

https://github.com/amjith/pgcli/blob/master/pgcli/packages/pgspecial.py#L706

amjith:
This is now fixed in master. Thanks to @darikg. 

Can you try it out? You can install from the git repo directly:

```
pip install -U -e git+https://github.com/amjith/pgcli.git#egg=pgcli
```

fzerorubigd:
I get this error on debian, and archlinux right after launch pgcli : 
```
Traceback (most recent call last):
  File "/usr/bin/pgcli", line 5, in <module>
    from pkg_resources import load_entry_point
[...]

PR #144, opened 2015-02-01
make \dt and \dv use verbose and pattern arguments

darikg:
see #82

### DIFF
--- a/pgcli/packages/pgspecial.py
+++ b/pgcli/packages/pgspecial.py
@@ -50,11 +50,11 @@ def list_schemas(cur, pattern, verbose):
         headers = [x[0] for x in cur.description]
         return [(cur, headers, cur.statusmessage)]
 
-def list_tables_or_views(cur, pattern, verbose, relkinds):
+def list_objects(cur, pattern, verbose, relkinds):
     """
         Returns (rows, header, status)
 
-        This method is used by list_tables and list_views.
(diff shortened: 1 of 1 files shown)


pgcli/packages/pgspecial.py
```diff
@@ -50,6 +50,68 @@ def list_schemas(cur, pattern, verbose):
...
+        Returns (rows, header, status)
+
+        This method is used by list_tables and list_views.
+
+        relkinds is a list of strings to filter pg_class.relkind
+
+    """
+    schema_pattern, table_pattern = sql_name_pattern(pattern)
+
+    verbose_columns = ('''
```
amjith: Minor nitpick: 

Would it be better to use: 

```
if verbose: 
    verbose_columns = ''' blah blah'''
else:
    verbose_columns = ''
```

I think it makes the code easier to follow. I only say this because I glossed over the if condition at the end of the long sql and I had to backtrack to see where you handled the non-versbose case.

amjith:
Apart from the minor nitpick. PR looks great. 

Thank you!  :penguin:

darikg:
Sure thing. I also just realized that \di uses the same sql so that accepts verbose & pattern args now too

amjith:
It's so nice when things come together like this. :)

Good work on spotting it. :snowman:

PR #152, opened 2015-02-12
bugfix: \dt query shouldn't restrict schemas if schema qualifer supplied

darikg:
this was also my fault!


amjith:
Looks good. Thanks for catching this case. 

:bike:

PR #155, opened 2015-02-17
Autocomplete custom function names

darikg:
This would suggest the names of functions create via `CREATE FUNCTION ...`

It doesn't get metadata from functions defined in the pg_catalog schema, which is where `max`, `count`, `sum`, et al. are defined. There's over 2000 functions there so I thought it made sense to stick with the hardcoded whitelist approach currently used for those.

### DIFF
--- a/pgcli/pgcompleter.py
+++ b/pgcli/pgcompleter.py
@@ -43,7 +43,7 @@ class PGCompleter(Completer):
 
         self.special_commands = []
         self.databases = []
-        self.dbmetadata = {'tables': {}}
+        self.dbmetadata = {'tables': {}, 'functions': {}}
         self.search_path = []
 
         self.all_completions = set(self.keywords + self.functions)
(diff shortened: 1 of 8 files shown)


pgcli/pgcompleter.py
```diff
@@ -43,7 +43,7 @@ def __init__(self, smart_completion=True):
 
         self.special_commands = []
         self.databases = []
-        self.dbmetadata = {}
+        self.dbmetadata = {'tables': {}, 'functions': {}}
```
amjith: Would it make sense to separate this into two different dictionaries?

self.tables_metadata = {}
self.funcs_metadata = {}

I'm not suggesting we should do this, but I'm just thinking out loud. I see a lot of `metadata = self.dbmetadata['tables']` call sprinkled in the code. I wonder if separate vars will help.

darikg: I found the current nested approach useful in a couple spots:  like [here](https://github.com/amjith/pgcli/pull/155/files#diff-36fcb873925169e72e8c5ca4e8a8a581R91), where I loop over dbmetadata.keys(), and the `populate_schema_objects` method, which takes an argument like 'tables' or 'functions'. Neither of these are particularly important though, and could be pretty trivially replaced with explicit reference to separate dicts.

I think most of the `metadata = self.dbmetadata['tables']` stuff is just an unconcious/premature optimization on my part to pull dictionary lookups outside of loops. That pattern could stay or go independently of the dictionary storage.

amjith: I'm going to merge this in and then refactor it into a different var. Deep nested dicts make me nervous about the complexity. Even though there are comment strings, I'm afraid that we won't remember it very well when we come back to it in 6 months.

PR #158, opened 2015-02-24
Add completion for CREATE WITH TEMPLATE

### LINKED ISSUE #156: CREATE DATABASE <foo> WITH TEMPLATE <bar> does not autocomplete
There's not much in the postgres documentation about this, but it's super useful for cloning databases, and is much faster than a pg_restore.

http://stackoverflow.com/a/876565/188

drocco007:
Closes #156


amjith:
I think apart from the two suggestions, it looks good.

drocco007:
Ah, that separation of keywords and special commands makes much more sense! I have updated the PR with your suggested changes. :)

amjith:
Cheers! 

Thank you for fixing the issues and making sure to add tests. :ice_cream:

PR #161, opened 2015-02-27
Some small \special improvements

darikg:
-  Move special-related suggestions out of `suggest_based_on_last_token` and into a new `suggest_special` method
    - Special commands don't really obey sql syntax so this makes it easier to work around sqlparse
    - Have \d, \dt, and \dv suggest schema and table names
  - Add support for \df 
      - \df suggests schema and function names
  - \dn suggests schema names

### DIFF
--- a/pgcli/packages/sqlcompletion.py
+++ b/pgcli/packages/sqlcompletion.py
@@ -3,6 +3,7 @@ import sys
 import sqlparse
 from sqlparse.sql import Comparison
 from .parseutils import last_word, extract_tables, find_prev_keyword
+from .pgspecial import parse_special_command
 
 PY2 = sys.version_info[0] == 2
 PY3 = sys.version_info[0] == 3
@@ -30,7 +31,8 @@ def suggest_type(full_text, text_before_cursor):
     # partially typed string which renders the smart completion useless because
     # it will always return the list of keywords as completion.
     if word_before_cursor:
-        if word_before_cursor[-1] in ('(', '.'):
+        if (word_before_cursor[-1] in ('(', '.')
+                or word_before_cursor[0] == '\\'):
             parsed = sqlparse.parse(text_before_cursor)
         else:
             parsed = sqlparse.parse(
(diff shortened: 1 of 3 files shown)


pgcli/packages/sqlcompletion.py
```diff
@@ -30,7 +31,8 @@ def suggest_type(full_text, text_before_cursor):
     # partially typed string which renders the smart completion useless because
     # it will always return the list of keywords as completion.
     if word_before_cursor:
-        if word_before_cursor[-1] in ('(', '.'):
+        if (word_before_cursor[-1] in ('(', '.')
+                or word_before_cursor[0] == '\\'):
```
amjith: There could be leading spaces that can thwart this logic. Try this for example: 

```
postgres>       \df <tab>
```

It doesn't trigger the right auto-completion. 

It might be worth stripping the leading white-space from word_before_cursor and then check for '\\'.

amjith:
Change looks good except for that one corner case with leading whitespace. 

Would it be possible to also get a list of views and suggest that for `\dv` instead of just schemas and tables? Kinda like how `\df` suggests the list of functions? 

In fact, there is an old issue #39 requesting view names to be included in the completion list. That's completely optional of course. :)

pgcli/packages/sqlcompletion.py
darikg: Thanks for catching that. Added an extra commit to fix it

darikg:
Right now \dt and \dv suggest views and table names interchangeably, because of [this](https://github.com/amjith/pgcli/blob/09da888f88d46eddf01b73c0453f603620f781cf/pgcli/pgexecute.py#L45). In order to separate them, I think we would have additional methods `pgexecute.views()` and `pgcompleter.extend_views()`, which would store view names/columns in `pgcompleter.dbmetadata['views']`, etc. Is that how you were picturing it? If so, I can do it as a separate PR sometime tomorrow.

amjith:
New commit works as expected. Thanks for fixing it. 

Yes, I did envision having `pgexecute.views()`, it's extend version and an entry into dbmetadata. If you're planning to do it, I'll stay off. It's nothing urgent, so don't worry if you can't get to it right away.

PR #165, opened 2015-03-01
Interpret incoming JSON as a string instead of via json.loads

### LINKED ISSUE #163: JSON display shows u"..."
When `pgcli` is run in Python 2, JSON fields (`json` and `jsonb` types in PostgreSQL), are shown with `u"..."` prefixes for strings. I guess this is a side-effect of using `repr()`. Maybe the `pprint` module can be leveraged instead?

drocco007:
Replace the default JSON (and JSONB, if defined) type casters with versions that handle the incoming data as strings instead of via json.loads. This fixes #163 and makes JSON render the same way in pgcli as it does in psql. The typecaster uses the current connection's encoding — which defaults to the database encoding — to interpret the incoming data; the implementation should continue to behave correctly if someone ever adds the ability to change the client encoding to pgcli.

I tested this against 9.3 and 9.4 servers (yay Docker!).

Comments welcome. :)

### DIFF
--- a/pgcli/pgexecute.py
+++ b/pgcli/pgexecute.py
@@ -23,6 +23,34 @@ ext.register_type(ext.new_type((17,), 'BYTEA_TEXT', psycopg2.STRING))
 # See http://initd.org/psycopg/articles/2014/07/20/cancelling-postgresql-statements-python/
 ext.set_wait_callback(psycopg2.extras.wait_select)
 
+
+def register_json_typecasters(conn, loads_fn):
+    """Set the function for converting JSON data for a connection.
+
+    Use the supplied function to decode JSON data returned from the database
+    via the given connection. The function should accept a single argument of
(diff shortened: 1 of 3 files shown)


drocco007:
Hmm. The build is failing because the test server's database doesn't have the JSON types. I'll adjust the skip conditions on those tests to check that the server has the types defined, which would also let anyone who installed the types via an extension run the tests as well.

drocco007:
OK, it passes Travis! :smiley:

amjith:
Thank you for the PR to handle `json` types the right way. I appreciate the tests and the updates to make them pass in Travis. :)

There is one small corner case in the exception handling I commented about. Otherwise the change looks great!

:soccer:

drocco007:
Done! Since the tests and the connection class used basically the same code, I made a common helper and eliminated some duplication.

amjith:
Thank you! 

:cow2:

PR #170, opened 2015-03-06
Better handle schema qualifications in current word

darikg:
short version: fixes a bug where `DROP TABLE schema_name.<TAB>` and `DROP FUNCTION schema_name.<TAB>` would both suggest both tables and functions from schema_name.

Previously, if the current word contained a period, `suggest_type` would parse text to the right of the period as `word_before_cursor`, so that characters to the left of the period were parsed as the final token. Then `suggest_based_on_last_token` would generate suggestions under the generic catch-all `elif token_v.endswith('.')`. This means that, for example, `DROP TABLE schema_name.<TAB>` and `DROP FUNCTION schema_name.<TAB>` would each suggest both tables and functions from schema_name.

This commit PR changes this behavior to include schema qualifications as part of word_before_cursor, so that `suggest_based_on_last_token` operates on the actual preceding token, not just the schema name. The schema name itself is parsed as an sqlparse Identifier, and passed as an extra argument to suggest_based_on_last_token. This allows each token value to handle dot-qualifications differently, and may eventually support things like function output arguments `func().fieldname` or custom types with `arbitrarily.deep.nested.child.fields`

### DIFF
--- a/pgcli/packages/parseutils.py
+++ b/pgcli/packages/parseutils.py
@@ -7,7 +7,7 @@ from sqlparse.tokens import Keyword, DML, Punctuation
 cleanup_regex = {
         # This matches only alphanumerics and underscores.
         'alphanum_underscore': re.compile(r'(\w+)$'),
-        # This matches everything except spaces, parens and comma.
+        # This matches everything except spaces, parens, comma, and period
         'most_punctuations': re.compile(r'([^\.(),\s]+)$'),
         # This matches everything except a space.
         'all_punctuations': re.compile('([^\s]+)$'),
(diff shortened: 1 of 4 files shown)


amjith:
Haven't done a thorough reading of the changes yet. Cursory look seems good. I'll give it a good read tonight and provide feedback (or merge it in).

amjith:
Works beautifully. :)

I was a little embarrassed when I first added the var `most_punctuations`, since that is a terrible name for a variable. Now I see you've added `many_punctuations` which either adds to the injury or gives me comfort in the fact that I'm not alone in my inability to come up with a more descriptive variable name.

:boat:

darikg:
lol. I figure it's fine until someone wants `mostest_punctuation`

PR #172, opened 2015-03-10
Format Decimal values directly as strings

### LINKED ISSUE #169: Big Numbers are displayed in scientific notation
```
misago_testforum> CREATE TABLE numeric_test (a numeric);
CREATE TABLE
misago_testforum> INSERT INTO numeric_test VALUES (100)
INSERT 0 1
misago_testforum> INSERT INTO numeric_test VALUES (10000000)
INSERT 0 1
misago_testforum> SELECT * FROM numeric_test;
+---------+
|       a |
|---------|
| 100     |
|   1e+07 |
+---------+
SELECT 2
misago_testforum>
```

This should not be the case. pgcli should always try and use non-scientific notation for displaying numbers.

drocco007:
The failure listed in the issue is rooted in how tabulate handles Decimal objects, which, at bottom, end up being rendered via

    >>> format(float(Decimal(10000000)), 'g')
    '1e+07'

The 'g' format specifier is parameterized in tabulate, but changing it to '' only fixes some of the cases.

Next, I tried asking psycopg2 to return these values as strings rather than Decimal objects, but this also fails due to tabulate's aggressive coercion logic, which ends up giving you

    >>> format(float('10000000'), 'g')
    '1e+07'

The best solution is to render Decimal values directly as strings, which will correctly handle the difference between 10000000 and 10000000.0. Here is an example in pgcli

    postgres> SELECT * FROM numeric_test ORDER BY a
    +---------------+
    | a             |
    |---------------|
    | 1.0           |
    | 1.01          |
    | 10.0          |
    | 100           |
    | 10000000      |
    | 10000000.0    |
    | 1000000000000 |
    +---------------+
    SELECT 7
    Command Time: 9.05990600586e-06
    Format Time: 0.00118017196655
    postgres> 

 and the same query in psql
    
    postgres=# select * from numeric_test order by a;
           a       
    ---------------
               1.0
              1.01
              10.0
               100
          10000000
        10000000.0
     1000000000000
    (7 rows)
    
    postgres=# 

I would normally avoid patching a vendored library like this, but this seems like the cleanest solution under the circumstances and matches previous fixes to a few other types.

Closes #169

### DIFF
--- a/pgcli/packages/tabulate.py
+++ b/pgcli/packages/tabulate.py
@@ -5,6 +5,7 @@
 from __future__ import print_function
 from __future__ import unicode_literals
 from collections import namedtuple
+from decimal import Decimal
 from platform import python_version_tuple
 import re
 
@@ -331,7 +332,7 @@ def _type(string, has_invisible=True):
 
     if string is None:
         return _none_type
-    if isinstance(string, bool):
+    if isinstance(string, (bool, Decimal,)):
         return _text_type
     elif hasattr(string, "isoformat"):  # datetime.datetime, date, and time
         return _text_type
--- a/tests/test_pgexecute.py
+++ b/tests/test_pgexecute.py
@@ -147,3 +147,13 @@ def test_jsonb_renders_without_u_prefix(executor, expanded):
     result = run(executor, "SELECT d FROM jsonbtest LIMIT 1", join=True)
 
     assert u'{"name": "Éowyn"}' in result
+
+
+@dbtest
+@pytest.mark.parametrize('value', ['10000000', '10000000.0', '10000000000000'])
+def test_large_numbers_render_directly(executor, value):
+    run(executor, "create table numbertest(a numeric)")
+    run(executor,
+        "insert into numbertest (a) values ({0})".format(value))
+
+    assert value in run(executor, "select * from numbertest", join=True)


amjith:
I started poking around the `format()` function and the float format specifier. Eventually I found this answer: http://stackoverflow.com/a/2440786

Which renders a reasonably good output:

```
misago_testforum> SELECT * FROM numeric_test ORDER BY a;
+----------------------+
|                    a |
|----------------------|
|                 1    |
|                 1    |
|                 1.01 |
|                10    |
|               100    |
|               100    |
|          10000000    |
|       10000000000    |
|       10000000000    |
| 10000000000000000    |
+----------------------+
SELECT 10
```

Compared to psql output: 

```
         a
-------------------
               100
          10000000
                 1
               1.0
              1.01
              10.0
               100
       10000000000
     10000000000.0
 10000000000000000
```

But then it removes the decimals and insignificant zeros, which can be misleading in some rare cases. So I think you made the right choice by forcing them to render as strings. We'll miss out on the right alignment, oh well. 

BTW, the tests are [failing](https://travis-ci.org/amjith/pgcli/jobs/53747313) on Python 2.6 because the tests you wrote use a 2.7 feature for formatting strings. 

`"insert into numbertest (a) values ({})".format(value))` should be rewritten as `"insert into numbertest (a) values ({0})".format(value))`, notice the {0} in the place holder instead of the empty {}. 

Would you mind updating the test to have it pass on 2.6? I'll merge it in once that is resolved. 

Thank you for diving into the code and finding the root cause. :+1:

drocco007:
Done! I always manage to forget that the shorter format syntax is a 2.7+ thing… :/

If, in the future, you wanted to add decimal alignment, I don't think it would be too hard: it would mostly be a matter of teaching tabulate how to treat the Decimal type. It would probably require some tabulate surgery, since the tool doesn't expose the hook points that would be needed, but the actual change would probably be pretty straightforward.

Cheers!
dan

amjith:
Yeah I've been meaning to do a lot of improvements to Tabulate to feed it the types instead of having it guess them. But haven't found the time to do it yet. :P 

Thank you for updating the tests. 

:alien:

PR #185, opened 2015-04-04
Chinese display

xalley:
displaychinese

Before computing the length of the string, encode it to `gbk` first,
the length of gbk bytecode is exactly equal to the visible length of the string.
And padding space manually instead of using function `format`.

By the way, it will crash when I type Chinese in windows,
I fix it with adding some code to `prompt_toolkit\terminal\win32_input.py`.
But who shall I give it to?
The prompt_toolkit in github is not version 0.26.

Finally, I am not sure whether it's robust and hope you can understand my poor English.

### DIFF
--- a/pgcli/packages/tabulate.py
+++ b/pgcli/packages/tabulate.py
@@ -23,6 +23,8 @@ if python_version_tuple()[0] < "3":
     def _is_file(f):
         return isinstance(f, file)
 
+    import __builtin__ as builtins
+
 else:
     from itertools import zip_longest as izip_longest
     from functools import reduce, partial
@@ -37,6 +39,14 @@ else:
     def _is_file(f):
         return isinstance(f, io.IOBase)
 
+    import builtins
+
+def len(lst):
+    """rewrite `len`, compute the visible length of _text_type"""
+    if isinstance(lst, _text_type):
+        return lst.encode('gbk').__len__()
+    else:
+        return builtins.len(lst)
 
 __all__ = ["tabulate", "tabulate_formats", "simple_separated_format"]
 __version__ = "0.7.4"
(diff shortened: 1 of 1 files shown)


amjith:
Hi @xalley,

Thank you for the PR. You're addressed an important issue with the display. 

But the solution only addresses Chinese characters. Would you be willing to use this package [wcwidth](https://pypi.python.org/pypi/wcwidth/0.1.4) which is designed for calculating the printable length of a string? 

In fact wcwidth is already one of the dependencies for prompt_toolkit, so it's just a matter of calling the right apis. :)

BTW, can you send me the fix for the win32_input.py fix? 

If it is not fixed in the latest version of prompt_toolkit, then you should also send it to the author of prompt_toolkit, but if it is already fixed in the newer version, I'll just backport the fix for now and upgrade to the newer version later.

Thank you!

xalley:
Oh, I see. Thank you very much ^^!

xalley:
I have use wcwidth now.

amjith:
Did you push the new changes with wcwidth to the Pull Request? 

I don't see the new commit or the use of wcwidth in the file diff.

xalley:
I am sorry that I don't know how to update a Pull Request. I just push to my repo. I am finding the update button now /_\

amjith:
That's very odd. Usually when you update your repo and if there is a pending Pull Request, then the changes will automatically updated into the Pull Request. 

But I don't see them in the PR. Which makes me wonder if there was a problem with the update. 

I'm going to close this PR, can you try creating a new PR once I close this?

xalley:
I seem that there is something wrong with github.
When I create a new PR, always appear "Server Error · GitHub"

amjith:
I'm getting the same error. I think there is a problem with Github. We can try in a few hours if that doesn't work, I'll just pull down your repo and manually merge them in tomorrow.

Thank a lot for your work. :+1:

PR #190, opened 2015-04-06
some join-related improvements

darikg:
* fix some bugs where `t1 JOIN t2 ...` worked fine but `t1 INNER JOIN t2...` was broken
  * support `t1 INNER JOIN t2 USING (` which suggests only columns in common

### DIFF
--- a/pgcli/packages/parseutils.py
+++ b/pgcli/packages/parseutils.py
@@ -85,8 +85,11 @@ def extract_from_part(parsed, stop_at_punctuation=True):
             # StopIteration. So we need to ignore the keyword if the keyword
             # FROM.
             # Also 'SELECT * FROM abc JOIN def' will trigger this elif
-            # condition. So we need to ignore the keyword JOIN.
-            elif item.ttype is Keyword and item.value.upper() not in ('FROM', 'JOIN'):
+            # condition. So we need to ignore the keyword JOIN and its variants
+            # INNER JOIN, FULL OUTER JOIN, etc.
+            elif item.ttype is Keyword and (
+                    not item.value.upper() == 'FROM') and (
+                    not item.value.upper().endswith('JOIN')):
                 raise StopIteration
             else:
                 yield item
--- a/pgcli/packages/parseutils.py
+++ b/pgcli/packages/parseutils.py
@@ -152,12 +152,22 @@ def extract_tables(sql):
     return list(extract_table_identifiers(stream))
 
 def find_prev_keyword(sql):
-    if not sql.strip():
-        return None
+    """ Find the last sql keyword in an SQL statement
 
-    for t in reversed(list(sqlparse.parse(sql)[0].flatten())):
+    Returns the value of the last keyword, and the text of the query with
+    everything after the last keyword stripped
+    """
+    if not sql.strip():
+        return None, ''
+
+    parsed = sqlparse.parse(sql)[0]
+    for t in reversed(list(parsed.flatten())):
         if t.is_keyword or t.value == '(':
-            return t.value
+            idx = parsed.token_index(t)
+            text = ''.join(tok.value for tok in parsed.tokens[:idx+1])
+            return t.value, text
+
+    return None, ''
 
 if __name__ == '__main__':
     sql = 'select * from (select t. from tabl t'
--- a/pgcli/pgcompleter.py
+++ b/pgcli/pgcompleter.py
@@ -5,6 +5,11 @@ from .packages.sqlcompletion import suggest_type
 from .packages.parseutils import last_word
 from re import compile
 
+try:
+    from collections import Counter
+except ImportError:
+    # python 2.6
+    from .packages.counter import Counter
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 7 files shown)


darikg:
does pgcli officially support python 2.6?

amjith:
Yes it does. It's funny you got tripped up by Counter. I had to deal with that in one of my other projects last week and I was mildly annoyed by it. 

I ended up using http://code.activestate.com/recipes/576611-counter-class/ which is the official implementation of Counter class by the great Raymond Hettinger. :)

I think it's worth maintaining that compatibility since a lot of linux distros still ship with 2.6 as the default. 

But if it becomes too much of a trouble we can always reevaluate.

darikg:
That's fair. Where do you want it? Something like pgcli.backports.Counter?

amjith:
I was thinking 'pgcli/packages/counter.py'. 

'packages' folder is supposed to be the folder for keeping all third-party packages. Sometime in the future they'll be spun off into separate projects (or dropped) and pgcli will install them as dependencies.

darikg:
now py26 compatible

amjith:
@darikg I'm getting ready for PyCon. So I haven't been able to give it due attention. I'll have plenty of time to review it in the plane. I'll provide feedback or merge it in once I get to PyCon. 

Are you planning to attend this year? It'll be nice to meet you in person. :)

darikg:
Yeah no worries. I can't make PyCon this year but have a great time! Montreal's amazing.

pgcli/packages/parseutils.py
```diff
@@ -149,12 +152,22 @@ def extract_tables(sql):
...
+        return None, ''
 
-    for t in reversed(list(sqlparse.parse(sql)[0].flatten())):
+    parsed = sqlparse.parse(sql)[0]
+    for t in reversed(list(parsed.flatten())):
         if t.is_keyword or t.value == '(':
-            return t.value
+            idx = parsed.token_index(t)
+            text = ''.join(tok.value for tok in parsed.tokens[:idx+1])
+            return t.value, text
```
amjith: I don't understand why this change was made. I see that it's used to update the `text_before_cursor` in line 230 in sqlcompletion.py. But I'm lacking the context as to why.

pgcli/pgcompleter.py
```diff
@@ -185,6 +190,15 @@ def get_completions(self, document, complete_event, smart_completion=None):
...
                 _logger.debug("Completion column scope: %r", tables)
                 scoped_cols = self.populate_scoped_cols(tables)
+
+                if suggestion.get('drop_unique'):
+                    # drop_unique is used for 'tb11 JOIN tbl2 USING (...'
+                    # which should suggest only columns that appear in more than
+                    # one table
+                    scoped_cols = [col for (col, count)
+                                         in Counter(scoped_cols).items()
+                                           if count > 1 and col != '*']
```
amjith: I was thinking that using a `set` and doing an intersection would produce the same result, but I presume the code won't be as simple. Is that right?

amjith:
I finally got around to reviewing the code. I like the `endswith('join')` change. So :+1:  on that. 

I don't understand the changes made to `parse_utils.find_prev_keyword()`. If you could explain that one, I'll be happy to merge it in.

darikg:
Sorry about the delay getting back to you on that. Let me try to explain what I was thinking, and you can let me know if it makes sense, or if I'm just recklessly over-complicating things (it's probably the latter.) 

Consider the following two cases, where the asterisk indicates the cursor position:

**Case 1**:  `...s JOIN t USING (*`
  In `suggest_based_on_last_token`, the last token is `(`, and we need to check the preceding token to determine whether the opening parens is due to a subselect, `USING` clause, or function call. Easy peasy.

**Case 2**: `...s JOIN t USING (foo, bar, *`
  This is a bit trickier. The general strategy in `suggest_based_on_last_token` when the last token is `,` is to find the preceding keyword, and recursively call `suggest_based_on_last_token` with that as `token`, effectively requesting suggestions for the first column in the list, i.e.: reduce Case 2 to Case 1. The problem is of course Case 1 doesn't just depend on the ultimate token `(`, it also depends on the penultimate token, which it tries to extract from `text_before_cursor`, which in this case is `bar`, a not very informative token.

I figured there's two avenues to address this: either by complicating code in the `elif token == '('` branch, by smartening up its search for the penultimate token, or by complicating code in the original `suggest_based_on_last_token` call, in the `elif token == ','` branch. My rationale was that this recursive strategy was attempting to reduce the problem to being the first item in a list, so in addition to converting `token` from `,` to `(`, it should also convert `text_before_cursor` from Case 2 to Case 1. That's what the extra code in `find_prev_keyword` does -- instead of just returning the previous keyword, it also returns `text_before_cursor` with everything after that keyword removed.

Wow that was a lot of words. Definitely overcomplicating things. If you want I can pop off the `USING` commits into a separate PR so you can go ahead and merge the simple `endswith('join')` stuff.

pgcli/pgcompleter.py
darikg: Consider `...s JOIN t USING (x) JOIN u USING (` which should suggest columns that are present in any two of s, t, u. I think set intersection would only suggest columns that are present in all three, right? So I guess we want the union of all pairwise intersections?

amjith: Yeah, you're right. I think Counter is better in this case.

amjith:
I've read through your explanation (twice). I think I just need to go through some examples (maybe tests) and work it out for myself. 

I'll do that and then merge it in. I'll probably write a big fat comment on the line to remind my future self. :)

amjith:
Thank you for the PR. I've merged it in since it was doing the right thing when I tried it out. 

But I still have a question. 

You mentioned that it was added to address the case of ` ... s JOIN USING (foo, bar, * `. But don't we already have to handle that case in `INSERT INTO demo (id, name, * `.

I'm going to read through the code to try and understand what made the `USING` case special.

ps: I got to learn about USING in Postgres as a result of this PR, which is always nice. :)

amjith:
Alright, I'm finally caught up. 

Brilliant work! 

I think this paves the way for adding more suggestion types when we discover cases that are more complicated. I'll still go through and add the comments. 

Thank you very much for doing this. :)

:tractor:

PR #194, opened 2015-04-09
Fix the checking to commands '\c' or 'use'

xalley:
I want to add a easter egg for my friend -- command `\cf` -- her name abbreviation,
but when I type `\cf`, it raises `Database name missing.`,
so I have fixed it.

I think it might be useful in the future.

### DIFF
--- a/pgcli/pgexecute.py
+++ b/pgcli/pgexecute.py
@@ -178,7 +178,7 @@ class PGExecute(object):
             # exception that cannot be offloaded to `pgspecial` lib. Because we
             # have to change the database connection that we're connected to.
             command = sql.split()[0]
-            if command == '\c' or command.lower() == 'use':
+            if command == '\c' or command == '\connect' or command.lower() == 'use':
                 _logger.debug('Database change command detected.')
                 try:
                     dbname = sql.split()[1]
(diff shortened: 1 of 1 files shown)


pgcli/pgexecute.py
```diff
@@ -177,7 +177,8 @@ def run(self, statement):
             # Check if the command is a \c or 'use'. This is a special
             # exception that cannot be offloaded to `pgspecial` lib. Because we
             # have to change the database connection that we're connected to.
-            if sql.startswith('\c') or sql.lower().startswith('use'):
+            command = sql.split()[0]
+            if command == '\c' or command.lower() == 'use':
```
amjith: Can you add a check for `command == '\connect'` ? 

That was the reason I had it set to startswith `\c` but you're right I shouldn't generalize too much.

amjith:
It warms my heart to know that people are adding easter eggs to pgcli. :)

Just make sure you add a check for `\connect` and you're good. Thanks!

xalley:
OK :smiley:

amjith:
:+1:  

Cheers! :bee:

xalley:
:smirk:
Cheers! :wine_glass:

Issue #192, opened 2015-04-08
pg_catalog in search_path causes error in completion.


asenyshyn:
if "search_path" contains schema "pg_catalog" application may crash.

```sh
File "/home/ubuntu/tmp/env/local/lib/python2.7/site-packages/pgcli/pgcompleter.py", line 199, in get_completions suggestion['schema'], 'tables')
File "/home/ubuntu/tmp/env/local/lib/python2.7/site-packages/pgcli/pgcompleter.py", line 268, in populate_schema_objects for obj in metadata[schema].keys()]
KeyError: u'pg_catalog'
```

amjith:
@asenyshyn Can you tell me which version of pgcli you're using? 

I believe @darikg actually fixed this in one of the previous releases.

darikg:
Nope :( I can confirm this is a bug, another one I'm responsible for. Will fix tomorrow. Sorry!

amjith:
This is now fixed in master. Thanks to @darikg.

You can install directly from master using the following command to check it out. 

```
pip install -U -e git+https://github.com/amjith/pgcli.git#egg=pgcli
```

I'm planning to do a release sometime this weekend, so you should be able to upgrade directly from PyPI at that time.

PR #195, opened 2015-04-09
Improve handling of pg_catalog in metadata

darikg:
Originally, tables and schemas starting with `pg_`, as well as `information_schema` were not stored, under the assumption they would generate too many superfluous suggestions. This broke as soon as someone tried to manually add pg_catalog to the search path. #192 

Now, metadata is stored for all tables and schemas, and pg_catalog is correctly reported as being on the search_path, even though it is so implicitly. In order to prevent extraneous suggestions from `pg_catalog`, suggestions starting with `pg_` are suppressed, unless the user has already typed `pg_`, which seems like a reasonable compromise to me.

### DIFF
--- a/pgcli/pgcompleter.py
+++ b/pgcli/pgcompleter.py
@@ -201,18 +201,37 @@ class PGCompleter(Completer):
 
             elif suggestion['type'] == 'schema':
                 schema_names = self.dbmetadata['tables'].keys()
+
+                # Unless we're sure the user really wants them, hide schema
+                # names starting with pg_, which are mostly temporary schemas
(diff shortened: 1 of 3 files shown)


amjith:
Sorry mate. Poor internet connection and lack of sleep is making it impossible to review code with clarity. 

Thank you for the quick turn around. I promise I'll get to it as soon as I'm back from PyCon, early next week. :)

amjith:
Looks good. BTW, that's a clever idea to suggest temporary tables/schemas when the user types 'pg_'.

:circus_tent:

PR #202, opened 2015-04-13
Completion search text matches user-defined entities anywhere in the name

### LINKED ISSUE #193: Switch the string prefix matching to subsequence matching.
Taken from a reddit comment: http://www.reddit.com/r/programming/comments/31pkwf/mysqlcli_cli_with_autocompletion_and_syntax/cq5ftcy

>  A critical thing to notice is that the completion filtering is NOT based on the input being a string prefix of the completion (but that works too). The input needs to be a subsequence match of a completion. This is a fancy way of saying that any input characters need to be present in a completion string in the order in which they appear in the input. So abc is a subsequence of xaybgc, but not of xbyxaxxc. After the filter, a complicated sorting system ranks the completion strings so that the most relevant ones rise to the top of the menu (so you usually need to press TAB just once).

It should be a simple matter of changing this function [find_matches()](https://github.com/amjith/pgcli/blob/master/pgcli/pgcompleter.py#L160..L165) to do substring matching.

drocco007:
When searching for completions in smart completion mode, a potential
suggestion of a user-specified name (e.g. tables, functions, etc.) is
considered a match if the search text appears anywhere in the
suggestion. Keywords, special commands, and built-in functions still
use startswith matching.

Closes #193

### DIFF
--- a/pgcli/pgcompleter.py
+++ b/pgcli/pgcompleter.py
@@ -31,9 +31,8 @@ class PGCompleter(Completer):
             'UPDATE', 'USE', 'USER', 'VALIDATE', 'VALUES', 'VARCHAR',
             'VARCHAR2', 'VIEW', 'WHEN', 'WHENEVER', 'WHERE', 'WITH', ]
 
-    functions = ['AVG', 'COUNT', 'DISTINCT', 'FIRST', 'FORMAT', 'LAST',
-            'LCASE', 'LEN', 'MAX', 'MIN', 'MID', 'NOW', 'ROUND', 'SUM', 'TOP',
-            'UCASE']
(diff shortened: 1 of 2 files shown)


amjith:
:+1: 

That's a clever idea to implement the start-only logic for built-ins. :clap: 

Thank you!

drocco007:
:blush: thanks! :)

amjith:
An unexpected pleasant surprise is now the auto-escaping with quote is twice as awesome. 

For eg:

screenshot 2015-04-16 07 21 32

Previously I would have to start with a quote character in order to show table names with spaces. Now I don't have to. 

This is brilliant. :)

PR #204, opened 2015-04-13
Added to-be-implemented commands.

j-bennet:
Added some commands yet-to-be implemented.
Added "unimplemented" message if such a command is called.
Reformatted help docstrings.


pgcli/packages/pgspecial.py
```diff
@@ -948,17 +954,25 @@ def toggle_timing(cur, arg, verbose):
...
+            '\\ds': (list_sequences, ['\\ds[+] [pattern]', 'List sequences.']),
+            '\\df': (list_functions, ['\\df[+] [pattern]', 'List functions.']),
+            '\e': (in_progress, ['\e [file] [line]', 'Not yet implemented.']),
+            '\ef': (in_progress, ['\ef [funcname [line]]', 'Not yet implemented.']),
+            '\sf': (in_progress, ['\sf[+] funcname', 'Not yet implemented.']),
+            '\z': (in_progress, ['\z [pattern]', 'Not yet implemented.']),
+            '\do': (in_progress, ['\do[S] [pattern]', 'Not yet implemented.']),
+            'dT': (in_progress, ['\dT[S+] [pattern]', 'Not yet implemented.']),
+            '\copy': (in_progress, ['\copy ...', 'Not yet implemented.']),
+            '\watch': (in_progress, ['\watch [sec]', 'Not yet implemented.'])
```
amjith: I'm not sure if we can implement `\watch` command. I'd be a bit more wary of giving people false hopes. :)

pgcli/packages/pgspecial.py
```diff
@@ -948,17 +954,25 @@ def toggle_timing(cur, arg, verbose):
...
+            '\\dv': (list_views, ['\\dv[+] [pattern]', 'List views.']),
+            '\\ds': (list_sequences, ['\\ds[+] [pattern]', 'List sequences.']),
+            '\\df': (list_functions, ['\\df[+] [pattern]', 'List functions.']),
+            '\e': (in_progress, ['\e [file] [line]', 'Not yet implemented.']),
+            '\ef': (in_progress, ['\ef [funcname [line]]', 'Not yet implemented.']),
+            '\sf': (in_progress, ['\sf[+] funcname', 'Not yet implemented.']),
+            '\z': (in_progress, ['\z [pattern]', 'Not yet implemented.']),
+            '\do': (in_progress, ['\do[S] [pattern]', 'Not yet implemented.']),
+            'dT': (in_progress, ['\dT[S+] [pattern]', 'Not yet implemented.']),
+            '\copy': (in_progress, ['\copy ...', 'Not yet implemented.']),
```
amjith: I think `\copy` could be an expensive operation when done in Python. I'd change the doc string on this from 'Not yet implemented' to 'Not supported.'

amjith:
Thanks for the PR. I've left a couple of suggestions. I'd be interested to know your thoughts on those. 

Even though the goal is to be a drop-in replacement for psql, I wan to set realistic expectations for the users in terms of the things we can do. There are certain things that are not suitable to do in Python.

j-bennet:
Good point. I removed those two. I think it's better to not show them at all if we're not going to support them.

amjith:
Thank you! 

:microscope:

PR #209, opened 2015-04-18
Register the json handler only for PY2.

### LINKED ISSUE #201: Error when selecting json/jsonb values
When a query returns a `json` or `jsonb` value, `pgcli` throws an exception: `'str' object has no attribute 'decode'`:
```
$ pgcli
Version: 0.16.3
Chat: https://gitter.im/amjith/pgcli
Home: http://pgcli.com
ilmari> select '{"foo":42}'::json; 
'str' object has no attribute 'decode'
ilmari> select '{"foo":42}'::jsonb;
'str' object has no attribute 'decode'
ilmari> 
```

amjith:
Closes #201. 

@drocco007 Can you please verify this PR? 

The json handler is trying to decode the incoming string into unicode based on the database encoding. But this is only required if we're in Python 2, because in Python 3 psycopg2 already converts the string to unicode. But I want to double check my understanding.

The test case shown in #201 clearly highlights the problem. 

```
pgcli>  select '{"foo":42}'::json;
```

### DIFF
--- a/pgcli/pgexecute.py
+++ b/pgcli/pgexecute.py
@@ -141,8 +141,7 @@ class PGExecute(object):
             self.conn.close()
         self.conn = conn
         self.conn.autocommit = True
-        if PY2:
-            register_json_typecasters(self.conn, self._json_typecaster)
+        register_json_typecasters(self.conn, self._json_typecaster)
         register_hstore_typecaster(self.conn)
 
     def _json_typecaster(self, json_data):
@@ -154,7 +153,10 @@ class PGExecute(object):
         See http://initd.org/psycopg/docs/connection.html#connection.encoding
         """
 
-        return json_data.decode(self.conn.encoding)
+        if PY2:
+            return json_data.decode(self.conn.encoding)
+        else:
+            return json_data
 
     def run(self, statement):
         """Execute the sql in the database and return the results. The results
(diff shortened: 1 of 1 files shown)


drocco007:
Hi @amjith,

psycopg2's default JSON handler also runs the result through `loads`, with the result that the displayed value will render as a Python dict instead of JSON. Here is an example that demonstrates the difference: this  branch running under Python 2 renders like this

```
postgres> SELECT '{"foo":42, "bar":true}'::json;
+------------------------+
| json                   |
|------------------------|
| {"foo":42, "bar":true} |
+------------------------+
SELECT 1
```

while under Python 3 the same query renders as

```
postgres> SELECT '{"foo":42, "bar":true}'::json;
+--------------------------+
| json                     |
|--------------------------|
| {'foo': 42, 'bar': True} |
+--------------------------+
SELECT 1
```

I think that we want the JSON version: as a user, if I'm storing JSON, I would expect to be able to copy and paste values returned from pgcli into whatever JSON tools I'm using without having to go through an additional translation step from Python to JSON first. To get that behavior we need to register the typecaster regardless of the Python version, with the typecaster itself handling the spelling difference between Python 2 and Python 3.

amjith:
I understand now. 

I've updated the PR with a new commit. Does this look good?

drocco007:
Looks good to me! :+1:

PR #213, opened 2015-04-18
Added external editor command.

j-bennet:

PR #217, opened 2015-04-20
Iryna's Editor Command PR

amjith:
This PR is a cleanup of #213 to remove extraneous commits added in due to the repo move.

I've cleaned up the commit history. I should have waited to move the repo after merging all pending PRs. That was my bad. 

@j-bennet Doe this look good? 

I'll review the changes and merge it in.

### DIFF
--- a/pgcli/packages/iospecial.py
+++ b/pgcli/packages/iospecial.py
@@ -24,6 +24,7 @@ def open_external_editor(filename=None, sql=''):
     :return: list with one tuple, query as first element.
     """
 
+    sql = sql.strip().strip('\\e')
     message = None
     filename = filename.strip().split(' ', 1)[0] if filename else None
 
@@ -31,7 +32,7 @@ def open_external_editor(filename=None, sql=''):
 
     # Populate the editor buffer with the partial sql (if available) and a
     # placeholder comment.
-    query = click.edit(sql.strip().strip('\e') + '\n\n' + MARKER,
+    query = click.edit(sql + '\n\n' + MARKER,
             filename=filename)
 
     if filename:
@@ -46,6 +47,6 @@ def open_external_editor(filename=None, sql=''):
     else:
         # Don't return None for the caller to deal with.
         # Empty string is ok.
-        query = ''
+        query = sql
 
     yield (query, None, None, message)
--- a/pgcli/packages/iospecial.py
+++ b/pgcli/packages/iospecial.py
@@ -2,6 +2,7 @@ from __future__ import print_function
 
 import re
 import logging
+from codecs import open
 import click
 
 _logger = logging.getLogger(__name__)
@@ -44,12 +45,12 @@ def open_external_editor(filename=None, sql=''):
 
     # Populate the editor buffer with the partial sql (if available) and a
     # placeholder comment.
-    query = click.edit(sql + '\n\n' + MARKER,
-            filename=filename)
+    query = click.edit(sql + '\n\n' + MARKER, filename=filename,
+            extension='.sql')
 
     if filename:
         try:
-            with open(filename, "r") as f:
+            with open(filename, encoding='utf-8') as f:
                 query = f.read()
         except IOError:
             message = 'Error reading file: %s.' % filename
--- a/pgcli/packages/iospecial.py
+++ b/pgcli/packages/iospecial.py
@@ -14,8 +14,12 @@ def editor_command(command):
     """
     # It is possible to have `\e filename` or `SELECT * FROM \e`. So we check
     # for both conditions.
-    return command.strip().endswith('\e') or command.strip().startswith('\e')
+    return command.strip().endswith('\\e') or command.strip().startswith('\\e')
 
+def get_filename(sql):
+    if sql.strip().startswith('\\e'):
+        command, _, filename = sql.partition(' ')
+        return filename.strip() or None
 
 def open_external_editor(filename=None, sql=''):
     """
@@ -49,4 +53,4 @@ def open_external_editor(filename=None, sql=''):
         # Empty string is ok.
         query = sql
 
-    yield (query, None, None, message)
+    return (query, message)
--- a/pgcli/packages/iospecial.py
+++ b/pgcli/packages/iospecial.py
@@ -1,6 +1,6 @@
 from __future__ import print_function
 
-
+import re
 import logging
 import click
 
@@ -28,7 +28,15 @@ def open_external_editor(filename=None, sql=''):
     :return: list with one tuple, query as first element.
     """
 
-    sql = sql.strip().strip('\\e')
+    sql = sql.strip()
+
+    # The reason we can't simply do .strip('\e') is that it strips characters,
+    # not a substring. So it'll strip "e" in the end of the sql also!
+    # Ex: "select * from style\e" -> "select * from styl".
+    pattern = re.compile('(^\\\e|\\\e$)')
+    while pattern.search(sql):
+        sql = pattern.sub('', sql)
+
     message = None
     filename = filename.strip().split(' ', 1)[0] if filename else None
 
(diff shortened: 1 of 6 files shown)


j-bennet:
Looks good. Thanks! How did you do that? I could not figure it out!

amjith:
Steps to reproduce: 

```
# Clone your fork
$ git clone git@github.com:j-bennet/pgcli.git iryna_pgcli
$ cd iryna_pgcli

# Move the current working directory to the commit sha before you added your commits. 
$ git co cd1c9ea  

# Create a new branch from the current working directory state.
$ git co -b clean_state  

# Add in the main line repo as a new remote called upstream.
$ git remote add upstream git@github.com:dbcli/pgcli.git

# Bring in all the new changes from mainline. 
$ git pull upstream master

# Cherry-pick your commits on top. 
$ git cherry-pick 4315699
$ git cherry-pick a1b2d15

# Bask in the glory of success.
$ git log

# Push it to my own personal remote. My own fork is added in using the git remote add amjith-fork .....
$ git push amjith-fork fixed_branch

Open a PR from my fork to the mainline.
```

TADA!

There are many ways to achieve the same result, so if someone has a better way to do this, let me know.

pgcli/packages/pgspecial.py
```diff
@@ -1,6 +1,7 @@
 from __future__ import print_function
 import sys
 import logging
+import click
```
amjith: Pulling in click as a dependency seems wrong to me. I'll explain the reasons in a longer comment.

amjith:
Longer comment as promised: 

`pgspecial.py` should be treated as a standalone package without any dependencies. My goal is to make it into a separate package that can be installed and reused by other application authors. For example, I know that [Catherine Devline](https://twitter.com/catherinedevlin)  is interested in using pgspecial for her own database repl at some point. The original goal of that file is to implement the backslash commands that are specific to postgres database like `\dt`, \dv` etc. So pulling in `click` as a dependency is not a good idea. `\e` and `\c` are weird edge cases that are not really dealing with the database.

I'm sorry about sending this negative comment later on in the review. I should have caught this at the early stage. :(

I'm open to discussing this decision to keep pgspecial.py a standalone.

j-bennet:
Ok. I didn't know this was the intent. I agree that even though "external editor" is also a special command because it's a backslash command, it does not fit between the rest of them that follow the pattern of "run some specific Postgres queries and return data". I see two ways of solving this:

1) Pull this command out into main.py (which already has click dependency). This feels ugly, and I don't think that main.py should deal with commands at all.
2) Pull this command out into a new module (perhaps iocommands.py ?) that would be used by main.

What do you think?

amjith:
It's admirable to have a goal to keep main.py as clean as possible and not deal with commands at all. But I've already set the precedent (a wrong one) to deal with `\c`. So if you're feeling up to it, I'm all for moving it into a separate file and compartmentalizing it. So main will just make call to these packages.

j-bennet:
Then I would move both \c and \e there.

j-bennet:
@amjith So looks like \c has too much going on (setting internal state of pgexecute) to pull it out. But editor command is now separate from the pgspecial. Perhaps you can add https://github.com/j-bennet/pgcli/commit/9c215d25ea9d3d523929e6d58141e73b9f76fd99 to this PR.

pgcli/packages/iospecial.py
```diff
@@ -0,0 +1,45 @@
...
+
+_logger = logging.getLogger(__name__)
+
+
+def editor_command(command):
+    """
+    Is this an external editor command?
+    :param command: string
+    """
+    return command.strip().startswith('\e')
```
amjith: Let's change this to `endswith('\e')`.

pgcli/packages/iospecial.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def open_external_editor(filename=None):
```
amjith: Let's have this take an extra argument that is current partial query.

pgcli/packages/iospecial.py
```diff
@@ -0,0 +1,45 @@
...
+    Open external editor, wait for the user to type in his query,
+    return the query.
+    :return: list with one tuple, query as first element.
+    """
+
+    message = None
+    filename = filename.strip().split(' ', 1)[0] if filename else None
+
+    MARKER = '# Type your query above this line.\n'
+    query = click.edit('\n\n' + MARKER, filename=filename)
```
amjith: Feed the partial query to `click.edit()` so it is auto-populated into the editor buffer.

pgcli/pgexecute.py
```diff
@@ -189,6 +190,11 @@ def run(self, statement):
                 _logger.debug('Successfully switched to DB: %r', dbname)
                 yield (None, None, None, 'You are now connected to database "%s" as '
                         'user "%s"' % (self.dbname, self.user))
+            elif iospecial.editor_command(command):
+                tokens = sql.split(' ', 2)
+                filename = tokens[1] if len(tokens) > 1 and tokens[1] else None
+                for result in iospecial.open_external_editor(filename=filename):
```
amjith: Can you pass in the statement to this function as an arg? That way a partially typed query such as:

```
SELECT * FROM \e
```

will have `SELECT * FROM \e` passed in which will be used to populate the partial buffer.

pgcli/packages/iospecial.py
amjith: You might have to strip out the `\e` at the end before sending in the partial query to click.edit().

j-bennet:
@amjith I already looked into changing this to "endswith", and it's not as simple as it seems. I'll explain later if you like, but it seems that it would require some more significant changes in main module. I was thinking perhaps this (smaller) change can be implementation stage 1, and I can add a stage 2 later.

amjith:
I've added a new commit to do what I had in mind. You're right it wasn't as straight forward as I explained in the comments. 

So would yo mind reviewing the latest commit? I've checked it manually and it seems to work as expected.

j-bennet:
@amjith Interesting... I remember seeing some odd behavior when I changed this to "endswith", in multiline mode it didn't accept "select *\e", and wanted ";". I had to invent a workaround for that. I have no time to look now, but will check later.

j-bennet:
@amjith I don't see that weird behavior anymore, seems to work fine. I'll check on mac (only verified on linux), but I think you're good to go.

j-bennet:
screenshot 2015-04-22 16 49 12
I reproduced the problem that I was talking about on my macbook. I enter the query and \e, and instead of running the editor command, the CLI just enters a newline and expects more input. The weirdest thing is - it did not happen on my Linux Mint.

amjith:
That is a good catch. I'm surprised that it didn't happen in Linux. I know the reason why. I'll have to think about a good way to fix it.

amjith:
While adding the multi-line case I discovered another corner case for `\e`. 

Details: https://github.com/dbcli/pgcli/pull/217/files#diff-38f3aeda57742756afa677ca09dcea43R205

I've refactored the code. It probably needs more tests. But I'd appreciate another pair of eyes to check my work.

j-bennet:
@amjith I have two suggestions to this:

https://github.com/dbcli/pgcli/commit/587c5a3906fcc1ec2c48067c45107dd90ae2b8ff
https://github.com/dbcli/pgcli/commit/b3c5a3fb935e294a6342fc8c9504f5cd6744b350

pgcli/packages/iospecial.py
```diff
@@ -0,0 +1,64 @@
...
+    return the query.
+    :return: list with one tuple, query as first element.
+    """
+
+    sql = sql.strip()
+
+    # The reason we can't simply do .strip('\e') is that it strips characters,
+    # not a substring. So it'll strip "e" in the end of the sql also!
+    # Ex: "select * from style\e" -> "select * from styl".
+    pattern = re.compile('(^\\\e|\\\e$)')
```
amjith: Good catch. :+1:

amjith:
Thank you for working with me on this PR. It was one of the tough commands with various corner cases. 

:key:

PR #223, opened 2015-04-28
Have Tab autocomplete at cursor

### LINKED ISSUE #222: TAB makes certain completions disappear
Here is an example: `SELECT * FROM sch` matches three entries, one a schema name in the middle:

before_tab

After pressing `Tab`, the first selection disappears:

after_tab

drocco007:
Closes #222

### DIFF
--- a/pgcli/key_bindings.py
+++ b/pgcli/key_bindings.py
@@ -36,12 +36,13 @@ def pgcli_bindings(vi_mode=False):
         _logger.debug('Detected F4 key.')
         key_binding_manager.enable_vi_mode = not key_binding_manager.enable_vi_mode
 
+    @key_binding_manager.registry.add_binding(Keys.Tab)
     @key_binding_manager.registry.add_binding(Keys.ControlSpace)
     def _(event):
         """
         Force autocompletion at cursor.
         """
-        _logger.debug('Detected <C-Space> key.')
+        _logger.debug('Detected <C-Space>/<Tab> key.')
         event.cli.current_buffer.complete_next()
 
     return key_binding_manager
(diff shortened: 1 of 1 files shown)


amjith:
This is an interesting one. I still don't understand what dictates the Tab key. I presume that prompt_toolkit has some default behavior for the Tab key.

But anyway, the interesting bit is, the Tab key right now acts like how we expect it to behave in bash which is if there are multiple entries with same prefix then it will fill in the portion of the prefix that is common. 

But with the new PR that is gone, hitting the Tab key will automatically select the first entry in the suggestion menu. 

I think this is the artifact of using substring matching instead of prefix matching. I'm perfectly happy with merging the PR and losing the common prefix completion since that is a bit counter-intuitive. 

Any thoughts before I merge it in? 

/cc @dbcli/pgcli-core

darikg:
I think in linux the autosuggestion menu shows up constantly and automatically, right? On windows it doesn't, and the tab key displays the suggestion menu, without filling in any text, whereas control-space shows the suggestions menu, and fills in the first suggestion. So I use the two different keys in different contexts, and would dislike how this pr would take that away. Maybe a reasonable compromise would be to reverse the control-space and tab mappings, instead of mapping them to the same thing?

drocco007:
@amjith  yes, I think you're right that that is the default `Tab` behavior in `prompt_toolkit`, which (almost) serves to explain what's happening in the issue I originally reported. Without changing the default, the initial set of completions includes substring matches, but hitting `Tab` then expands the common prefix (`schema_information`), eliminating the first entry. I'm not sure why that first entry isn't included in the common prefix calculation, but it's probably either a bug in `prompt_toolkit` or in how we are providing completion entries to it.

I think @darikg's proposed solution is reasonable, incoming update that implements it. :) I don't have a Windows machine to test on, so please take a look and let me know if this change meets your expectations.

amjith:
I'll leave this one for @darikg to test and merge.

darikg:
:+1: Thanks @drocco007!

PR #224, opened 2015-04-28
fix broken suggestions after compound where clause

darikg:
Previously, 'SELECT * FROM foo WHERE ' correctly suggested columns, but 'SELECT * FROM foo WHERE bar = 1 AND ' incorrectly suggested keywords. This was due to the fact that sqlparse was grouping the entire where clause into a single Where token.

This seems such a simple fix I feel like I'm missing something! Suggestions on any extra tests here would be appreciated

### DIFF
--- a/pgcli/packages/parseutils.py
+++ b/pgcli/packages/parseutils.py
@@ -151,6 +151,7 @@ def extract_tables(sql):
     stream = extract_from_part(parsed[0], stop_at_punctuation=insert_stmt)
     return list(extract_table_identifiers(stream))
 
+
 def find_prev_keyword(sql):
     """ Find the last sql keyword in an SQL statement
 
@@ -161,10 +162,26 @@ def find_prev_keyword(sql):
         return None, ''
 
     parsed = sqlparse.parse(sql)[0]
-    for t in reversed(list(parsed.flatten())):
-        if t.is_keyword or t.value == '(':
-            idx = parsed.token_index(t)
-            text = ''.join(tok.value for tok in parsed.tokens[:idx+1])
+    flattened = list(parsed.flatten())
+
+    trivial_keywords = ('AND', 'OR', 'NOT', 'BETWEEN')
+
+    for t in reversed(flattened):
+        if t.value == '(' or (t.is_keyword and (
+                              t.value.upper() not in trivial_keywords)):
+            # Find the location of token t in the original parsed statement
+            # We can't use parsed.token_index(t) because t may be a child token
+            # inside a TokenList, in which case token_index thows an error
+            # Minimal example:
+            #   p = sqlparse.parse('select * from foo where bar')
+            #   t = list(p.flatten())[-3]  # The "Where" token
+            #   p.token_index(t)  # Throws ValueError: not in list
+            idx = flattened.index(t)
+
+            # Combine the string values of all tokens in the original list
+            # up to and including the target keyword token t, to produce a
+            # query string with everything after the keyword token removed
+            text = ''.join(tok.value for tok in flattened[:idx+1])
             return t.value, text
 
     return None, ''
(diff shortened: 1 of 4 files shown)


amjith:
Additional tests: 

```
SELECT * FROM foo WHERE bar = 10 OR <tab>
```

This might be a harder case to satisfy, so I'm fine if this is not handled
```
SELECT * FROM foo WHERE (bar > 10 AND <tab>
```

This is a weird case, but it should suggest cols. 

```
SELECT * FROM foo WHERE 10 < 
```

darikg:
Thanks! I got all of those working, but found another one related to subqueries that breaks:
  `select * from foo where exists(`
which should suggest keywords, in order to do a subselect.

This one's even tricker:
  `select foo from bar where baz in (`
which should suggest either subqueries or columns, because both of these are legal:
  `select a from b where c in (select d from e)`
  `select a from b where c in (d, e)`

I think the solution to these intersects with the solution to the easier ones above, so I'll keep picking away at it.

darikg:
Ok, I think I got it finally! See tests for all the cases handled currently. One inchoate problem is that `suggest_based_on_last_token` is starting to get pretty long -- I'm not sure to handle that beyond breaking it up into a bunch of subfunctions like `suggest_when_last_token_ends_with_open_parens` et al.

pgcli/packages/parseutils.py
```diff
@@ -161,10 +162,26 @@ def find_prev_keyword(sql):
         return None, ''
 
     parsed = sqlparse.parse(sql)[0]
-    for t in reversed(list(parsed.flatten())):
-        if t.is_keyword or t.value == '(':
-            idx = parsed.token_index(t)
-            text = ''.join(tok.value for tok in parsed.tokens[:idx+1])
+    flattened = list(parsed.flatten())
+
+    trivial_keywords = ('AND', 'OR', 'NOT', 'BETWEEN')
```
amjith: Let's call this logical_operators. 2 months down the line that list won't seem so trivial. :P

pgcli/packages/sqlcompletion.py
```diff
@@ -135,23 +134,69 @@ def suggest_special(text):
...
+                return [{'type': 'keyword'}]
+            elif prev_tok in ('any', 'some', 'all'):
+                return column_suggestions + [{'type': 'keyword'}]
+            elif prev_tok == 'in':
+                # Technically, we should suggest columns AND keywords, as
+                # per case 4. However, IN is different from ANY, SOME, ALL
+                # in that it can accept a *list* of columns, or a subquery.
+                # Because of the way lists of columns are reduced in this
+                # function, "SELECT * FROM foo WHERE bar IN (baz, qux, " would
+                # also suggest keywords, which is kinda gross.
```
amjith: This comment is talking about a case that was remedied by adding this code. It's a bit misleading in that sense, when I first read it, I thought "SELECT * FROM foo WHERE bar IN (" would suggest both columns and keywords. But that's not the case. It only suggests columns. Can we edit the comment to make that a bit explicit.

amjith:
I'm not quite done with the full review, but the initial perusal looks very promising. I'll get this reviewed tonight and merge in.

amjith:
I've updated the branch with the edits that I recommended. I'm happy with the changes in the PR.

@darikg You're welcome to spot check my commit and merge it in.

drocco007:
I just noticed that there is an interaction between this branch and #230… I'm sorry I didn't notice this sooner… pitfalls of distributed, concurrent development I guess :}

My approach to fixing find_prev_keywords is different from @darikg's, so we should compare notes and see which bits we want to preserve from each. If nothing else, we probably want to keep the new test cases from my PR. But have a look and let's discuss! :)

amjith:
@darikg and @drocco007 I hope you're not waiting on me. I'm perfectly fine with the changes in this PR. 

If you're good then one of you can merge it in.

PR #230, opened 2015-05-05
Ignore identifier lists and 'AS' when finding SQL keywords.

### LINKED ISSUE #228: Column suggestion in 2nd and subsequent columns explodes if entered text is a keyword
Many valid column names start with a prefix that is a complete keyword; this is especially true for the shorter keywords (e.g. AS, IN, IS, TO). However, if the entered text in the second or subsequent column of a SELECT clause is a complete keyword, pgcli raises an exception.

Example: enter text `SELECT 1, in`

```
postgres> SELECT 1, inException in thread Thread-12:
  File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 763, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/home/drocco/.envs/brighttrac/lib/python2.7/site-packages/prompt_toolkit/__init__.py", line 472, in run
    CompleteEvent(text_inserted=True)))
  File "/home/drocco/source/foss/pgcli/pgcli/pgcompleter.py", line 201, in get_completions
    suggestions = suggest_type(document.text, document.text_before_cursor)
  File "/home/drocco/source/foss/pgcli/pgcli/packages/sqlcompletion.py", line 86, in suggest_type
    full_text, identifier)  [F3] Multiline: OFF  [F4] Emacs-mode                                                                                                                                                                    
  File "/home/drocco/source/foss/pgcli/pgcli/packages/sqlcompletion.py", line 236, in suggest_based_on_last_token
    prev_keyword, text_before_cursor = find_prev_keyword(text_before_cursor)
  File "/home/drocco/source/foss/pgcli/pgcli/packages/parseutils.py", line 166, in find_prev_keyword
    idx = parsed.token_index(t)
  File "/home/drocco/.envs/brighttrac/lib/python2.7/site-packages/sqlparse/sql.py", line 343, in token_index
    return self.tokens.index(token)
ValueError: <Keyword 'in' at 0x7f35828f21b0> is not in list
```

Testing against f0dcb38, PG 9.3.5

drocco007:
Fixes #228

### DIFF
--- a/pgcli/packages/parseutils.py
+++ b/pgcli/packages/parseutils.py
@@ -151,6 +151,30 @@ def extract_tables(sql):
     stream = extract_from_part(parsed[0], stop_at_punctuation=insert_stmt)
     return list(extract_table_identifiers(stream))
 
+
+def is_free_keyword(token):
+    """Does this token represent a keyword that is not part of an identifier?
+
+    For example, in the fragment `SELECT 1, in`, `SELECT` is a free keyword but
+    `in` is not since it appears as part of an identifier (the column list).
+
+    `AS` is never considered to be a free keyword, nor are any tokens that are
+    not keywords (that is, for which `token.is_keyword` is False).
+
+    Returns True if `token` represents a keyword outside of an identifier,
+    False otherwise.
+
+    """
+
+    # sqlparse doesn't set the AS keyword's parent for some reason, so checking
+    # token.within() doesn't work. Since AS always appears within an identifier
+    # we treat it as a special case.
+    if token.value.upper() == 'AS':
+        return False
+    else:
+        return token.is_keyword and not token.within(IdentifierList)
+
+
 def find_prev_keyword(sql):
     """ Find the last sql keyword in an SQL statement
 
@@ -162,7 +186,7 @@ def find_prev_keyword(sql):
 
     parsed = sqlparse.parse(sql)[0]
     for t in reversed(list(parsed.flatten())):
-        if t.is_keyword or t.value == '(':
+        if is_free_keyword(t) or t.value == '(':
             idx = parsed.token_index(t)
             text = ''.join(tok.value for tok in parsed.tokens[:idx+1])
             return t.value, text
--- a/tests/test_sqlcompletion.py
+++ b/tests/test_sqlcompletion.py
@@ -331,3 +331,26 @@ def test_handle_pre_completion_comma_gracefully(text):
     suggestions = suggest_type(text, text)
 
     assert iter(suggestions)
+
+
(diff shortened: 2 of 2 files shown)


drocco007:
That `.format` thing gets me every time… :/

pgcli/packages/parseutils.py
```diff
@@ -151,6 +151,30 @@ def extract_tables(sql):
...
+
+    Returns True if `token` represents a keyword outside of an identifier,
+    False otherwise.
+
+    """
+
+    # sqlparse doesn't set the AS keyword's parent for some reason, so checking
+    # token.within() doesn't work. Since AS always appears within an identifier
+    # we treat it as a special case.
+    if token.value.upper() == 'AS':
```
amjith: Should we send a PR upstream to sqlparse to treat `AS` as a keyword?

We can patch it in the meantime here by add a comment that says, remove the special case in the future when sqlparse is fixed.

pgcli/packages/parseutils.py
```diff
@@ -162,7 +186,7 @@ def find_prev_keyword(sql):
 
     parsed = sqlparse.parse(sql)[0]
     for t in reversed(list(parsed.flatten())):
-        if t.is_keyword or t.value == '(':
+        if is_free_keyword(t) or t.value == '(':
             idx = parsed.token_index(t)
```
amjith: I understand the fix now, but I'm very nervous now, here's why. 

We are iterating through the flattened tokens and when a condition is met, we look up the index of that token in the unflattened token list. Which does not sound right. 

Here's an easy and completely plausible way to break it:

`SELECT * FROM abc WHERE a,`

This is broken even before this PR. So that call to token_index() is just a ticking time-bomb. 

Tagging @darikg since his change relies on that index being available.

Is there an alternate way to get the same result?

darikg: i fixed that in #224 -- if this gets merged first it will be easy to update

amjith: Yup. You're right. My bad. I'll get that merged in soon. 

@drocco007 Can you try out #224. It looks like the crash is fixed in that branch. Which obviates this additional check for is_free_keyword(). 

But I like to keep the tests.

drocco007:
Gah! Sorry for the crosstalk, github emails stopped showing up in my inbox… :confused: 

Yes, I'll take a look at this and let you know if it fixes the cases I noticed.

darikg:
@drocco007 when you get a chance can you rebase this? Like you said, we should definitely keep the tests

amjith:
Closing this PR as it is outdated.

PR #231, opened 2015-05-08
Fix incorrect super() calls

drocco007:
From the “not a problem now but could be later” department, this PR is preventative maintenance to fix some incorrect calls to `super()`. The spelling

```python

super(self.__class__, self)…
```

is a subtle error that unfortunately tends to manifest itself long after the call is written. To see why it's a problem, consider this class hierarchy:

```python

class A(object):
    def report(self):
        print 'A reporting'

class B(A):
    def report(self):
        print 'B reporting'
        super(self.__class__, self).report()

class C(B):
    def report(self):
        print 'C reporting'
        super(self.__class__, self).report()

C().report()
```

The expected output is

    C reporting
    B reporting
    A reporting

but you actually get a stack overflow. The first `super()` call from `C` to `B` works as expected, but in `B.report()`, when called from `C`, `self.__class__` is `C`, so `super(self.__class__, self)` == `super(C, self)` == `B`! This is correct from a semantic point of view but almost certainly not what was intended; the result is that `B.report()` keeps calling itself recursively until stack space is exhausted.

### DIFF
--- a/pgcli/pgbuffer.py
+++ b/pgcli/pgbuffer.py
@@ -4,7 +4,7 @@ class PGBuffer(Buffer):
     def __init__(self, always_multiline, *args, **kwargs):
         self.always_multiline = always_multiline
         is_multiline = lambda doc: self.always_multiline and not _multiline_exception(doc.text)
-        super(self.__class__, self).__init__(*args, is_multiline=is_multiline, **kwargs)
+        super(PGBuffer, self).__init__(*args, is_multiline=is_multiline, **kwargs)
 
 def _multiline_exception(text):
     text = text.strip()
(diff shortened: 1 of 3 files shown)


amjith:
Wow! I didn't think about that intricacy. I assumed that doing it this way will make it easy to change the name of the class in the future without having to change every instance of that name. 

I swear I saw that pattern in one of Raymond's talks. I'll have to go back and find it.

Thanks! :beetle:

j-bennet:
@drocco007 @amjith This blew my mind but apparently it's a well known issue! http://stackoverflow.com/questions/4235078/how-to-avoid-infinite-recursion-with-super. But looks like this is much easier in Python 3.

PR #233, opened 2015-05-15
Amjith/general cleanup

amjith:
Minor issues discovered while I was working on the mysql-cli which should be fixed in pgcli as well. 

* The test harness setup was done incorrectly. This should enable everyone who has postgres to run our tests without skipping. 
* Exceptions in a generator function aren't thrown right away, they're only thrown when we consume the generator. Hence a new addition of try/except to suppress the exception and prevent crashing pgcli if one of the result sets had a error. 
* Minor efficiency improvement by moving the `all_completion` update into the for loop instead of a separate list comprehension. 

/cc @drocco007 Can you review and merge? 

I'm going to start tagging individuals (picked randomly) to review PRs instead of tagging the entire core dev team.

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -317,8 +317,8 @@ class PGCli(object):
                         for title, cur, _, _ in res:
                             if hasattr(cur, 'close'):
                                 cur.close()
-                    except Exception:
-                        pass
+                    except Exception as e:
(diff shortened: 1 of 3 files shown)


j-bennet:
@amjith no logging of the exception? only suppressing it?

amjith:
Fair point. Added a new commit to log the exception. :)

PR #234, opened 2015-05-15
Remove unnecessary finally clause.

amjith:
@j-bennet

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -310,15 +310,6 @@ class PGCli(object):
                     if pgspecial.TIMING_ENABLED:
                         print('Command Time:', duration)
                         print('Format Time:', total)
-                finally:
-                    # Ignore any errors thrown when consuming the generator
-                    # 'res'.
-                    try:
-                        for title, cur, _, _ in res:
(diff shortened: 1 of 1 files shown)


j-bennet:
@amjith Is it possible that an exception was thrown and we got out of the loop before the generator was exhausted?

j-bennet:
@amjith Also... do you think it would be useful to we add a context manager wrapper to handle cursor closing?

amjith:
1. It is possible that an exception was thrown before the generator is exhausted. But if that happens then we can't continue to consume the generator anyway because whatever exception caused it to break out of the loop will prevent further consumption. 

2. There is already a context manager present in psycopg2 for the cursors, but we can't use it in this case because the cursor is created in pgexecute.py but it is returned to main.py where it is consumed. Since the cursor has to stay open across two different functions (one in pgexecute and one is main) we can't use the context manager. Otherwise the cursor will be closed by pgexecute and the closed cursor will be handed to main which can't do anything with it.

j-bennet:
Got it. Thanks for explaining.

PR #235, opened 2015-05-15
Fix for listing triggers on \d special.

droustchev:
Small fix to show the actual triggerdef, instead of the position of the triggerdef after "CREATE TRIGGER ".

### DIFF
--- a/pgcli/packages/pgspecial.py
+++ b/pgcli/packages/pgspecial.py
@@ -781,7 +781,7 @@ def describe_one_table_details(cur, schema_name, relation_name, oid, verbose):
                     if triggerpos >= 0:
                         tgdef = triggerpos + 9;
 
-                    status.append("    %s\n" % tgdef);
+                    status.append("    %s\n" % row[1][tgdef:])
 
     #/*
     #* Finish printing the footer information about a table.


droustchev:
Btw, my editor throws a lot of pylint warnings (e.g. for the semicolons). However, I did not want to go ahead and cause a lot of formatting changes. Maybe it would be nice to have a .editorconfig? I would be happy to add one. Otherwise, I love pgcli! Thanks a lot for all your effort!

amjith:
@droustchev Thank you for the PR. 

Do you have a small example that exercises this new change? 

For example an output before and after this change would be nice. 

I love fixes to pylint errors and a .editorconfig would be a welcome addition. :)

amjith:
I just tested it locally. That was a good catch. 

Thank you for the PR. I look forward to the pylint cleanup and a .editorconfig addition. :)

:radio:

droustchev:
Sorry, I was in a meeting. Guess you don't need the example anymore? I will try to add a .editorconfig as soon as possible. What is your default indentation setting? I had some mismatch there with mine (4 spaces for python). Thanks for merging my PR so quickly, btw and again also gib thanks for this little gem you coded up! Really enjoy working with pgcli :)

amjith:
I do use 4 spaces. The pgspecial.py file is an oddity. 

It was a literal translation of psql's C-code to python code. Some of the formatting might be odd because I was copying some code and moving things around. So a cleanup to that file would be fantastic. :)

PR #239, opened 2015-05-22
`drop schema` and `create schema` should suggest schemas

darikg:


amjith:
:+1: 

Nice and simple.

PR #240, opened 2015-05-23
Suggest datatypes

darikg:
1. The following contexts now suggest datatypes:
  - `SELECT foo::<TAB>`
  - `CREATE TABLE foo (bar <TAB>`
  - `CREATE FUNCTION foo (bar <TAB>`
  - `ALTER TABLE foo ALTER COLUMN bar TYPE <TAB>`

2. The special command `\dT` is now supported, and suggests types (but not tables)

Datatypes are suggested both from a hardcoded whitelist of builtin postgresql types, as well as user-defined custom types. Furthermore, since postgresql treats a table row as a form of composite type, table names are suggested in addition to data types.

The whitelist I just typed from memory of datatypes I use frequently, and would appreciate feedback on what belongs there. (Or we can just use all of them.) You should be able to use `\dT pg_catalog.*` to list all of the built-in types.

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -363,6 +363,9 @@ class PGCli(object):
         # functions
         completer.extend_functions(pgexecute.functions())
 
+        # types
+        completer.extend_datatypes(pgexecute.datatypes())
+
         # databases
         completer.extend_database_names(pgexecute.databases())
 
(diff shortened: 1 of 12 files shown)


amjith:
Thank you for adding this feature. It works great. :)

Regarding the list of builtin datatypes, there are only 82 types, let's add them all the list. I'll merge this PR now and you're welcome to add the rest in a future PR. 

:jack_o_lantern:

PR #241, opened 2015-05-24
Fix editor mode

fpietka:
It seems there was a leftover from prompt-toolkit migration.

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -183,7 +183,8 @@ class PGCli(object):
             if message:
                 # Something went wrong. Raise an exception and bail.
                 raise RuntimeError(message)
-            document = Document(sql, cursor_position=len(sql))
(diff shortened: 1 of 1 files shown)


amjith:
Good catch. Editor mode does indeed seem broken. 

But this PR doesn't quite fix it. For example, after I enter a query in the editor and close the editor, the behavior should be the prompt will be filled in with the query in the editor, but not actually send it to the server. 

This is to give the user another chance to edit the query before pressing enter.

fpietka:
Sorry, didn't quite catch that. It is now fixed.

Good luck on mysql version! I'm now "forced" to use mysql, so I'll be closely watching it :grin:

amjith:
Thank you for the quick turn around. I appreciate the initiative. 

I've sent you an invite to the mycli repo. Hopefully I would have fixed the editor command there before you get a chance to try it out. ;)

:beers:

PR #244, opened 2015-05-27
Saving named queries

brettatoms:
This is the beginning of a patch for saving named queries in pgcli.  Named queries allow you to save and re-execute a query with a short name.  E.g.

```
; Save a query
\ns simple_query select 1;

; Execute saved query
\n simple_query

; List all saved query names
\n

; List all saved query names and the queries
\n+

; Delete saved query
\nd simple_query
```

There's a couple shortcoming that should probably be addressed before merging this PR:

1. Queries are saved in `~/.pgcli`.  This destroys any formatting and comments in the
config file.   It could be fixed by either saving the queries in a separate file or replacing `ConfigParser` with `ConfigObj` which should preserve the formatting and comments.
2. Autocomplete saved queries after entering `\n` and as you type. 
3. Tests

### DIFF
--- a/pgcli/packages/namedqueries.py
+++ b/pgcli/packages/namedqueries.py
@@ -4,10 +4,9 @@ from ..config import load_config
 class NamedQueries:
 
     section_name = 'named queries'
-    filename = '~/.pgclirc'
 
-    def __init__(self):
-        self.filename = expanduser('~/.pgclirc')
+    def __init__(self, filename):
+        self.filename = expanduser(filename)
         self.config = load_config(self.filename)
 
     def list(self):
@@ -36,4 +35,4 @@ class NamedQueries:
 
 
 
-namedqueries = NamedQueries()
+namedqueries = NamedQueries('~/.pgclirc')
--- a/pgcli/packages/namedqueries.py
+++ b/pgcli/packages/namedqueries.py
@@ -6,33 +6,24 @@ class NamedQueries:
     section_name = 'named queries'
 
     def __init__(self, filename):
-        self.filename = expanduser(filename)
-        self.config = load_config(self.filename)
+        self.config = load_config(filename)
 
     def list(self):
-        if self.config.has_section(self.section_name):
-            return self.config.options(self.section_name)
-        return []
+        return self.config.get(self.section_name, [])
 
     def get(self, name):
-        if not self.config.has_section(self.section_name):
-            return None
-        return self.config.get(self.section_name, name)
+        return self.config.get(self.section_name, {}).get(name, None)
 
     def save(self, name, query):
-        if not self.config.has_section(self.section_name):
-            self.config.add_section(self.section_name)
-        self.config.set(self.section_name, name, query)
-        self._write()
+        if self.section_name not in self.config:
+            self.config[self.section_name] = {}
+        self.config[self.section_name][name] = query
+        self.config.write()
 
     def delete(self, name):
-        self.config.remove_option(self.section_name, name)
-        self._write()
-
-    def _write(self):
-        with open(self.filename, 'w') as f:
-            self.config.write(f)
-
+        if self.section_name in self.config and name in self.config[self.section_name]:
+            del self.config[self.section_name][name]
+            self.config.write()
 
 
 namedqueries = NamedQueries('~/.pgclirc')
--- a/pgcli/packages/namedqueries.py
+++ b/pgcli/packages/namedqueries.py
@@ -1,7 +1,6 @@
-from os.path import expanduser
 from ..config import load_config
 
-class NamedQueries:
+class NamedQueries(object):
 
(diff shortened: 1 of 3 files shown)


amjith:
This is a cool feature. I like it. :)

I noticed that tests are failing. I'll leave some inline comments on how to fix them.

pgcli/config.py
```diff
@@ -9,7 +9,7 @@
 
 def load_config(filename, default_filename=None):
     filename = expanduser(filename)
-    parser = ConfigParser()
+    parser = ConfigParser(interpolation=None)
```
amjith: Let's replace this with `RawConfigParser()` because ConfigParser in Python 2 does not take a keyword arg `interpolation`.

pgcli/packages/namedqueries.py
```diff
@@ -0,0 +1,39 @@
+from os.path import expanduser
+from ..config import load_config
+
+class NamedQueries:
+
+    section_name = 'named queries'
+    filename = '~/.pgclirc'
+
+    def __init__(self):
+        self.filename = expanduser('~/.pgclirc')
```
amjith: I would prefer to remove the hard coding of ~/.pgclirc. Instead have the constructor take in a file path as it's argument. 

```
def __init__(self, filename):
    self.filename = expanduser(filename)
```

pgcli/packages/namedqueries.py
```diff
@@ -0,0 +1,39 @@
...
+        self.config.remove_option(self.section_name, name)
+        self._write()
+
+    def _write(self):
+        with open(self.filename, 'w') as f:
+            self.config.write(f)
+
+
+
+namedqueries = NamedQueries()
```
amjith: This will have to change to pass in `~/.pgclirc` as an arg.

pgcli/packages/pgspecial.py
```diff
@@ -1025,6 +1026,44 @@ def toggle_timing(cur, arg, verbose):
...
+    namedqueries.delete(arg)
+    return [(None, None, None, "Deleted.")]
+
+def execute_named_query(cur, arg, verbose):
+    """Returns (title, rows, headers, status)"""
+    if arg == '':
+        return list_named_queries(cur, arg, verbose)
+
+    query = namedqueries.get(arg)
+    title = '> {}'.format(query)
```
amjith: This type of string interpolation will fail in Python 2.6. Instead use

```
title = '> %s' % query
```

pgcli supports Python 2.6, 2.7, 3.3 and 3.4.

pgcli/packages/pgspecial.py
```diff
@@ -1025,6 +1026,44 @@ def toggle_timing(cur, arg, verbose):
...
+    if query is None:
+        message = "No named query: {}".format(arg)
```
amjith: Same as above.

amjith:
I've left some comments inline from a cursory code review. I haven't had a chance to pull this down and try it out. I'll do that later today when I get home. 

I appreciate you taking the time to send in the patch. 

/cc @dbcli/pgcli-core

brettatoms:
Fixed the py2 compatibility and the hardcoded config filename.  

What do you think about using `~/.pgclirc` for saving the named queries?  When you save a named query it destroys the comments and formatting.  If you want to keep the formatting and comments we should either use ConfigObj or save them to a separate file like `~/.pgclirc-namedqueries` or something less ugly.

amjith:
I didn't know about ConfigObj. I didn't know that a library exists for doing this. I love it. 

Using ConfigObj will pave way for a couple of other ideas I had which require reading and writing config files with the comments intact.

If you'd like to modify `pgcli` to use ConfigObj, I'd be totally open to merging it in. If that's too much to handle right now, you're welcome to use a different config file like `~/.pgcli-namededqueries`. 

It is completely up to you. :)

amjith:
@brettatoms I have a few suggestions to polish this feature. If you'd rather get this merged and polish this later. I can save my comments for later and come back to it. If you're not yet tired of the back and forth, I can provide you more feedback. 

Do you have a preference?

amjith:
This look good. I'm going to merge it in and write some tests. 

Thank you for extending pgcli. I like this feature :). 

:tanabata_tree:

brettatoms:
@amjith I'm all for adding polish and improving the feature.  I think this could be used for some really cool stuff like using bind parameters or python string formatting to make queries or dynamic or even using config file interpolation to allow you to embed named queries in other named queries.

amjith:
It's great that you have bigger visions of this feature. I'd highly encourage you to keep plugging away. 

The polish I had in mind was to see if you can get autocompletion working for `\n`. 

Right now `\n` doesn't print anything if there are no saved queries. It'll be good to print something reasonable when there are no saved queries. 

Finally add some tests so we don't accidentally break them in the future.

PR #245, opened 2015-05-28
replace configparser with configobj

brettatoms:


pgcli/config.py
```diff
@@ -1,23 +1,17 @@
...
 def load_config(filename, default_filename=None):
     filename = expanduser(filename)
-    parser = ConfigParser()
+    config = ConfigObj(filename, interpolation=False)
 
-    # parser.read will not fail in case of IOError,
-    # so let's not try/except here.
     if default_filename:
-        parser.read(default_filename)
+        config.merge(load_config(default_filename))
```
amjith: This has changed the order of precedence. The original code would read the default config then apply the overrides from the user config on top. 

But this change here makes it the other way around. So the default settings will always take precedence which is not a desirable behavior.

amjith:
The code looks good except for that reversal of the precedence order. ConfigObj looks much cleaner, so I'm happy with the decision. 

If you can fix the order of merging, I'll be happy to merge this in and thanks for telling me about ConfigObj. This is a great find. :)

brettatoms:
Good catch on the precedence order.  Should be fixed now.

amjith:
Looks good. :+1: 

Thank you for the PR and the subsequent fixes. 

:bee:

PR #246, opened 2015-05-28
Fix error for (\d <table_name> command) if table has rules

sancopanco:
error like "coercing to Unicode: need string or buffer, int found"


amjith:
Thank you for the fix. Changes look good. 

:mushroom:

PR #248, opened 2015-05-30
Simplify config file handling.

amjith:
Reduce the complexity of config files. 

Reviewer: @j-bennet

### DIFF
--- a/pgcli/config.py
+++ b/pgcli/config.py
@@ -5,7 +5,7 @@ from configobj import ConfigObj
 def load_config(usr_cfg, def_cfg=None):
     cfg = ConfigObj()
     cfg.merge(ConfigObj(def_cfg, interpolation=False))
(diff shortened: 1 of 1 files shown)


j-bennet:
@amjith It doesn't need an expanduser call anymore?

amjith:
Good catch. It does need the expanduser call. I'll update the PR.

amjith:
Updated.

j-bennet:
So, default config can be None. In that case, things won't break, but you'll be creating and merging an unnecessary empty ConfigObj. I think adding an "if" is better, even though it makes the code less pretty.

amjith:
I tested the None case and made a conscious decision to create an empty ConfigObj. Objects are cheap and I value code clarity. :)

j-bennet:
Fair enough.

PR #252, opened 2015-05-31
Upgrade to prompt_toolkit==0.38

jonathanslenders:
Upgrade to the newest version of prompt_toolkit. This should also fix the issue with slow pasting.

Feel free to make changes to the pull request before merging.
I'm not sure how you are feeling about the circular reference between pgcli_bindings and the main class. It can be a weakref, but it doesn't have to.

### DIFF
--- a/pgcli/key_bindings.py
+++ b/pgcli/key_bindings.py
@@ -1,14 +1,19 @@
 import logging
 from prompt_toolkit.keys import Keys
 from prompt_toolkit.key_binding.manager import KeyBindingManager
+from prompt_toolkit.filters import Condition
 
 _logger = logging.getLogger(__name__)
 
-def pgcli_bindings(vi_mode=False):
+def pgcli_bindings(get_vi_mode_enabled, set_vi_mode_enabled):
     """
     Custom key bindings for pgcli.
     """
-    key_binding_manager = KeyBindingManager(enable_vi_mode=vi_mode)
+    assert callable(get_vi_mode_enabled)
+    assert callable(set_vi_mode_enabled)
+
+    key_binding_manager = KeyBindingManager(
+        enable_vi_mode=Condition(lambda cli: get_vi_mode_enabled()))
 
     @key_binding_manager.registry.add_binding(Keys.F2)
     def _(event):
@@ -34,7 +39,7 @@ def pgcli_bindings(vi_mode=False):
         Toggle between Vi and Emacs mode.
         """
         _logger.debug('Detected F4 key.')
-        key_binding_manager.enable_vi_mode = not key_binding_manager.enable_vi_mode
+        set_vi_mode_enabled(not get_vi_mode_enabled())
 
     @key_binding_manager.registry.add_binding(Keys.Tab)
     def _(event):
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -10,12 +10,12 @@ from time import time
 
 import click
 import sqlparse
-from prompt_toolkit import CommandLineInterface, AbortAction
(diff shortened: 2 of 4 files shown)


amjith:
I've tested this locally and it works as expected. Can the @dbcli/pgcli-core team please take it for a spin?

@darikg Can you verify that things work as expected in Windows? 

* Paste is now fast. 
* Haven't tried the fuzzy completion branch, but I'll try it soon to make sure that it doesn't crash when the entered word is longer than the suggestion.

j-bennet:
You missed one @jonathanslenders:

https://github.com/jonathanslenders/pgcli/blob/prompt_toolkit_0.38/pgcli/main.py#L187

jonathanslenders:
Good catch @j-bennet . I missed that one.

Pull request updated.

j-bennet:
@jonathanslenders: I do have doubts, like you said, about passing in the whole pgcli into pgcli_bindings call, as opposed to passing in only what's necessary. What was the reasoning for doing it this way?

jonathanslenders:
@j-bennet: The reasoning is that ``KeyBindingManager`` became stateless. It doesn't know anymore whether ``vi_mode`` is on or off by itself. It takes a ``vi_mode`` filter instance as input. In our case, that's this condition: ``Condition(lambda cli: pgcli.vi_mode)``. It's something that evaluates to a boolean when it's called, and in our case it evaluates by taking the value of ``pgcli.vi_mode``.

So what happened is that now the actual state of ``vi_mode`` is only present in the ``pgcli`` class. (I think this is what we call some simplified kind of reactive programming.)  Previously, we also had the ``vi_mode`` variable inside ``pgcli``, but that variable only contained the initial state for the manager. Every following change only updated the ``vi_mode`` variable in the manager, not in ``pgcli``. So, there we had some inconsistency.

We have two options. Either we keep this state somewhere in a settings object, and we pass that around to the code that needs to change it. Or we keep it in ``pgcli``, but pass a function to retrieve and set it. e.g.:

```python
def set_vi_mode(value):
    self.vi_mode = value

... = pgcli_bindings(get_vi_mode=lambda: self.vi_mode,
                               set_vi_mode=set_vi_mode)

```

What do you think?

j-bennet:
@jonathanslenders I was going to suggest passing in a function. I like the lambda approach in your code snippet.

jonathanslenders:
@j-bennet Pull request updated.

j-bennet:
Looks great now @jonathanslenders ! Thank you for being so responsive. I will test it some more tonight to make sure nothing else is missing / breaks.

j-bennet:
Tested again and don't see any problems. Pasting a large query is a hundred times faster (this is the query I used: https://gist.github.com/j-bennet/fac59560b017d7d77b76). I think it's good to merge when @darikg has a chance to verify things work on Windows.

amjith:
Just tested the fuzzy completion branch. It works beautifully. :)

For the curious souls, you can check out the branch https://github.com/dbcli/pgcli/tree/amjith/fuzzy_match

You will have to run `pip install -e . ` after you pull to install new dependencies.

I'd like to have a bigger discussion about this feature before making it a default. So treat it as purely experimental.

/cc @dbcli/pgcli-core

darikg:
Everything's cool in Windows as of jonathanslenders/python-prompt-toolkit@c3efe84a084afd8d1020edc7f587118c5c8b7642 -- maybe wait for its next Pypi release?

jonathanslenders:
prompt_toolkit==0.39 released and pull request updated.

amjith:
Thank you all for taking the time to contribute, review and test this PR thoroughly. I appreciate it very much. 

:candy: :camel: :tractor:

Issue #225, opened 2015-04-29
Provide a way to update autocomplete without restarting


chrisanderton:
If the structure of a table changes, the autocomplete options don't seem to refresh - i assume the definitions are cached somewhere?

It would be good to provide a command to invalidate the cache to avoid having to quit and reload to get the latest structure in the autocomplete..

amjith:
@chrisanderton Can you tell us how you're altering the table? We're automatically refreshing the completion when we encounter a `CREATE` or `ALTER` statement. We're also investigating various ways to auto-detect when schemas are modified (see #211). 

But I'm open to the idea of adding a command called `refresh` or `rehash` which will refresh the autocompletion.

chrisanderton:
I'm not altering the table from within pgcli - in this particular case, another tool i'm working with is setting up tables/altering them - and i am using pgcli to interact with them.

I know when the table has changed - as i have made it happen - so i'd be happy with a manual command instead of auto-detection magic! :)

amjith:
@chrisanderton This is now available in 'master'. You can install directly from master using:

```
pip install -U -e git+https://github.com/dbcli/pgcli.git#egg=pgcli
```

You can refresh the completions using `\#` or `\refresh` commands.

PR #253, opened 2015-06-05
Amjith/specials refactor

amjith:
Reviewer: @darikg & @j-bennet 

This is a complete rewrite of the specials package. A few things have changed. 

1. pgspecial is now a directory with all the relevant files to make it into a standalone package. 
2. `\c, \connect and use` are now handled by pgspecial instead of pgexecute.py (which was always felt icky).  We now pass in the PGExecute as `db_obj` to the specials package. The idea is that `db_obj` should have a `connect()` call which will be invoked to change the database. That way whoever decides to use the pgspecial can provide their own db_obj and still achieve the `\c` functionality.
3. There are `@export` decorators sprinkled around the pgspecial code. The `export` decorator is defined in the `__init__.py` file. When a function is decorated by this `@export` it will be exposed by that package and available to use when that package is imported. It is a technique I picked up from David Beazly's tutorial at this PyCon 2015. http://www.dabeaz.com/modulepackage/index.html. Check the slides 53 to 56 on this pdf: http://www.dabeaz.com/modulepackage/ModulePackage.pdf for more details. BTW, that video is worth watching. :)

I'm tagging both of you on this PR since you two have been the most active in that part of the code. I've ran through most of the special commands by hand to make sure I didn't break anything. 

Give it a whirl.


amjith:
> 40 new problems were found (including 1 error and 39 code smells).

Holy crap! What did I do to deserve this? ;)

darikg:
What's with adding `**kwargs` to all the special function signatures?

amjith:
The `pgspecial.main.execute` takes an extra argument (`db_obj`) which is then passed in to the `executor` call. This means all the special command handlers now get cur, pattern, verbose and db_obj as it's params. I could have just added the 'db_obj' to the handle functions but I wanted to future-proof it. So I decided to make it **kwargs, so if we decide to add more in the future, we can do it in the 'executor' call and the handlers don't have to be touched.

j-bennet:
To make the linter happier, you could do some additional tricks on your code, but I'm not sure if you want to.

See for example this:

```
def show_help(**kwargs):  # All the parameters are ignored.
```

Linter will complain because you're not using ```kwargs```. To tell it specifically that you're going to ignore the parameter, you can use _:

```
def show_help(**_):  # All the parameters are ignored.
```

Or assign to _:

```
def show_help(**kwargs):

_ = kwargs # Be explicit about ignoring.
```

Problem is, it does not necessarily add to the clarity of your code. But I would do it anyway, because if you work around these things, then hopefully the remaining linter's complains are more valid and to the point and worth listening to. Right now, you're just going to dismiss everything from linter as "noise".

darikg:
Just spitballing here, but what if pgspecial had an api for registering custom special commands?

Then handling `\connect` could be done wherever, but the the default special functions wouldn't need to ignore any keyword args. 

As a rough sketch:

```
class PGCli ():
  def __init__ (self):
    self.pgspecial = PGSpecial()  # initializes default special commands
    self.pgspecial.register_custom_special('\connect', self.connect_special)
  
  def connect_special (self, cmd_str):
      if cmd_str:
        self.pgexecute.connect()
      else:
        self.pgexecute.connect(database=cmd_str)
```

amjith:
Ooh. That's a interesting idea. 

On one hand I like keeping all the special commands in a single place, but if I were to step back and treat pgspecial as a truly separate package, then having that extensibility makes it huge.

Because my next step was going to be adding a way to manually refresh the completions to address #225. Which would require updating the execute to take another argument `pgcompleter` but if we do the register_custom_special(), then we can simplify execute to take *args and pass that along to the existing handlers. 

I'm not sure if I'm making sense, but writing it out helped me clarify my thought process. 

I like the idea. @darikg Would you like to take a stab at it or should I do it?

darikg:
@amjith I'm not going to have much time in the next couple weeks so by all means go for it. It's nice you a second use for a more complicated special command, that sound help clarify the api a lot more.

amjith:
I had to rebase against master and force push to this PR since github couldn't merge it automatically. 

So if you had this branch checked out you might want to delete that branch and refetch it. 

```
git checkout master
git branch -D amjith/specials-refactor
git pull origin 
git checkout amjith/specials-refactor
```

The updated PR has the `register_special_command()` hook. I haven't implemented the rehash

pgcli/packages/pgspecial/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+def export(defn):
+    """Decorator to explicitly mark functions that are exposed in a lib."""
+    globals()[defn.__name__] = defn
+    __all__.append(defn.__name__)
+    return defn
+
+from .dbcommands import *
+from .iocommands import *
+
+__all__.append(iocommands.TIMING_ENABLED)
```
j-bennet: I think it should be just TIMING_ENABLED here, since you're importing * from iocommands.

pgcli/packages/pgspecial/main.py
```diff
@@ -0,0 +1,108 @@
+import logging
```
j-bennet: Those two imports are unused:

```
import logging

from . import export
```

pgcli/packages/pgspecial/iocommands.py
```diff
@@ -0,0 +1,135 @@
...
+_logger = logging.getLogger(__name__)
+
+TIMING_ENABLED = True
+use_expanded_output = False
+
+@export
+def is_expanded_output():
+    return use_expanded_output
+
+def toggle_expanded_output(cur, pattern, verbose):
```
j-bennet: To make linter happy:

```
def toggle_expanded_output(*_):
```

pgcli/packages/pgspecial/iocommands.py
```diff
@@ -0,0 +1,135 @@
...
+    global use_expanded_output
+    use_expanded_output = not use_expanded_output
+    message = u"Expanded display is "
+    message += u"on." if use_expanded_output else u"off."
+    return [(None, None, None, message)]
+
+def toggle_timing(cur, pattern, verbose):
```
j-bennet: To make linter happy:

```
def toggle_timing(*_):
```

pgcli/packages/pgspecial/main.py
amjith: These two imports are used. The logging module is used to initialize logging which is used to log a message in the `register_special_command()` and `from . import export` is used to import the decorator called `@export` which is also used in the code.

darikg:
Sorry for being a little fixated here (I'll drop it after this comment, I promise), but I still think the ignored `*args` / `**kwargs` are weird. I guess they're there because pgspecial assumes all special commands adhere to the `\special pattern[verbose]` syntax, and has to supply the `pattern` and `verbose` arguments separately? Why not reserve the pattern/verbose parsing for only the default standard special commands, but let non-standard registered custom special commands parse the whole line however they see fit?

amjith:
> Sorry for being a little fixated here (I'll drop it after this comment, I promise). 

It's a very good thing that you're asking me these question. So keep 'em coming.

You have an interesting point about reserving the (pattern, verbose) for standard commands. I'll take another stab at the refactor.

amjith:
We can get rid of patter and verbose but we still need to pass in the cursor. For example take `\n` as an example. It's a custom command but it still behaves like a built in where it needs the cursor to execute a saved query. 

Just thinking out loud to validate my thoughts.

amjith:
I took a stab at the suggestion but pushed it to a separate branch because I'm not sure if it's a better approach. Here's the comparison between the changes in the current PR and the new branch.

https://github.com/dbcli/pgcli/compare/amjith/specials-refactor...amjith/specials-experimental-refactor

I'm not convinced this is better.

Concerns:

* Doesn't solve the **kwargs problem, because some of the custom commands don't even need the query string (for eg: expanded_output() or toggle_timing()). 
* Too much assumed context about what constitutes a STANDARD_COMMAND vs CUSTOM_COMMAND.(For eg: \x, \t, \c are all custom commands even though they are builtin psql commands, but \n is a standard command because it needs the cursor, verbose and pattern args). So a new comer who wants to contribute a special command has to have this context. Even for core devs we'll have to think twice about a special command. Which to me sounds like code smell. 

I'd rather have a convention that says special commands must take 3 args (cur, pattern, verbose), but you're welcome to ignore those params and here's how you can do it (**_).

What do you folks think?

j-bennet:
Didn't think I'd say that (got very excited about "register_special" approach), but I would agree. I still think dynamic command registration was a good thought, but I don't like how this approach comes out. I think the previous refactoring was a better stopping point.

There is a couple more things that come to mind:
1. Because we have quite a bit of cases that don't fit into `cur, pattern, verbose` signature, we could standardize all the signatures using the normal python approach, i.e. everything would take *args, **kwargs. Ignore whatever we don't use. (Except in a lot functions, we'd have copy-paste like this: 
`cur, pattern, verbose = args`. Ouch.)
2. Use reflection to determine what parameters the method accepts. Then everything could have its own signature. (Overkill, anyone?)

I have to think some more about this.

darikg:
At risk of pushing us further down this rabbit hole, what if
  
1. All special command callbacks take two arguments: cursor and the command string
2. Pgspecial exports a `@standard_pattern_command` decorator that wraps standard command callbacks and supplies them with cur, pattern, verbose.

Like you said though, we still end up ignoring arguments in expanded_output and toggle_timing, so it's not a whole lot better

amjith:
@j-bennet I want to force only `**kwargs` no `*args`. That way all the functions will have this signature `def fn(cur, pattern, verbose, **_)` or `def fn(_**)`. The second case is for `timing_enabled` or `expanded_output` where they don't take any args at all. I'm not a fan of reflection, since it can complicated things if a function has a decorator. Because trying reflection on a decorated function will return the args of the decorator. It just complicates things.

amjith:
@darikg Interesting idea to provide the `parse_special_command()` helper as a decorator. Let's do a thought experiment, with three cases, `\timing`, `\c <dbname>`, `\dt+ <tablename>`. 

```
# \timing
def toggle_timing(**_):
    # Doesn't need cur or query.
    TIMING_ENABLED = not TIMING_ENABLED
```

```
# \c <dbname>
# This is a registered command from PGCli class.
def change_db(self, query, **_):
    # Needs the query which will be '\c test_db', but no need for cursor
    dbname = query.split()[-1]
    self.connect(database=dbname)
```


```
@standard_pattern_command
def describe_one_table(cur, pattern, verbose):
    # Do stuff with all the params.
    return results
```

I don't see a marked improvement in the ergonomics. 

Are there use cases that you have in mind which require unparsed raw query? Say implementing `SHOW TABLES` or `SHOW DATABASES` like the way MySQL does? 

Even for those we have to parse them with `query.split()` and see if the first token is `SHOW` and then take the second token as a pattern.

Thoughts?

amjith:
One other situation where the entire query will be useful is the `\e` command. 

```
SELECT * FROM \e
```

That should launch an editor and populate the entire query. I had previously dismissed that as a unique snowflake command that doesn't fit this whole pgspecial paradigm. But maybe it's time to rethink that? 

I'll keep thinking more about this. But if you guys have counter arguments or examples, I'd love to see them.

darikg:
Well I was thinking you'd have `def toggle_timing(cur, query)` which is at least explicit about what's being ignored. Since each callback is expected to parse its own command string, there's no need to 'future-proof' additional kwargs. 

`\copy` would require the entire unparsed command string. (And also I guess a way of injecting custom suggestions, just to throw another spanner in the works.)

amjith:
Summary: All special commands should take `cur, query`. `pgspecial` should provide `parse_special_command()` helper function that will return `(command, pattern, verbose)`.

So `describe_one_table(), toggle_timing(), change_db()` all of them will take `cur, query` as params and call `parse_special_commands()` inside the function body to extract the pattern if needed or ignore it completely. 

Reasons: This is a superset of the `cur, pattern, verbose` args since some special commands need the entire query to operate and not just the pattern after the query.

BTW, `\copy` only needs `pattern` not the entire query. Maybe I should clarify how pattern works. 

Let's take these examples:

command | pattern | verbose 
--------------|-----------|-----------
\dt table_name | table_name | False
\d+ | None | True
\copy (SELECT * FROM tmp JOIN tbl USING (id)) TO '/var/lib/postgres/test1.csv' | (SELECT * FROM tmp JOIN tbl USING (id)) TO '/var/lib/postgres/test1.csv' | False
\x | None | False
\e filename | filename | False
SELECT * FROM blah \e | * FROM blah \e | False

`pattern` is everything that is after the first token.

The only good example I have so far is the editor command `\e` which needs the entire query without chopping off the first token.

I'm still on the fence, but I'm leaning towards `cur, query`, since it's a superset of `cur, pattern, verbose`.

darikg:
You're right, I was conflating `query` and `pattern`. Another thing in favor of `cur, query` is that `\copy` and `\e` could have plus signs used as the addition symbol in their queries. Right now, `parse_special_command` has `command = command.strip().replace('+', '')` which would obviously be incorrect in that case.

amjith:
It's a good thing you both kept pushing the refactor to go even further. Because it forced me to rethink the strategy a little bit and take a slightly different approach. 

See: https://github.com/dbcli/pgcli/compare/amjith/specials-refactor...amjith/specials-decorator?expand=1

This implements a new way of defining special commands using decorators as well as `register_special_command()` hook. 

No more listing of all special commands in a dictionary. Instead functions that are decorated using a `@special_command` is added to the commands dict automatically. Kinda like Flask, where you define the routes of your webapp using decorators. 

Also, no more separate COMMANDS and HIDDEN_COMMANDS, instead the values in the dictionary are a namedtuple called SpecialCommand() which defines everything you need to know about a command. 

I think this is a much better approach. But I'd love to hear your thoughts. If you're satisfied with that approach, I'll close this PR and open a new PR from that branch. 

There are a few minor fixes to do before opening a PR, such as `\e` is not listed in the `\?` anymore. But that is a minor fix.

darikg:
:+1: from me

j-bennet:
Looks great @amjith. Much cleaner and does not look complex, while in fact being complex.

amjith:
Replaced by a new pr #255

Issue #200, opened 2015-04-11
Show meta info in the completions menu


darikg:
Spinning this off from a derailment in #193 

Python prompt toolkit has supporting for showing extra information in the completions menu, as seen in this screenshot:

![](https://raw.githubusercontent.com/jonathanslenders/ptpython/master/docs/images/file-completion.png)

This seems to be as easy as specifying the `display_meta` parameter when constructing each `Completion` object.

We could make use of this in pgcli.

  - Indicate whether each suggestion is a schema, table, view, function, alias, column, or keyword
  - Show the quick help text for all the \special commands

amjith:
I'm fairly certain we need to upgrade prompt_toolkit to the latest version before we can utilize this feature. So this has a hard dependency on #168.

amjith:
Since prompt_toolkit is now upgraded, it'll be nice to add this feature. @darikg Are you interested in taking a stab?

amjith:
This is now merged into master. Thanks @darikg for the PR.

PR #254, opened 2015-06-05
Implement fuzzy completion using regex.

amjith:
Reviewer: @drocco007 

This is an implementation of fuzzy completion (or true subsequence matching) for schemas/tables/columns/views. This is not enabled for keywords.

Typing `djms` will match `django_migrations`. It is implemented as follows:

When user types `djms` we create a regex pattern that looks like this `(d.*j.*m.*s)` and try to match against the collection. So `django_migrations` will be a positive match. 

In order to produce the relevant ordering in the suggestion menu, there are two ranking values used. 

1. Length of a matching group. 
2. Position of the first matching character. 

Here's how it is used:

Let's say there are two tables `django_migrations` and `misago_threads_post_mentions`. 

Case 1:
If a user starts to type `djm` it will be converted to `(d.*j.*m)`, which will match `django_migrations` which is what we want. 

Case 2:
If a user types `migr` it will be `(m.*i.*g.*r.*)` which will match both tables and depending on the location of the table in the list it is possibel that `misago_threads_post_mentions` will be listed first, but the user clearly intended to match `django_migrations` since they typed `migr`. The ranking algo will first check the length of the matching group in this case for `misago_threads_post_mentions` the matching group will be `misago_thr`. For `django_migrations` the matching group will be `migr`. The rank is assigned by the length of the matching group. So `django_migration` wins.

Another scenario: 

Let's take two table `django_migrations` and `migrations`. If the user types `migr`, the matching group will be the same for both strings, so the tie breaker is the appearance of the first matching character. In this case `migrations` table will win and it will be suggested first. 

This was a fun feature to implement. :)

Looking forward to your feedback.

ps: Thanks @zoltu.

### DIFF
--- a/pgcli/pgcompleter.py
+++ b/pgcli/pgcompleter.py
@@ -1,9 +1,9 @@
 from __future__ import print_function, unicode_literals
 import logging
+import re
 from prompt_toolkit.completion import Completer, Completion
 from .packages.sqlcompletion import suggest_type
 from .packages.parseutils import last_word
-from re import compile, escape
 
 try:
     from collections import Counter
@@ -48,7 +48,7 @@ class PGCompleter(Completer):
         self.reserved_words = set()
         for x in self.keywords:
             self.reserved_words.update(x.split())
-        self.name_pattern = compile("^[_a-z][_a-z0-9\$]*$")
+        self.name_pattern = re.compile("^[_a-z][_a-z0-9\$]*$")
 
         self.special_commands = []
         self.databases = []
@@ -175,8 +175,7 @@ class PGCompleter(Completer):
                            'datatypes': {}}
         self.all_completions = set(self.keywords + self.functions)
 
-    @staticmethod
-    def find_matches(text, collection, start_only=False, fuzzy=True):
+    def find_matches(self, text, collection, start_only=False, fuzzy=True):
         """Find completion matches for the given text.
 
         Given the user's input text and a collection of available
@@ -197,11 +196,10 @@ class PGCompleter(Completer):
         completions = []
 
         if fuzzy:
-            #pat = compile(''.join([escape(c) + r'.*' for c in text]))
-            regex = '.*'.join(map(escape, text))
-            pat = compile('(%s)' % regex)
+            regex = '.*?'.join(map(re.escape, text))
+            pat = re.compile('(%s)' % regex)
             for item in sorted(collection):
-                r = pat.search(item)
+                r = pat.search(self.unescape_name(item))
                 if r:
                     completions.append((len(r.group()), r.start(), item))
         else:
--- a/pgcli/pgcompleter.py
+++ b/pgcli/pgcompleter.py
@@ -3,7 +3,7 @@ import logging
 from prompt_toolkit.completion import Completer, Completion
 from .packages.sqlcompletion import suggest_type
 from .packages.parseutils import last_word
-from re import compile
+from re import compile, escape
 
 try:
     from collections import Counter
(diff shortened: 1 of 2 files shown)


drocco007:
Hey Amjith!

This looks great! I think this will be a nice enhancement to this feature, as I've gotten quite used to this sort of matching in PyCharm, Sublime, and other tools that I use.

I have a couple thoughts on the ranking. Here is an example query using the current version in this PR:

pgcli_fuzzy_completion_existing

It feels weird to me that `"user"` ranks below `user_action` etc.; intuitively, I expect the quotes to be ignored, which would put `"user"` first since it's the shortest. This is pretty easy to change: have the `pat.search` call on line `204` use the unquoted string, leaving everything else the same. Although there is an `unescape_name` method in this class, `find_matches` is defined statically so we can't say `self.unescape_name(item)` as is. Here is a proof of concept demonstrating the idea (which might be modified with a refactor of `unescape_name`):

 ```python

        def unquote(s):
            return s[1:-1] if s[0] == s[-1] == '"' else s

        if fuzzy:
            # …
            for item in sorted(collection):
                r = pat.search(unquote(item))  # modification to existing line 204
                # …
```

and the result:

pgcli_fuzzy_unquote


Looking at this output, I again thought it was odd that `api_user` appears before `user_group`. Digging in a little, it turns out the search `user` expands to `u.*s.*e.*r`; notice that `user_group` has *two* `r`s, so the pattern matches `user_gr` instead of the expected `user`. The longer match length forces the ranking of this match down, which again seems counterintuitive. This is easy to change by using the non-greedy version of the repeat pattern on line `201`(`.*?` instead of `.*`) between the search term characters:

```python

         if fuzzy:
            #pat = compile(''.join([escape(c) + r'.*' for c in text]))
            regex = '.*?'.join(map(escape, text))
```

and the result:

pgcli_fuzzy_nongreedy


Another thought: both Sublime Text and PyCharm highlight the matched characters in the completions. Is this something prompt_toolkit supports? If so, I think that would be a nice enhancement to have (probably as a separate issue, though :)

I'll keep this branch checked out as I'm working tomorrow and let you know if any other thoughts arise from using this change in a live environment!

— dan

amjith:
Those are some great observations. I love that you dug in when the ranking was off and found the root cause. :)

Feel free to make the changes and push to the branch. I'll keep an eye. :+1:

amjith:
I've updated the PR with your feedback @drocco007.

I've made the regex non-greedy and unquoted the items. I've made `find_matches()` a non-static function instead of duplicating the unescape_name() method. 

There is still one corner case where the ranking can be off. 

1__3_1_python_-__pgcli___users_amjith_dropbox_code_python_pgcli___tmux_

Even though I've typed 'se' it still recommend information_schema above the entry named "select" which is wrong. 

But the reason for that is because we do this ranking for each category, first we apply the ranking to schemas, then tables, then views etc. But there is no overall ranking for all of them put together. So the schema suggestions always float to the top. 

Frankly I'm ok with that case because once we add the metadata (#200) to the suggestions, it'll make a lot more sense and I'd prefer to have them grouped together.

drocco007:
Yes, I noticed the schema and table groupings when I first took a look and just assumed that behavior was intentional.

I wanted to capture the intent behind the changes I proposed somewhere in the code base. After all, changing a regular expression from `.*` to `.*?` is a pretty subtle and easy to overlook change, especially a few months from now when we haven't been staring at this part of the code for a while. :wink: To that end, I added a couple of tests that capture and document the intent of these changes. Have a look and let me know what you think! :)

amjith:
Very good point. The tests look good. Thanks for adding them in. Go ahead and merge it in if you're satisfied with the changes. :)

PR #255, opened 2015-06-14
Amjith/specials decorator

amjith:
Created a new PR from the decorators branch. 

Description: 
----------------
This implements a new way of defining special commands using decorators as well as register_special_command() hook.

No more listing of all special commands in a dictionary. Instead functions that are decorated using a `@special_command` is added to the commands dict automatically. Kinda like Flask, where you define the routes of your webapp using decorators.

Also, no more separate `COMMANDS` and `HIDDEN_COMMANDS`, instead the values in the dictionary are a namedtuple called `SpecialCommand()` which defines everything you need to know about a command.

I think this is a much better approach. But I'd love to hear your thoughts. If you're satisfied with that approach, I'll close this PR and open a new PR from that branch.

/cc @j-bennet & @darikg


pgcli/main.py
```diff
@@ -79,9 +78,28 @@ def __init__(self, force_passwd_prompt=False, never_passwd_prompt=False,
         # Initialize completer
         smart_completion = c['main'].as_bool('smart_completion')
         completer = PGCompleter(smart_completion)
-        completer.extend_special_commands(CASE_SENSITIVE_COMMANDS.keys())
-        completer.extend_special_commands(NON_CASE_SENSITIVE_COMMANDS.keys())
+        completer.extend_special_commands(COMMANDS.keys())
         self.completer = completer
+        self.register_special_commands()
+
+    def register_special_commands(self):
```
darikg: Can this be replaced with a decorator on `change_db` now? One way to avoid repetition might be to add an argument for hidden aliases in special_command directly, like
`@special_command('\c, '\c[onnect] database_name', 'Change to a new database', hidden_aliases=['use', '\connect'])

pgcli/packages/pgspecial/main.py
```diff
@@ -0,0 +1,77 @@
...
+
+@export
+def parse_special_command(sql):
+    command, _, arg = sql.partition(' ')
+    verbose = '+' in command
+
+    command = command.strip().replace('+', '')
+    return (command, verbose, arg.strip())
+
+def special_command(command, syntax, description, arg_type=PARSED_QUERY,
```
darikg: export special_command instead of register_special_command ?

pgcli/packages/pgspecial/main.py
```diff
@@ -0,0 +1,77 @@
...
+@export
+def register_special_command(handler, command, syntax, description,
+        arg_type=PARSED_QUERY, hidden=False, case_sensitive=True):
+    cmd = command.lower() if not case_sensitive else command
+    COMMANDS[cmd] = SpecialCommand(handler, syntax, description, arg_type,
+                                   hidden, case_sensitive)
+
+@export
+def execute(cur, sql):
+    command, verbose, pattern = parse_special_command(sql)
```
darikg: can you move this line under `elif special_cmd.arg_type == PARSED_QUERY`? It's the only branch that uses it

pgcli/main.py
amjith: I can't replace it with a decorator because when you decorate a method that is in a class, the method is still unbound (it is not attached to an instance). So when the execute() methods looks up the special command in the COMMANDS dict, it'll get back an unbound method which can't be used. We need the actual instance so we can call change_db on the instance of the PGCli class. I'm not sure if I made that point clear. If it's not clear please let me know and I'll try to rephrase it.

I like the idea of passing in the aliases. I'm going to make the aliases always hidden. Because if you have two shortcuts doing the same thing then you don't need them both listed in the help command. Even if we want to list them both we can instead use a second decorator like here: https://github.com/dbcli/pgcli/pull/255/files#diff-f353ad4995ec86e0878d2a80b589c6daR288. Because there is only one 'syntax' and 'description' field and I can't display those for aliases.

pgcli/packages/pgspecial/main.py
amjith: I have to export them both. I'll update the PR. Thanks for catching that.

amjith:
I've added `aliases` to the `register_special_command()` and the `@special_command()` decorator.

I'm using the aliases keyword to the change_db method for 'use', '\c' and '\connect'. But I'm not using it for describe_table_details() here https://github.com/dbcli/pgcli/pull/255/files#diff-f353ad4995ec86e0878d2a80b589c6daR288 because one of them is case_sensitive and the other one is not.

darikg:
Your explanation of how decorators interact with unbound methods made me think of something. Right now pgspecial stores state in its module-level globals, currently `COMMANDS`, `TIMING_ENABLED` and `use_expanded_output`. For standard usage in a cli, when there's only one `PGCli` instance in the python interpreter, this is fine. But when pgcli is used embedded in a larger application, as with the ipython magic, or text editor plugins, you can have multiple `PGCli` instances all sharing the same `pgspecial` module. I can picture some weird stuff happening when different instances try to register the same `\c` command, for instance. This could be worked around by either wrapping pgspecial up into a class, where state is stored in instance variables, or by making the pgspecial module stateless. But I'm not sure how that would interact with the decorator approach. Any thoughts? This doesn't need to be solved immediately, but it'd be nice to have sketch of how that would work.

amjith:
@darikg Yes. It's been bugging me ever since I started the refactor. I think making it a class is the right thing to do. But I think it can wait till we pull it out of pgcli and make it into an independent project. 

I think I'd prefer to merge this PR (if there are no objections) and file an separate extraction issue for it.

darikg:
@j-bennet I'm cool with merging this if you are

j-bennet:
Absolutely. I think it's great.

amjith:
:confetti_ball: :tada: :balloon: 

Thanks for questioning my assumptions and pushing me to go further with the refactor. I think we ended up with a pretty nice interface to the pgspecial module. :)

PR #258, opened 2015-06-17
Do not escape the database names, since they clash with \c commands.

amjith:
Database names should not be quoted even if they are keywords or have non-alphabets. 

Before this change database names with unicode characters were quoted. So when you try to change db, it will suggest  

1__2_0_python_-__pgcli___users_amjith_dropbox_code_python_pgcli___tmux_

But choosing the database name with the quotes will result in error. 

1__2_0_python_-__pgcli___users_amjith_dropbox_code_python_pgcli___tmux__and_comparing_master___amjith_no-escape-db-names_ _dbcli_pgcli

So quoting database names is not a desired behavior. 

/cc @darikg Can you review and merge?

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -87,12 +87,11 @@ class PGCli(object):
                 aliases=('use', '\\connect', 'USE'))
 
     def change_db(self, pattern, **_):
-        db = pattern[1:-1] if pattern[0] == pattern[-1] == '"' else pattern
-
-        if pattern is None:
-            self.pgexecute.connect()
-        else:
+        if pattern:
+            db = pattern[1:-1] if pattern[0] == pattern[-1] == '"' else pattern
             self.pgexecute.connect(database=db)
+        else:
+            self.pgexecute.connect()
 
         yield (None, None, None, 'You are now connected to database "%s" as '
                 'user "%s"' % (self.pgexecute.dbname, self.pgexecute.user))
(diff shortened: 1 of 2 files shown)


darikg:
doesn't that break `drop database` though?

amjith:
You're right. Perhaps I should modify chage_db() to strip the quotes. I'll update the PR.

amjith:
Updated.

darikg:
:stuck_out_tongue_closed_eyes:  `\c` by itself is supposed to reconnect to the current database, but now it gives a `string index out of range` error

amjith:
Excellent catch. I've fixed it now. Thank you!

:)

darikg:
Cool. Wanna squash those commits or merge as is?

amjith:
I've been debating to myself about the practice of squashing. On one hand it makes the master git tree nice and clean. But on the other hand, shouldn't we record our mistakes or refactors in history? That sounded less philosophical in my head. I'd be interested to hear the thoughts of the core team. 

/cc @dbcli/pgcli-core 

I've often wished that the PR author didn't squash the commit then I could see the difference between the previous commit and the commit that fixes the issue. Like the prompt_toolkit upgrade PR. We went from 0.36 to 0.38 all in one PR with one single commit that got overwritten multiple times. It makes for clean history but I would have liked to see the progression.

Thoughts?

amjith:
We don't have to decide this right now. I think I'm ok with not squashing them. You can merge them in @darikg. Unless you'd prefer them to be squashed and you can make a case for it. ;)

darikg:
You got it

PR #259, opened 2015-06-17
Add \# to manually refresh the completions.

### LINKED ISSUE #225: Provide a way to update autocomplete without restarting

amjith:
Closes #225. 

This adds a new special command called `\#` to refresh completions. 

Reviewer: @j-bennet 

/cc @darikg This is a testament to the idea of having register hooks for special commands. It was ridiculously easy to add this new command. Thank you for the idea. :)

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -23,7 +23,7 @@ from pygments.token import Token
 
 from .packages.tabulate import tabulate
 from .packages.expanded import expanded_table
-from .packages.pgspecial.main import (COMMANDS)
+from .packages.pgspecial.main import (COMMANDS, NO_QUERY)
 import pgcli.packages.pgspecial as special
 from .pgcompleter import PGCompleter
 from .pgtoolbar import create_toolbar_tokens_func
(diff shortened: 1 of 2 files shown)


j-bennet:
Useful! Seems like an odd command name choice though - not very intuitive. I know postgres already has "\r" and "\reset" so those are out, but why not "\refresh"?

amjith:
I got the idea from MySQL. In MySQL `\#` is used to refresh auto-completion. But `\refresh` is a good idea. I'll add that as well.

amjith:
I've added in `\refresh` to manually refresh.

j-bennet:
Something weird is going on with completions @amjith. See screenshots.
screenshot 2015-06-17 21 15 41
screenshot 2015-06-17 21 16 51

amjith:
Looks like this bug is not isolated to this branch, I notice this in the master branch as well, but this only happens after I trigger one of the special commands that have been registered using the hook. Such as `\c`. 

I'll look into it. Nice catch, btw.

amjith:
Fixed!

Turns out I wasn't clearing the special commands when I refresh completions. So every time refresh completion was triggered, it added the special commands but never removed the previous set.

j-bennet:
You're fast @amjith! Looks great, going to merge.

PR #260, opened 2015-06-20
add autocompletion for named queries

brettatoms:


pgcli/packages/sqlcompletion.py
```diff
@@ -132,6 +132,9 @@ def suggest_special(text):
             return [{'type': 'schema'},
                     {'type': rel_type, 'schema': []}]
 
+    if cmd == '\\n':
```
amjith: Can we make this 

```
if cmd in ('\\n', '\\ns', '\\nd'):
```

That way it'll suggest the names for all those special commands?

amjith:
While testing this feature I discovered a bug that I introduced with my refactor that prevents the named queries from being written to the ini file.

I'll fix that in a PR separately.

amjith:
Thanks for adding autocompletion for saved queries. Changes look good. 

:bowling:

PR #261, opened 2015-06-20
Assign a filename to the config obj.

amjith:
Reviewer: @macobo 

During my refactor of the config.py I had chosen to instantiate an empty ConfigObj() and then merge in the default and the user config files on it. But this meant the ConfigObj() that I created didn't have a filename associated with it. This broke the behavior of NamedQueries when it tried to save the queries to the ~/.pgclirc file. This PR fixes that issue.

/cc @brettatoms

### DIFF
--- a/pgcli/config.py
+++ b/pgcli/config.py
@@ -6,6 +6,7 @@ def load_config(usr_cfg, def_cfg=None):
     cfg = ConfigObj()
     cfg.merge(ConfigObj(def_cfg, interpolation=False))
     cfg.merge(ConfigObj(expanduser(usr_cfg), interpolation=False))
+    cfg.filename = expanduser(usr_cfg)
 
     return cfg
 


j-bennet:
@amjith Where is the "save config" code?

brettatoms:
I think there's a potential bug where the PGCli instance holds a config  instance and the NamedQueries object also holds an instance so they affectively both have open file handles to the same object.  We should probably have a single shared instance of the config.

amjith:
@j-bennet The save config code is in named queries. https://github.com/dbcli/pgcli/blob/master/pgcli/packages/pgspecial/namedqueries.py#L18

@brettatoms You're right. They do both have copies of the file open. It's not a big deal right now since only one of them will write to it. But this is something to be fixed. 

I think it's time to make the config.py into a class and make it a singleton (I'm not really sure how to do that in Python). 

The config refactor will have to be a separate PR.

macobo:
This lgtm, what are the missing pieces here?

amjith:
No missing pieces at the moment. The config class refactor will have to wait.

PR #262, opened 2015-06-21
Make the horizontal separator in expanded mode a constant length.

amjith:
In expanded mode the record separator looks like this `-[ RECORD 26 ]-------------------------`. 

The number of dashes used to be calculated based on the longest value of a column. Now this is made a constant for two reasons. 

1. If a column has a really long value (say a multi-line paragraph). The separator will look ridiculous which  would make the expanded output useless. 
2. It is more performant to use a constant value instead of trying to iterate over the whole set of values just to determine the max length to use. 

Reviewer: @macobo

### DIFF
--- a/pgcli/packages/expanded.py
+++ b/pgcli/packages/expanded.py
@@ -4,16 +4,8 @@ def pad(field, total, char=u" "):
     return field + (char * (total - len(field)))
 
 def get_separator(num, header_len, data_len):
-    total_len = header_len + data_len + 1
-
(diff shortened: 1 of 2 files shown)


amjith:
@macobo Can you review and merge?

amjith:
I'm guessing @macobo is busy. 

@j-bennet Can you take a look?

j-bennet:
Ok.

j-bennet:
Looks fine. Merging.

PR #263, opened 2015-06-21
Make named query names to the fuzzy matched.

amjith:
Reviewer: @j-bennet 

Previously the named queries were only matched based on starting characters. This change makes the matching to be fuzzy. 

Reasoning: We do start_only matching for keywords, special commands and datatypes. There is no need to restrict the named query matching to start_only. 

/cc @brettatoms


j-bennet:
Works fine, going to merge.

Issue #265, opened 2015-06-21
There's nothing output on "\n" when there's no named queries.


j-bennet:
There should be some kind of message, for example, "No named queries".

amjith:
I'm on it.

j-bennet:
Usage is now printed - closing.

PR #268, opened 2015-06-22
Amjith/named query docs

### LINKED ISSUE #264: More help needed for named queries.
It would be good to add help to special commands. For named queries specifically, because they are a non-standard feature. So you could run `\n --help` and see the usage.

amjith:
Reviewer: @j-bennet 

Closes #264 #265 

When a user invokes `\n` and there are no saved queries, the documentation for that command will be printed. 

When a user makes a mistake like invoking `\ns` or `\nd` without giving it an argument, it will print the usage. 

This should help educate users about this awesome feature. :)

/cc @brettatoms


j-bennet:
Nice. Merging.

PR #266, opened 2015-06-21
This is a beginning of integration tests for pgcli.

j-bennet:
This adds behavioral style integration tests using `behave` and `pexpect`.

Before any scenarios are run, environment.py makes sure to create a test DB with user, pwd, host, dbname configured in behave.ini.

Right now, only two scenarios are added:
1. Connect to pgcli using the test DB, make sure that prompt is displayed.
2. Connect to pgcli using the test DB, make sure that prompt is displayed, send Ctrl + D and exit.

P.S. Not sure why it's also showing @amjith' named queries commit on this PR. :-/

PR #269, opened 2015-06-22
Started working on integration tests.

j-bennet:
Copy of PR #266.

### DIFF
--- a/DEVELOP.rst
+++ b/DEVELOP.rst
@@ -130,9 +130,9 @@ Integration tests use `behave package http://pythonhosted.org/behave/`_.
 Configuration settings for this package are provided via ``behave.ini`` file
 in root directory.
 
-The database user ``pg_test_user`` has to have permissions to create and drop
-test database. Dafault user is ``postgres`` at ``localhost``, without the
-password (authentication mode ****).
+The database user (``pg_test_user = postgres`` in .ini file) has to have
+permissions to create and drop test database. Dafault user is ``postgres``
+at ``localhost``, without the password (authentication mode trust).
 
 First, install the requirements for testing:
 
--- a/DEVELOP.rst
+++ b/DEVELOP.rst
@@ -122,3 +122,33 @@ The rpm package can be installed as follows:
 ::
 
     $ sudo yum install pgcli*.rpm
+
+Running the integration tests
+-----------------------------
+
+Integration tests use `behave package http://pythonhosted.org/behave/`_.
+Configuration settings for this package are provided via ``behave.ini`` file
+in root directory.
+
+The database user ``pg_test_user`` has to have permissions to create and drop
+test database. Dafault user is ``postgres`` at ``localhost``, without the
+password (authentication mode ****).
+
+First, install the requirements for testing:
+
+::
+
+    $ pip install -r requirements-dev.txt
+
+After that, tests can be run with:
+
+::
+
+    $ behave
+
+To see stdout/stderr, use the following command:
+
+::
+
+    $ behave --no-capture
+
--- a/behave.ini
+++ b/behave.ini
@@ -1,5 +1,5 @@
 [behave.userdata]
-pg_test_user = behave_user
+pg_test_user = postgres
 pg_test_pass =
 pg_test_host = localhost
 pg_test_db = pgcli_behave_tests
--- a/behave.ini
+++ b/behave.ini
@@ -1,0 +1,5 @@
+[behave.userdata]
+pg_test_user = behave_user
+pg_test_pass =
+pg_test_host = localhost
+pg_test_db = pgcli_behave_tests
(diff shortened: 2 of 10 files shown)


amjith:
@j-bennet Is there more to be done here or should we merge this in and start adding more behavioral tests?

j-bennet:
I think it' good to merge. I'll add more tests later, but the current setup can also be used by others to build upon.

amjith:
Would you mind adding some instructions to Readme or Develop.rst on how to run this?

I tried running `behave` and I get the following error:

```
Exception OperationalError: FATAL:  role "behave_user" does not exist

  File "/Users/amjith/.virtualenvs/pg-cli/bin/behave", line 11, in <module>
    sys.exit(main())
  File "/Users/amjith/.virtualenvs/pg-cli/lib/python2.7/site-packages/behave/__main__.py", line 109, in main
    failed = runner.run()
  File "/Users/amjith/.virtualenvs/pg-cli/lib/python2.7/site-packages/behave/runner.py", line 672, in run
    return self.run_with_paths()
  File "/Users/amjith/.virtualenvs/pg-cli/lib/python2.7/site-packages/behave/runner.py", line 693, in run_with_paths
    return self.run_model()
  File "/Users/amjith/.virtualenvs/pg-cli/lib/python2.7/site-packages/behave/runner.py", line 471, in run_model
    self.run_hook('before_all', context)
  File "/Users/amjith/.virtualenvs/pg-cli/lib/python2.7/site-packages/behave/runner.py", line 405, in run_hook
    self.hooks[name](context, *args)
  File "features/environment.py", line 48, in before_all
    context.conf['dbname'])
  File "/Users/amjith/Dropbox/code/python/pgcli/features/db_utils.py", line 19, in create_db
    cn = create_cn(hostname, password, username, 'postgres')
  File "/Users/amjith/Dropbox/code/python/pgcli/features/db_utils.py", line 47, in create_cn
    cn = connect(host=hostname, user=username, database=dbname)
  File "/Users/amjith/.virtualenvs/pg-cli/lib/python2.7/site-packages/psycopg2/__init__.py", line 164, in connect
    conn = _connect(dsn, connection_factory=connection_factory, async=async)
psycopg2.OperationalError: FATAL:  role "behave_user" does not exist
```

Could you create the user (or use `postgres` user) as part of the test? 

Also create the database that is needed to run these tests? 

Eg: We do something similar in our regular test suite: https://github.com/dbcli/pgcli/blob/master/tests/utils.py

j-bennet:
Will do.

j-bennet:
Added instructions to DEVELOP.rst (looks like a better place than normal readme?). Changed default test user to postgres.

amjith:
Were you able to get the tests passing? The \? test always times out for me.

j-bennet:
Yes. Timing out means something's wrong though. What's the trace?

j-bennet:
If you look here:

https://github.com/dbcli/pgcli/blob/j-bennet/itegration-tests/features/fixture_data/help.txt

This is what the test expects to be output when user enters `\?` command. So what can happen is, maybe your output is somehow different, and then the process waits and waits for this expected output and never gets it, until it times out.

What do you see in your cli when you enter `\?`?

amjith:
I was using 'fancy_grid' for the table_format in .pgclirc. 

So the output didn't match exactly. Can we use a specific table_format in the integration tests?

j-bennet:
Ah, that explains it. I should set all the output parameters explicitly then, they can be set when calling pgcli from command line, correct?

amjith:
I don't think there is a way to set the grid at launch time via command line arg. It relies on the config file.

j-bennet:
Is there a way to start with a given configuration file?

amjith:
I was thinking the same thing. Not right now but that's a good idea. We should add an option to pass in the config file.

j-bennet:
Ok, so I'm not expecting the exact table output anymore. The tests should pass now. But definitely it would be nice to add the config option. Makes reproducing the exact environment so much easier. Do you mind if I do that (in a different branch)?

DEVELOP.rst
```diff
@@ -122,3 +122,33 @@ The rpm package can be installed as follows:
...
     $ sudo yum install pgcli*.rpm
+
+Running the integration tests
+-----------------------------
+
+Integration tests use `behave package http://pythonhosted.org/behave/`_.
+Configuration settings for this package are provided via ``behave.ini`` file
+in root directory.
+
+The database user ``pg_test_user`` has to have permissions to create and drop
```
amjith: This should be `postgres` user right?

amjith:
After you removed the formatting, it passes the tests as expected. 

Thanks for taking the time to write integrations tests for pgcli. I appreciate it very much. :)

DEVELOP.rst
j-bennet: The setting in .ini file is called "pg_test_user". I updated the readme to make that clear.

amjith:
Thank you! Looks great.

Issue #183, opened 2015-04-02
bizarre windows formatting bug


darikg:
```
select generate_series(1,9), round(random());
+-------------------+---------+
|   generate_series |   round |
|-------------------+---------|
|                 1 |       1 |
|                 2 |       1 |
|                 3 |       1 |
|                 4 |       0 |
|                 5 |       1 |
|                 6 |       1 |
|                 7 |       1 |
|                 8 |       1 |
|                 9 |       0 |
+-------------------+---------+
SELECT 9
> select generate_series(1,10), round(random());
?????????????????+?????????+?????????????????????++++++++??+++???++++++++'?+++%??++++++++??+++???++++++++??+++???++++++++`?+++???++++++++??+++???++++++++??+++%??++++++++??+++%??++++++++<?+++???+++++++?%?+++%??????????????????????
```

Of course everything works fine in linux.

amjith:
Hrm. It seems to work fine in OS X as well. 

Can you try the expanded output mode (`\x`) and see if it has the same problem? 

I'm wondering if the issue is in the formatter or in the output coming from psycopg2. 

Sorry I don't have a windows box to test this out locally. :(

darikg:
bah, turns out it was just windows default pager

PR #272, opened 2015-06-25
Upgrade to prompt_toolkit==0.42.

jonathanslenders:
Upgrade to the latest prompt-toolkit.
Windows support should now have been improved significantly. Especially the rendering (the completion menu doesn't leave traces anymore), and handling of unicode input. French characters of latin-1 like ``éôë``, etc.. should work correctly.

Normally, there are no backwards incompatible changes in this release.

@darikg: Please have a look.

Cheers,
Jonathan

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,7 @@ setup(
         install_requires=[
             'click >= 3.2',
             'Pygments >= 2.0',  # Pygments has to be Capitalcased. WTF?
-            'prompt_toolkit==0.40',
+            'prompt_toolkit==0.42',
             'psycopg2 >= 2.5.4',
             'sqlparse == 0.1.14',
             'configobj >= 5.0.6'


darikg:
Thanks a lot for all your work on the windows side of things. Everything looks good to me, except for some unicode oddities:

```
test> select 'é'
+------------+
| ?column?   |
|------------|
| 82          |
+------------+
SELECT 1
test> select 'xxxé'
+------------+
| ?column?   |
|------------|
| xxx82       |
+------------+
SELECT 1
```
I assume this is a pgcli problem, not a prompt-toolkit one. Hopefully @amjith can point me in the right direction to confirm.

amjith:
I tested it on OS X and I don't see this issue. Are you referring to the alignment issue or the fact that é is returned as 82? 

If you're referring to alignment, we use `wcswidth` library to determine the width of non-ascii characters. You can see how it's used here: https://github.com/dbcli/pgcli/blob/master/pgcli/packages/tabulate.py#L384

When you're on master branch how does the output look when unicode is involved?

darikg:
The 82 thing, yeah. Maybe it's an issue with my pager? Will look into this further when I get a chance. Go ahead and merge whenever

jonathanslenders:
I'm not sure, but maybe it could be a bug in the pager implementation of click.

See: https://github.com/mitsuhiko/click/blob/a0d426d89434fcade5e96545d5203682c4fc5249/click/_termui_impl.py#L255

Click writes to output to a text file and feeds that to the "more" application of Windows. Have a look at whether the text file that click creates, has the cp437 encoding (or whatever ``sys.stdout.encoding`` reports.)

darikg:
Sorry about not getting back to you guys on this earlier, but it's definitely the pager. I was actually using less.exe (which I think came with msysgit) because more.exe was randomly interpreting non-unicode text files as unicode (presumably because of the prevalence of non-text characters from the table outlines; see #183). Switching *back* to more.exe fixes the unicode display issue. Eventually I'll try to figure out whether the problem is with less or click.

amjith:
@darikg Thanks for testing it on Windows. You 're welcome to merge it in if you're satisfied. I've tested this PR on OS X and I'm fairly confident that it's all good. 

BTW, I did get a Windows 7 VM and got python installed but I couldn't get psycopg2 to install. I'll send you an email describing the error message, maybe you can help me. :)

PR #273, opened 2015-06-30
More specials refactoring

darikg:
Building on @amjith's last specials refactor, this wraps up a lot of the specials.main file into a PGSpecial class, which makes it possible for multiple PGCli instances to coexist in the same python interpreter without interfering with each other through module level globals in pgspecial. While doing this I was also able to weaken some of the coupling between pgspecial, pgexecute, and pgcli.

As before, special commands that do not manipulate state are defined by the `@special_command` decorator, which stores metadata in the `COMMANDS` variable. During initialization of a new PGSpecial object, a *copy* of `COMMANDS` is stored in the member variable `commands`. Further special method registration is done through the method PGSpecial.register, which of course modifies only the member variable, not the module one.

One hesitation I had was where the code for `\x` and `\timing` belonged. Because they're ultimately responsible for mutating the state of the cli application I moved them out of the specials package and into PGCli, in parallel to `\c` and `\#`.

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -448,13 +448,13 @@ def cli(database, user, host, port, prompt_passwd, never_prompt, dbname,
 
     pgcli.run_cli()
 
-def format_output(title, cur, headers, status, table_format):
+def format_output(title, cur, headers, status, table_format, expanded=False):
     output = []
     if title:  # Only print the title if it's not None.
         output.append(title)
     if cur:
         headers = [utf8tounicode(x) for x in headers]
-        if special.is_expanded_output():
+        if expanded:
             output.append(expanded_table(cur, headers))
         else:
             output.append(tabulate(cur, headers, tablefmt=table_format,
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -62,15 +62,13 @@ class PGCli(object):
         default_config = os.path.join(package_root, 'pgclirc')
         write_default_config(default_config, '~/.pgclirc')
 
-        self.use_expanded_output = False
-
         self.pgspecial = PGSpecial()
 
         # Load config.
         c = self.config = load_config('~/.pgclirc', default_config)
         self.multi_line = c['main'].as_bool('multi_line')
         self.vi_mode = c['main'].as_bool('vi')
-        self.timing_enabled = c['main'].as_bool('timing')
+        self.pgspecial.timing_enabled = c['main'].as_bool('timing')
         self.table_format = c['main']['table_format']
         self.syntax_style = c['main']['syntax_style']
 
@@ -95,12 +93,6 @@ class PGCli(object):
                               'Refresh auto-completions.', arg_type=NO_QUERY)
         self.pgspecial.register(self.refresh_completions, '\\refresh', '\\refresh',
                               'Refresh auto-completions.', arg_type=NO_QUERY)
-
-        self.pgspecial.register(self.toggle_expanded_output, '\\x', '\\x',
-                              'Toggle expanded output.', arg_type=NO_QUERY)
-
-        self.pgspecial.register(self.toggle_timing, '\\timing', '\\timing',
-                      'Toggle timing of commands.', arg_type=NO_QUERY)
 
     def change_db(self, pattern, **_):
         if pattern:
@@ -320,7 +312,7 @@ class PGCli(object):
 
                         formatted = format_output(title, cur, headers, status,
                                                   self.table_format,
-                                                  self.use_expanded_output)
+                                                  self.pgspecial.expanded_output)
                         output.extend(formatted)
                         end = time()
                         total += end - start
@@ -354,7 +346,7 @@ class PGCli(object):
                     click.secho(str(e), err=True, fg='red')
                 else:
                     click.echo_via_pager('\n'.join(output))
-                    if self.timing_enabled:
+                    if self.pgspecial.timing_enabled:
                         print('Command Time: %0.03fs' % duration)
                         print('Format Time: %0.03fs' % total)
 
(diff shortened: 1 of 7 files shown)


pgcli/packages/pgspecial/main.py
```diff
@@ -18,6 +18,32 @@
...
+class PGSpecial(object):
+    def __init__(self):
+        self.timing_enabled = True
+
+        self.commands = COMMANDS.copy()
+        self.register(self.show_help, '\\?', '\\?', 'Show Help.',
+                      arg_type=NO_QUERY)
+
+    def register(self, *args, **kwargs):
+        register_special_command(*args, command_dict=self.commands, **kwargs)
```
amjith: Why do we still have the global `register_special_command()` function anymore? Why not move that function body into this function and get rid of the global function?

pgcli/packages/pgspecial/main.py
```diff
@@ -18,6 +18,32 @@
...
+
+    def execute(self, cur, sql):
+        return execute(cur, sql, self.commands)
```
amjith: Same as above, why do we need the global equivalent of this function?

pgcli/packages/pgspecial/main.py
```diff
@@ -18,6 +18,32 @@
 class CommandNotFound(Exception):
     pass
 
+
+@export
+class PGSpecial(object):
+    def __init__(self):
+        self.timing_enabled = True
+
+        self.commands = COMMANDS.copy()
```
amjith: Once we take the route of using a class, I'd prefer not to maintain a global COMMANDS var. It'll be better to just encapsulate it as a member var in the class.

pgcli/main.py
```diff
@@ -405,6 +420,19 @@ def get_completions(self, text, cursor_positition):
...
+        message = u"Expanded display is "
+        message += u"on." if self.use_expanded_output else u"off."
+        return [(None, None, None, message)]
+
+    def toggle_timing(self):
+        self.timing_enabled = not self.timing_enabled
+        message = "Timing is "
+        message += "on." if self.timing_enabled else "off."
+        return [(None, None, None, message)]
+
```
amjith: Since we're making PGSpecial as a class, can we maintain these global vars and functions inside that class as member vars and methods? 

Imagine PGSpecial as a standalone library that someone else wants to use in their own REPL (say someone wants to do a GUI equivalent of PGCli). It'll be good to provide as many commands as possible through PGSpecial and only fallback to using the registration hooks when we need special access to connection object (\c) or completion object (\#). 

Thoughts?

pgcli/packages/pgspecial/main.py
darikg: If we want to keep the `@special_command` decorator, it has to run on unbound methods during module loading, right?

amjith:
I like this refactor. This is perfectly inline with what I wanted to do (once I got some time). So thanks for picking it up. 

I'd like us to move even further on the decoupling front and eventually spin off pgspecial as a standalone library.

I've left a few comments regarding the usage of global vars and using the methods in the class to proxy it through to the global functions. 

I think it's a fine way to start the refactor but I think by the time we merge this PR, we should move all those globals into the class. 

Good work @darikg. :)

pgcli/packages/pgspecial/main.py
darikg: There's a weird divide in the special commands, where half of them are just shortcuts for longer queries. It makes sense to me, for pgspecial as a standalone package, to be able to run those commands without having to instantiate an an object. It's not important for pgcli though.

pgcli/packages/pgspecial/main.py
darikg: I think that means we have to discard the `@special_command` decorator though -- it needs somewhere to store special command metadata

pgcli/main.py
darikg: Yeah, but then your standalone library starts making pretty strong assumptions about how it's being used. Say the GUI equivalent doesn't support a trivial distinction between compact and extended display (`\x on` and `\x off`) because it has all sorts of crazy gui bells and whistles. Why would it want PGSpecial to support that when it won't be used?

pgcli/packages/pgspecial/main.py
amjith: You're right. This does create a complication. :(

Here's a thought. 

We have two sets of special commands. 

1. Default Special Commands
2. Custom Special Commands

Default special commands come with the library and they are defined as functions in this package. Custom special commands are registered using the hook. 

Default special commands can be saved under a class variable in PGSpecial class. The Custom special commands are stored in an instance variable in the PGSpecial object.

We have one decorator which is defined as a class method on PGSpecial. This decorator will only be used by the library author to add default special commands to the class var. So we can keep the `describe_table`, `list_views` etc as unbound functions and decorate them using this decorator. This is merely for convenience at this point. 

Custom special commands will have to be added using the reigster_special_command() method which can only be called on an instance of the PGSpecial class. 

So we eschew the `@special_decorator` for external use and only use it for internal purposes and change it to add commands to the class var. 

The absence of decorator for external use is fine in this case, because users can still define their custom commands using the register_special_command() method. 

This idea is fairly new and I haven't thought this through completely, so feel free to disagree.

pgcli/main.py
amjith: What PGSpecial provides is a central place to keep the state about the various special commands. Even now PGSpecial's timing and expanded are just states that are reported to PGCli and we choose to act based on that state.

I think it is reasonable to provide a place to maintain state that is related to the special commands.

pgcli/packages/pgspecial/main.py
amjith: Hrm. So you envision a scenario where someone imports pgspecial but doesn't instantiate PGSpecial object and merely calls pgspecial.execute(cur, '\d')? 

I suppose that is possible with this current approach, but do we need to cater to that scenario? 

I think we should strive to prescribe a single entry point for the 'execute' api, rather than giving two execute methods where one of them is hanging off the pgspecial module and the other hanging off the PGSpecial object.

darikg:
Ok, I think we're on the same page now.

amjith:
All the changes look good. I'll give it a try tomorrow and merge it in. 

Thank you for taking the time to do the clean up.

pgcli/packages/pgspecial/main.py
```diff
@@ -12,12 +12,81 @@
...
+                result.append((value.syntax, value.description))
+        return [(None, result, headers, None)]
+
+    def toggle_expanded_output(self):
+        self.expanded_output = not self.expanded_output
+        message = u"Expanded display is "
+        message += u"on." if self.expanded_output else u"off."
+        return [(None, None, None, message)]
+
+    def toggle_timing(self):
```
amjith: Good call to make `toggle_timing`, `toggle_expanded_output` and `show_help` to be methods on the PGSpecial class. :)

amjith:
Works well. I'm happy to see you take on the second phase of the refactor and complete it. 

:koala:

Issue #288, opened 2015-07-24
Ctrl-C exits pgcli and introduces strange behavior in bash


mikekaminsky:
Connecting to a remote DB and running `\d` lists relations as expected. Hitting Ctrl-C aborts pgcli and causes issues in bash on exit(can't see input, return doesn't jump to new line). Ctrl-C probably should not abort the whole program.

See this discussion about this behavior in the mysql command line client:
https://bugs.mysql.com/bug.php?id=66583

Found this behavior in pgcli Version: 0.17.0

amjith:
The Ctrl-C aborts pgcli only if you're in the pager. If you're just in the regular pgcli prompt it doesn't abort when you Ctrl-C. 

This is due to a bug in one of the libraries I use called 'Click' for piping the output through a pager. 

I've fixed the bug in click and sent a PR which is already merged. https://github.com/mitsuhiko/click/pull/351

But I'm waiting for them to release a new version so I can bump up the dependency of Click. 

Sorry about the aborting behavior. You can recover from the terminal weirdness by typing 'reset' in the terminal.

PR #290, opened 2015-07-24
Bump up click version to address #288

amjith:
Reviewer: @macobo 

Upgrading the minimum required version of click to 4.1. We use the output_via_pager() function in pgcli. When the pager is active (if the output is longer than the screen) if a user presses Ctrl-C it will abort pgcli and cause the terminal to go weird. 

This latest version fixes the issue. 

Relevant PR: https://github.com/mitsuhiko/click/pull/351

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -345,7 +345,10 @@ class PGCli(object):
                     logger.error("traceback: %r", traceback.format_exc())
                     click.secho(str(e), err=True, fg='red')
                 else:
-                    click.echo_via_pager('\n'.join(output))
+                    try:
+                        click.echo_via_pager('\n'.join(output))
+                    except KeyboardInterrupt:
+                        pass
                     if self.pgspecial.timing_enabled:
                         print('Command Time: %0.03fs' % duration)
                         print('Format Time: %0.03fs' % total)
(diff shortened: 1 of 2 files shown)


amjith:
@darikg Can you take a look? 

It's a pretty straight forward minor version bump for click.

darikg:
What's the expected behavior when hitting control-C in a pager now? I still get kicked out of pgcli (after manually quitting the pager)

amjith:
Hitting Ctrl-C should not quit pgcli anymore. The expected behavior is just ignore that keystroke. 

Maybe the behavior is not consistent in Windows. I'll have to try it in Windows.

darikg:
I think it's still KeyboardInterrupt on Windows so I'm not sure where the problem is. Hopefully you can reproduce on Windows -- let me know

amjith:
You're right. It still quits right away when I press Ctrl-C in Windows. I wonder what the hell is going on. This will require poking around in click again. I'll need some more time to find out what is going on. I'll get back to this PR once I settle down the influx of bug reports.

amjith:
I have added a try/except clause around the `echo_via_pager()` to handle the `Ctrl-C` in pager. 

I tried it in windows and it works as expected. Let me know if you find any other issues with it.

Issue #295, opened 2015-07-25
Measure test coverage


thedrow:
Either coveralls or codecov is fine but test coverage should be measured.

amjith:
I've tried a few times to add coverage but I'm always doing something wrong and it never works. 

If you can get the coverage working, I'll be happy to accept a PR.

j-bennet:
I managed to get it working for dockercli with codecov and this config:

https://github.com/j-bennet/dockercli/blob/master/.travis.yml

Will try to add for pgcli.

PR #299, opened 2015-07-27
Added test coverage with codecov.io

j-bennet:
Link to review coverage:

https://codecov.io/github/dbcli

(login with github).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,14 @@ python:
   - "3.4"
 
 install: 
-  - pip install . pytest mock
+  - pip install codecov
+  - pip install -e .
 
-script: py.test
+script:
+  - coverage run --source pgcli -m py.test
+
+after_success:
+  - codecov
 
 notifications:
   webhooks:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -6,9 +6,14 @@ python:
   - "3.4"
 
 install: 
-  - pip install . pytest mock
+  - pip install codecov
```
amjith: Why was pytest and mock removed from the `pip install`?

j-bennet:
Forgot that they're not in setup.py. How does the build not fail? Interesting. I'll add them back.

amjith:
Yeah I was wondering the same thing. It didn't fail because travis ci installs them by default. 

http://docs.travis-ci.com/user/languages/python/#Pre-installed-packages

But I think it is better to install them explicitly.

j-bennet:
Added them back.

amjith:
Looks good. Thanks for taking the time to add it in. 

Closes #295. 

:mailbox:

PR #308, opened 2015-07-30
Integration with prompt_toolkit==0.44

jonathanslenders:
- Re-enable the open-in-editor functionality. Pressing 'v' in Vi mode or Control-X Control-E in emacs mode should open the EDITOR when the input can be edited.

- Visualization of completions in multiple columns. (This is off by default, but can be configured.)

edit: as always, test before merging ;)

### DIFF
--- a/pgcli/key_bindings.py
+++ b/pgcli/key_bindings.py
@@ -13,6 +13,7 @@ def pgcli_bindings(get_vi_mode_enabled, set_vi_mode_enabled):
     assert callable(set_vi_mode_enabled)
 
     key_binding_manager = KeyBindingManager(
+        enable_open_in_editor=True,
         enable_vi_mode=Condition(lambda cli: get_vi_mode_enabled()))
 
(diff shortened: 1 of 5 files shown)


amjith:
I tried this out locally and it works as expected. I noticed that when the multiple columns is on, the meta data information is not displayed. I'm not sure what's a good solution for that. 

It was a nice beginner friendly touch to try nano and pico before jumping into vi or emacs. 

I haven't tried it in windows yet. I'll try it later tonight.

@dbcli/pgcli-core Can you please take it for a spin?

darikg:
:+1:

jonathanslenders:
@amjith What about showing the meta-text in the floating control, right below the completions? Is that a good default for multi-column? (That line is hidden when there is no single completion that has meta information.)

menu

amjith:
I'm on the fence about the meta info at the bottom. I don't have a strong preference.

amjith:
This looks good to me. I'm going to merge it. 

We can worry about the meta info for multi-column menu later.

PR #313, opened 2015-08-02
Add a comment about the keybindings in pgclirc.

amjith:
@j-bennet Can you take a look? 

This is just adding a comment in the config file. But this seems to trip up people all the time.


amjith:
I've update this PR with an additional commit. I've renamed the config option to show completions in columns. This is a new change that was introduced by the upgrade of prompt_toolkit to 0.45. 

I felt that `wider_completion_menu` instead of `display_completions_in_columns` is a more intuitive name.

j-bennet:
Verified, doesn't break anything. Merging.

PR #321, opened 2015-08-05
Fixes for test database being in use problem.

j-bennet:


j-bennet:
Nope. :( Still working on it.

j-bennet:
This build passed but the previous one failed. I think this PR helps but there is something else going on that I didn't catch yet.

amjith:
Thanks hopefully this reduces the false positives.

PR #323, opened 2015-08-05
Fix #283

### LINKED ISSUE #283: Autocompletion breaks when a quoted table is aliased
eg: 

    SELECT * FROM "access" a WHERE a.

darikg:
Pretty easy -- just bump sqlparse from 0.1.14 to 0.1.16, which includes some improvements in alias extraction.


amjith:
Works great. :)

Thank you for finding a more robust way to do this. I would have started parsing through quotes or some other shenanigan. 

And thank you for adding a nice set of tests. 

:baby_chick:

Issue #302, opened 2015-07-29
Add the ability to start pgcli providing path to config file on command line


j-bennet:
Something like this:

```
pgcli -U <user name> --pgclirc /some/custom/path/.pgclirc <database name>
```

rafi:
Please add an environment variable (as well) to override `.pgclirc` path exactly like `psql` does:
```sh
export PSQLRC="$XDG_CONFIG_HOME/psql/config"
export PGCLIRC="$XDG_CONFIG_HOME/pgcli/config"
```
Then, a configuration option to set the history file path:
```ini
hist_file = ~/.cache/pgcli_history
```
Note, `psql` supports this feature as:
```sh
\set HISTFILE ~/.cache/psql_history
```

j-bennet:
This is now in master.

rafi:
Thank you j-bennet!

PR #325, opened 2015-08-06
Pgclirc file location is being read from command line or env.

j-bennet:
This is to address #302. Adds pgclirc location as command line argument or environment variable.


amjith:
@darikg Would you mind reviewing this and merging? I'm swamped.

Issue #310, opened 2015-08-01
Is there a way to modify popup menu colors?


stardiviner:
RT

amjith:
I don't know what you mean by RT. 

But to answer your question, right now there is no easy way to change it. We wanted to make sure the suggestions and the menu are visible in any background, so we decided to hard code it to a known value. 

Is there a specific reason why you want this changed?

stardiviner:
Because the selected candidate background color is not very clear for me.
I usually use black background & white foreground for selected candidate, and white background & black foreground for candidates.

amjith:
Can you take a screenshot with the completion menu showing and paste it here?

These are the lines that control the colors for the completion menu. 

https://github.com/dbcli/pgcli/blob/master/pgcli/pgstyle.py#L20-L21

It is possible to set those colors from a function argument, but I'm reluctant to add more config options unnecessarily. Reason: http://c2.com/cgi/wiki?ConfigurabilityIsTheRootOfAllEvil

But if you have a legitimate reason I'm open to changing that behavior.

stardiviner:
![pgcli-popup-menu-screenshot](https://cloud.githubusercontent.com/assets/378699/9025517/619be5cc-393a-11e5-9a4b-d6663203682b.png)

stardiviner:
@amjith Can you show me how to modify those colors from a function? I learned a little about Python before, but I forgot it now.

amjith:
Here's one way to do it: 

https://gist.github.com/amjith/6825dc9693120db529e5#file-pgstyle-py-L20

You'll see that style_factory takes in two new arguments `completion_selected_color` and `completion_unselected_color` which are used to set the colors down below on Line 20 and 21. 

This needs to be called from the main.py file here: https://github.com/dbcli/pgcli/blob/master/pgcli/main.py#L255

Where you'll pass in two additional arguments for the colors as strings which are formatted like this: 
`bg:#ffffff #000000`

If you were to do this, then you'll have to maintain a separate fork of pgcli with your own customizations. I'm not convinced that this is a necessary config option.

jonathanslenders:
People always like to play with colors.

The hardest part is that colors are actually defined using Pygments colorschemes, which are Python code, and pgcli uses plain text for configuration files.
A configuration section to override the defaults could look like this:
```
[colors]
Token.Menu.Completions.Meta = 'bg:#448888 #ffffff'
Token.Menu.Completions.ProgressButton = 'bg:#003333'
```
[...]

PR #326, opened 2015-08-07
Add custom styles for menus, toolbars etc.

amjith:
@darikg Can you please review? 

This is an attempt at addressing #310. 

It uses `eval()` which is generally frowned upon. But I can't think of a better way to do this. 

Thoughts?


darikg:
Nice! I replaced `eval` with `pygments.string_to_tokentype`.

amjith:
I didn't know about `string_to_tokentype`. Thank you for doing that. :)

PR #328, opened 2015-08-08
Upgrade to prompt_toolkit 0.46

jonathanslenders:


jonathanslenders:
The changelog: https://github.com/jonathanslenders/python-prompt-toolkit/blob/master/CHANGELOG

Relevant for pgcli:
- the meta text is displayed when completions are shown in "wide" multicolumn mode.
- Python 2.6 fix.
- Correct indentation of query.
- Several fixes/improvements for forward/backward search.

Further, I enabled system bindings. This was already longer in prompt-toolkit, but just not enabled. Pressing Control-Z will now suspend the application (use fg to bring back to foreground), and meta+! will show the "system prompt".) And I had to rename/add a few keys in the styling. (I noticed they were changed already a while ago as well.)

Have a try.
Jonathan

pgcli/main.py
```diff
@@ -260,7 +261,8 @@ def prompt_tokens(cli):
         application = Application(style=style_factory(self.syntax_style, self.cli_style),
                                   layout=layout, buffer=buf,
                                   key_bindings_registry=key_binding_manager.registry,
-                                  on_exit=AbortAction.RAISE_EXCEPTION)
+                                  on_exit=AbortAction.RAISE_EXCEPTION,
+                                  ignore_case=True)
```
amjith: What does this `ignore_case` do?

amjith:
@jonathanslenders Thanks for the PR. I just got done testing it. 

I like the meta info at the bottom of the wide_menu. Good choice. :)
The auto-indent for multi-line queries works great. Thank you!
Can you tell me where to learn more about system-keybindings? I see that C-z and M-! works now, but is there more? 
I also like the C-r reverse-search prompt at the bottom. 

Looks like a very nice release for prompt_toolkit.

@dbcli/pgcli-core Can you please try it out as well?

pgcli/main.py
jonathanslenders: This causes search to be case insensitive. I guess that's what we want, because the autocompleter is also case insensitive.

jonathanslenders:
About the key bindings, there are currently 10 sets of bindings that can be enabled/disabled according to certains conditions in the ``KeyBindingManager``: https://github.com/jonathanslenders/python-prompt-toolkit/blob/dd5c76f998e724f4252d4163acb5c15d74c7df91/prompt_toolkit/key_binding/manager.py#L26
```
    load_basic_bindings
    load_basic_system_bindings
    load_emacs_bindings
    load_emacs_open_in_editor_bindings
    load_emacs_search_bindings
    load_emacs_system_bindings
    load_vi_bindings
    load_vi_open_in_editor_bindings
    load_vi_search_bindings
    load_vi_system_bindings
```
Actually, all the key bindings that are by default in GNU readline should also be present here. If not, I consider it a bug. One other thing that was implemented was the 'arg'-style repetition for ``Control-R``. You can type ``Control-R``, type what you are searching for, type again ``Control-R`` to go to the previous instance. Now if you do ``meta+4`` followed by ``Control-R`` it goes four matches back. Same for ``Control-S``. Another useful thing (already present in the previous version.) ``Control-X`` followed by ``Control-E`` in emacs mode opens the editor.
One limitation for now in the Vi navigation mode is that numbers should always come first. ``4dw`` is possible. ``d4w`` is not. Maybe I'll find a way, but that's no priority. (Nobody noticed yet.)

j-bennet:
@amjith : what is Meta + ! on a Mac? I tried "!" with Command, Ctrl, Option, and it does not seem to do anything.

amjith:
Alt-Shift-!

jonathanslenders:
Or Escape followed by '!'. (Not at the same time.)

darikg:
looks good to me

amjith:
Thank you for the PR Jonathan. 

:milky_way:

Issue #289, opened 2015-07-24
support "service="


zakm:
With psql, one can create a .pg_service.conf file with the following format:

```ini
[someservice]
host=the.host
user=the-user
dbname=the-db
```

and then run `psql service=someservice`. Support for this feature would be nice.

amjith:
Reasonable request, I'll see if I can add that feature in. 

Can you tell me if `.pg_service.conf` is always on the home folder of the user or is there a common location like `/etc/postgres`?

Dinoshauer:
Here's the docs for `.pg_service.conf` http://www.postgresql.org/docs/9.4/static/libpq-pgservice.html :)

amjith:
Thank you for the doc. It seems simple enough, I'll give it a shot. Thank you!

Dinoshauer:
Maybe you could even support `.pgpass` files! http://www.postgresql.org/docs/9.3/static/libpq-pgpass.html

Ofcourse, that is if you don't do so already :) I couldn't get the package to work on arch linux `pgcli-git` so I never got around to check

amjith:
There is already support for pgpass. https://github.com/dbcli/pgcli/issues/53 I get it for free since it's a libpq feature. 

pg_service might need some tinkering to parse the file.

Dinoshauer:
It's a regular ini file, python is quite good at that: https://docs.python.org/2/library/configparser.html

amjith:
@Dinoshauer There might be an arch AUR package. Related: https://github.com/dbcli/pgcli/issues/168

/cc @svenstaro

j-bennet:
This is now in master.

PR #331, opened 2015-08-09
Add --service option to pgcli

j-bennet:
This is to address issue #289. Pgcli can be run with:

```
pgcli --service boo
```

and configuration for service name "boo" would be read from ~/.pg_service.conf or a system-wide pg_service.conf file.

It does not exactly copy psql behavior, as it uses "--service=boo" instead of "service=boo". If this is a problem, let me know.

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -73,6 +73,7 @@ class PGCli(object):
         self.pgspecial.timing_enabled = c['main'].as_bool('timing')
         self.table_format = c['main']['table_format']
         self.syntax_style = c['main']['syntax_style']
+        self.cli_style = c['colors']
         self.wider_completion_menu = c['main'].as_bool('wider_completion_menu')
 
         self.logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


amjith:
How hard would it be to make it compliant with Postgres format? I'd prefer not to have deviations unless absolutely necessary, so people can use pgcli as a drop in replacement for psql.

pgcli/config.py
```diff
@@ -16,3 +16,40 @@ def write_default_config(source, destination, overwrite=False):
...
+    :returns: dict, with None for missing keys.
+    """
+    cnf = ConfigObj()
+    for filename in files:
+        try:
+            cnf.merge(ConfigObj(
+                expanduser(filename), interpolation=False))
+        except ConfigObjError as e:
+            if error_logger:
+                assert callable(error_logger)
```
amjith: Why do we assert this? If error_logger is not a callable then this will crash pgcli. If we want to make sure that error_logger is always passed in then we shouldn't be using a default value for it which is `None`.

j-bennet: An assert only runs if the handler is passed in (not None). I think it makes sense to raise an error if the calling code passed in something that can't be called, because it's the calling code's mistake.

j-bennet:
@amjith Looking into implementing "service=something" format to comply with psql.

j-bennet:
Going to rework this.

j-bennet:
Closing because of newer PR #333.

Issue #286, opened 2015-07-24
Use PAGER if the output is to wide


Knorkebrot:
That'd be nice :)

amjith:
Can you elaborate? I'm not sure what you mean by READER.

Knorkebrot:
Oh, I mean PAGER, sorry.

amjith:
I already use the PAGER for output that doesn't fit in the screen. Wider output is a different issue and using a PAGER there doesn't really help.

Knorkebrot:
Why would that be a different issue?

amjith:
Can you jump on gitter? We can hash it out. 

https://gitter.im/dbcli/pgcli

Knorkebrot:
I'm in a hurry, I'll join later.

Knorkebrot:
For illustration: http://tmp.kbct.de/pager.gif

amjith:
Thoughts from gitter chat: 

Easy solution is to add -S to the less_opts and that'll achieve the right results. 

Correct solution is to append to the existing LESS env var instead of replacing the current LESS env var with -SRXF.

Extra credits for providing a way for the user to set the pager options at run time (kinda like how mysql repl does). Using a 'pager' special command. 

Thank you @Knorkebrot for the illustration gif.

amjith:
@Knorkebrot This is now implemented in master. You can try it out by installing pgcli directly from master, as follows: 

```
pip install -U https://github.com/dbcli/pgcli/archive/master.zip
```

You can change the pager at runtime using `\pager [command]`. Eg: `\pager less -S`.

fernandomora:
Could it be possible to set default pager and options in .pgclirc using a line like  `pager = less -iMSx4FXRe` ?
Otherwise it is necessary to set preferred pager options each time pgcli is executed.

amjith:
I can add the pager to the config file. 

In the mean time pgcli does honor the env var PAGER. So you can set your PAGER env var to `less -iMSx4FXRe` and it will be used by pgcli.

Knorkebrot:
Works like a charm :)

PR #332, opened 2015-08-12
Add a \pager special command.

amjith:
@j-bennet Can you review this PR? 

It adds a `\pager` command to special commands. 

This will address #286.

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -383,7 +383,7 @@ class PGCli(object):
     def adjust_less_opts(self):
         less_opts = os.environ.get('LESS', '')
         self.logger.debug('Original value for LESS env var: %r', less_opts)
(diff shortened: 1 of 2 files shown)


j-bennet:
It works fine, but there's a bit of odd behavior. Steps:

1) I open pgcli and select from "wide" table. See screenshot 1.
pgcli-wide-1

2) I press "right" arrow key to scroll right. It scrolls, but the table "jumps" up to the top of my terminal screen. See screenshot 2.
pgcli-wide-2

Expected: for the table to stay at the same vertical position on screen.
Actual: table shifted up.

It's no big deal, if everyone else is fine with it, it's good to merge.

amjith:
I can't do much about that one. It's the behavior of `less` command.

PR #333, opened 2015-08-14
Added "service" option.

j-bennet:
Using dsn-based approach (no config parsing required).
Connect #289.

### DIFF
--- a/pgcli/pgexecute.py
+++ b/pgcli/pgexecute.py
@@ -140,28 +140,42 @@ class PGExecute(object):
               AND n.nspname <> 'information_schema'
         ORDER BY 1, 2;'''
 
-    def __init__(self, database, user, password, host, port):
+    def __init__(self, database, user, password, host, port, dsn):
         self.dbname = database
         self.user = user
         self.password = password
         self.host = host
         self.port = port
+        self.dsn = dsn
         self.connect()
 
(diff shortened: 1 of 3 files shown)


pgcli/pgexecute.py
```diff
@@ -175,6 +189,16 @@ def connect(self, database=None, user=None, password=None, host=None,
...
 
+    def _select_one(self, cur, sql):
+        """
+        Helper method to run a select and retrieve a single field value
+        :param cur: cursor
+        :param sql: string
+        :return: string
+        """
+        cur.execute(sql)
+        return cur.fetchall()[0][0]
```
amjith: I'd prefer to use `cur.fetchone()` instead of `fetchall()`.

amjith:
The PR works as expected. It's nice that we don't have to handle the parsing of pg_service.conf in pgcli. 

Good find that dsn is handled automatically by psycopg2. 

I think once you replace fetchall() with fetchone(), you might consider getting rid of the _select_one() function since it will be doing nothing more than proxying the function call.

j-bennet:
Good point about fetchone! Changed that. I'm not sure I want to get rid of helper function, since it's not a 1-liner but 2-liner, and it's used 4 times.

amjith:
Looks good. Thank you for handling this. :)

:100:

PR #334, opened 2015-08-17
Make \i to run multiple commands and show results.

amjith:
@darikg Can you review? 

This PR enables `\i` to run a file with multiple queries (semi-colon separated) and display the results for each of them. Previously only the results from the last query was displayed.

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -7,6 +7,7 @@ import sys
 import traceback
 import logging
 from time import time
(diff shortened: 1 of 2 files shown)


darikg:
Looks pretty good. One shortcoming is that if you have multiple commands in the sql file, and one of the commands has a syntax error, only the error message is shown, and not the results from the preceding commands.

PR #336, opened 2015-08-18
Multiple queries with syntax errors

### LINKED ISSUE #335: Errors in later semicolon separated commands hide output from earlier commands
E.g.
```
test_db> select 1; select;
syntax error at end of input
LINE 1: select
```

The output from the first command should be printed before the error traceback in the second

amjith:
This fixes #335. 

@darikg  Can you review?

### DIFF
--- a/pgcli/pgexecute.py
+++ b/pgcli/pgexecute.py
@@ -5,7 +5,7 @@ import psycopg2.extras
 import psycopg2.extensions as ext
 import sqlparse
 from .packages import pgspecial as special
-from .encodingutils import unicode2utf8, PY2
+from .encodingutils import unicode2utf8, PY2, utf8tounicode
 import click
 
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


darikg:
Now after an error, it continues to run queries -- I think it should stop after an error. E.g. the 'select 3' query below shouldn't run.

```
test> select 1; select; select 3;
+------------+                   
|   ?column? |                   
|------------|                   
|          1 |                   
+------------+                   
SELECT 1                         
syntax error at end of input     
LINE 1: select                   
              ^                  
                                 
+------------+                   
|   ?column? |                   
|------------|                   
|          3 |                   
+------------+                   
SELECT 1                         
```

amjith:
This is a conundrum. This new behavior complies with how psql behaves. But I remember making a big deal about stopping the execution of statements if one of them has an error in it.

What are your thoughts on this matter?

j-bennet:
I think it should match psql behavior.

darikg:
It looks like psql has an `ON_ERROR_STOP` variable which controls this behavior

amjith:
I'm torn on the decision. 

I want the default behavior for interactive commands to match psql. But running scripts using `\i` should stop after the first error.

I didn't know about `ON_ERROR_STOP`, but in order to use it we need to add support for `SET` which will toggle the feature at runtime. 

Unless one of you is not happy with the current behavior, let's merge this change as is, then add a new special command or a command-line switch to control this behavior.

darikg:
Agreed

PR #337, opened 2015-08-18
Don't end multi-line query within a quote

### LINKED ISSUE #236: pgcli can't handle creating functions
Typically functions in postgres are multi-line. Here's an example:

```
CREATE FUNCTION delete_address() RETURNS TRIGGER AS $_$
BEGIN
    DELETE FROM address WHERE address.id_user = OLD.id;
    RETURN OLD;
END $_$ LANGUAGE 'plpgsql';
```

But it is impossible to enter this in pgcli, because in multi-line mode we treat a semi-colon at the end of a line to be the end of a query. So at the end of `RETURN OLD;` it will end the query and send it to the server which causes a syntax error. :(

I'm open to suggestions for solving this issue. 

My current thoughts are manipulate the [is_multiline()](https://github.com/dbcli/pgcli/blob/master/pgcli/pgbuffer.py#L9) to do clever things. 

Look into what [ptpython](https://github.com/jonathanslenders/ptpython/tree/master/ptpython) does for multi-line mode and see if we can adopt it's behavior.

darikg:
Fixes #236

### DIFF
--- a/pgcli/packages/parseutils.py
+++ b/pgcli/packages/parseutils.py
@@ -213,8 +213,11 @@ def _parsed_is_open_quote(parsed):
         elif (tok.ttype in Token.Name.Builtin
                 and dollar_quote_regex.match(tok.value)):
             # Find the matching closing dollar quote sign
-            for (i, tok2) in enumerate(tokens[i+1:], i+1):
+            for (j, tok2) in enumerate(tokens[i+1:], i+1):
                 if tok2.match(Token.Name.Builtin, tok.value):
+                    # Found the matching closing quote - continue our scan for
+                    # open quotes thereafter
+                    i = j
                     break
             else:
                 # No matching dollar sign quote
--- a/pgcli/packages/parseutils.py
+++ b/pgcli/packages/parseutils.py
@@ -2,7 +2,7 @@ from __future__ import print_function
 import re
 import sqlparse
 from sqlparse.sql import IdentifierList, Identifier, Function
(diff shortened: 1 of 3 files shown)


amjith:
Wow. This is fantastic. 

I'm impressed you've managed to get this working with just regex. I've been testing it for an hour now and it works great. Good job on tackling a thorny issue. 

I was reading through the code changes and I have one minor comment about the reuse of index variable in the for loop, other than that it looks great.

pgcli/packages/parseutils.py
```diff
@@ -188,6 +188,43 @@ def find_prev_keyword(sql):
...
+    i = 0
+    while i < len(tokens):
+        tok = tokens[i]
+        if tok.match(Token.Error, "'"):
+            # An unmatched single quote
+            return True
+        elif (tok.ttype in Token.Name.Builtin
+                and dollar_quote_regex.match(tok.value)):
+            # Find the matching closing dollar quote sign
+            for (i, tok2) in enumerate(tokens[i+1:], i+1):
```
amjith: I'm not a fan of reusing the index variable `i` here. It utilizes one of the obscure scoping rules of Python and it's not readily apparent. 

Can you change this to use `j` and then increment it to `i` as `i += j` in that `elif` clause?

darikg:
That's a fair point -- I updated that line to be more clear

amjith:
Looks good. Thanks for updating it. 

:mega:

PR #340, opened 2015-08-21
Fix autocomplete after an identifier in a where-clause

darikg:
Fixes dbcli/mycli#135 (Sorry don't have mycli installed so fixing it here instead)

Statements like `SELECT * FROM foo WHERE bar ` now suggest only keywords. (Is that right? I feel like I'm forgetting some other syntax option)

### DIFF
--- a/pgcli/packages/sqlcompletion.py
+++ b/pgcli/packages/sqlcompletion.py
@@ -154,6 +154,13 @@ def suggest_based_on_last_token(token, text_before_cursor, full_text, identifier
         # list. This means that token.value may be something like
         # 'where foo > 5 and '. We need to look "inside" token.tokens to handle
         # suggestions in complicated where clauses correctly
+
+        # Check to see if the last token in the where clause is an identifier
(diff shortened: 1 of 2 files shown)


amjith:
I'm still running some tests and investigating this.

Your PR seems to work for this case but I wonder if this fix goes far enough. I'll give it a thorough test before merging. 

Thanks for handling this so promptly.

amjith:
I was testing this case https://github.com/dbcli/mycli/issues/16

```
SELECT * FROM foo where created > now() - INTERVAL 1 DAY
```

I had to add `INTERVAL` to the list of keywords. When I typed 

```
SELECT * FROM foo WHERE created > now() - IN
```

It shows `INT`, `INTEGER` and `INTERVAL`, but when type `INT` the `INTERVAL` suggestion is gone and we only get columns and functions. 

I think the problem is after a `WHERE` clause we need to suggest columns, functions and keywords. 

I made that change in a separate branch and I'm trying to get the tests passing since it'll break a lot of existing tests.

PR #341, opened 2015-08-23
Amjith/where clause completion

amjith:
@darikg Can you review the PR? 

This is to replace #340. 

I've made the WHERE clause to always return columns, functions and keywords. Since keywords is listed at the bottom of the list, we won't seen them most of the time in the menu unless the user is going for a keyword specifically.


darikg:
This makes sense. At first I thought suggesting keywords everywhere was kind of overkill but `INTERVAL` is a perfect example of why we pretty much have to.

amjith:
Yeah I worried about adding keywords but then it ranks so low in the list, it shouldn't be a hassle. 

Is this PR ready for merge or are you still reviewing the changes?

darikg:
Yeah I'll ago ahead and merge it. It might be nice to have some regression tests specifically for the `LIKE` and `INTERVAL` suggestions in case we keep tweaking keyword suggestions down the road but we can do that later.

PR #342, opened 2015-08-24
Autocompletion in named queries.

### LINKED ISSUE #270: There is no autocompletion inside a named query.
When inside the named query, suggestions don't show up.

named-query-no-autocomplete

Boo.

j-bennet:
This fixes #270.

### DIFF
--- a/pgcli/packages/sqlcompletion.py
+++ b/pgcli/packages/sqlcompletion.py
@@ -34,7 +34,7 @@ def suggest_type(full_text, text_before_cursor):
         '\ns zzz SELECT * FROM abc'   -> 'SELECT * FROM abc'
         '  \ns zzz SELECT * FROM abc' -> 'SELECT * FROM abc'
         """
-        pattern = re.compile(r'^\s*\\ns\s+[a-z0-9\-_]+\s+')
+        pattern = re.compile(r'^\s*\\ns\s+[A-z0-9\-_]+\s+')
         if pattern.match(txt):
             txt = pattern.sub('', txt)
         return txt
--- a/pgcli/packages/sqlcompletion.py
+++ b/pgcli/packages/sqlcompletion.py
@@ -29,9 +29,14 @@ def suggest_type(full_text, text_before_cursor):
     identifier = None
 
     def strip_named_query(txt):
-        pattern = re.compile(r'\\(n|nd|ns\s+[a-z]+)\s+')
+        """
+        This will strip "save named query" command in the beginning of the line:
+        '\ns zzz SELECT * FROM abc'   -> 'SELECT * FROM abc'
+        '  \ns zzz SELECT * FROM abc' -> 'SELECT * FROM abc'
+        """
+        pattern = re.compile(r'^\s*\\ns\s+[a-z0-9\-_]+\s+')
         if pattern.match(txt):
-            txt = pattern.sub('', txt, 1)
+            txt = pattern.sub('', txt)
         return txt
 
     full_text = strip_named_query(full_text)
--- a/pgcli/packages/sqlcompletion.py
+++ b/pgcli/packages/sqlcompletion.py
@@ -1,5 +1,6 @@
 from __future__ import print_function
 import sys
+import re
 import sqlparse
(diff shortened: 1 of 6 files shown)


pgcli/packages/sqlcompletion.py
```diff
@@ -27,6 +28,15 @@ def suggest_type(full_text, text_before_cursor):
 
     identifier = None
 
+    def strip_named_query(txt):
+        pattern = re.compile(r'\\(n|nd|ns\s+[a-z]+)\s+')
```
amjith: You might want to anchor it to the beginning of the query so it doesn't match `\n or \nd` in the middle of the query.

j-bennet: That's why I used "match" vs "search".
https://docs.python.org/2/library/re.html#search-vs-match

amjith: Yes but that's only for checking the match, when you call `re.sub()` it still uses the original regex which can replace `\n` anywhere in the query.

j-bennet: The third parameter is there to replace max 1 occurrence. I suppose I can anchor it for clarity but I believe it currently works as it should.

amjith: I don't think the regex is right. 

I think it should be 

```
        pattern = re.compile(r'^\s*\\ns\s+')
```

A few reasons:

1. Always match at the beginning. 
2. Anticipate a leading space in the front of the query. So if someone types `<space> \ns `, it should still work. 
3. \ns should be the only one that allows completion of sql query. \n and \nd should suggest names of the named queries.

j-bennet: Sounds good. I was thinking if user started typing sql, may as well help him autocomplete it, but this is probably a better behavior. I'll change it.

pgcli/packages/sqlcompletion.py
```diff
@@ -27,6 +28,20 @@ def suggest_type(full_text, text_before_cursor):
 
     identifier = None
 
+    def strip_named_query(txt):
+        """
+        This will strip "save named query" command in the beginning of the line:
+        '\ns zzz SELECT * FROM abc'   -> 'SELECT * FROM abc'
+        '  \ns zzz SELECT * FROM abc' -> 'SELECT * FROM abc'
+        """
+        pattern = re.compile(r'^\s*\\ns\s+[a-z0-9\-_]+\s+')
```
j-bennet: @amjith This should fix the problem. Thanks for pointing it out!

amjith: One more minor update. 

```
pattern = re.compile(r'^\s*\\ns\s+[A-z0-9\-_]+\s+')
```

That should include all the alphabets uppercase and lowercase.

amjith:
Looks like the branch is not up to date. I'll make the regex change and rebase against master before merging. 

But other than that it's a good change. 

I believe you (@j-bennet) added autocompletion for the named queries in mycli where it suggests the named query names after \n and \nd. Do you want to take a stab at porting that over?

j-bennet:
I can port it over to mycli.

amjith:
What I meant was porting over this functionality (https://github.com/dbcli/mycli/commit/913bcf71cd342e9de195b4d55cbcd482c6a54d56) from mycli to pgcli. 

That will nicely complement the named query completion.

amjith:
This is good to merge. Thanks for enhancing the completion for named queries. 

:umbrella:

j-bennet:
@amjith I must be sleepy. :) Thought this was already ported! Ok.

darikg:
Just noticed today that [strip_named_query](https://github.com/dbcli/pgcli/blob/d222e085b7e4df351c76ca582952c7cca8b8940e/pgcli/packages/sqlcompletion.py#L31) is a nested method under `suggest_type`. Is that correct? It doesn't seem to depend on anything in the outer scope. Seems like it's just unnecessarily recompiling that regex everytime `suggest_type` gets called.

j-bennet:
Yikes. I moved it into a separate function call for readability but the regex should probably be moved higher up for performance. Good catch @darikg.

PR #343, opened 2015-08-25
Add tests for where clause with keywords.

amjith:
@darikg These are the regression tests for the WHERE clause completion from PR #341.

You're right I should have added the tests with the PR. My bad.


darikg:
:100:

PR #344, opened 2015-08-25
Added read_from_file back into iospecial.

j-bennet:
The 'read_from_file' method was removed but is still being called in external editor command. This adds it back, plus an integration test to make sure the call to external editor works.

### DIFF
--- a/tests/features/iocommands.feature
+++ b/tests/features/iocommands.feature
@@ -1,0 +1,11 @@
+Feature: I/O commands
+
+  @wip
+  Scenario: edit sql in file with external editor
+     Given we have pgcli installed
+      when we run pgcli
+      and we wait for prompt
+      and we start external editor providing a file name
+      and we type sql in the editor
+      and we exit the editor
+      then we see the sql in prompt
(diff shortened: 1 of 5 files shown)


darikg:
These integration tests are so cool

tests/features/iocommands.feature
```diff
@@ -0,0 +1,11 @@
+Feature: I/O commands
+
+  @wip
```
amjith: What does this @wip decorator do? 

I didn't understand their explanation in the docs.

j-bennet: With behave library, you can add any tags to your tests, to run just a subset of them. For example, if I mark one scenario with the tag "@foo", and then run "behave --tags=foo", only this scenario will be run.

The "wip" tag is special in behave, because this is how they chose to denote "work in progress". If you mark your feature "@wip", you can run it with as usual, with "behave --tags=wip". Or, you can use a shortcut: "behave -w". In addition to just running one feature, the "-w" switch will switch off stderr and stdout capture, so you'll see all that in the console (normally it's all hidden), and will add a flag to stop on any error.

I got into the habit of marking the last feature I worked  on as "@wip", because this way I can quickly run it separately. It does not matter for travis, because it always runs "behave" on the whole suite.

amjith:
Cool. Thanks for the explanation. 

Looks good. :monkey:

PR #348, opened 2015-09-11
Let pgcli load config from .config/pgcli/config, if not exists, load from .pgclirc

inkn:
It's better to put config file into .config dir.

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -458,8 +458,8 @@ class PGCli(object):
 @click.option('-v', '--version', is_flag=True, help='Version of pgcli.')
 @click.option('-d', '--dbname', default='', envvar='PGDATABASE',
         help='database name to connect to.')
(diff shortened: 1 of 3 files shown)


amjith:
I agree that .config/pgcli/config would be a better place. 

But the change you've made doesn't quite cover it. This needs to be updated in a few places in pgcli/main.py file. https://github.com/dbcli/pgcli/blob/master/pgcli/main.py#L472

Would you be willing to take a stab?

inkn:
Yes,I'm willing to do, but I'm new in learning python, so, if there are problems, please let me know,thanks.

amjith:
@inkn You're getting close. One small piece you missed was that `~/.config/pgcli` is unrecognizable for os.mkdir(). Because the `~` symbol is not auto-expanded by python. But you can explicitly expand it as follows: 

```
    if not os.path.exists(os.path.expanduser('~/.config/pgcli')):
        os.mkdir(os.path.expanduser('~/.config/pgcli'))
```

PR #349, opened 2015-09-13
Let pgcli load config from .config/pgcli/config

inkn:


inkn:
Oh no, also build failed.

amjith:
The tests are passing locally when I run it in Python 2.7. I'm going to rerun the tests in Travis CI to see if this was a transient failure.

amjith:
@j-bennet Can you help? I can't figure out why the tests are failing in Travis. It is passing locally on my box. 

Also can the @dbcli/pgcli-core team please weigh in on this change? This will move the config file (pgclirc), log file(pgcli.log) and history file (pgcli_history) inside the `~/.config/pgcli` folder instead of using the home folder.

j-bennet:
I'm looking.

pgcli/main.py
```diff
@@ -469,6 +469,9 @@ def cli(database, user, host, port, prompt_passwd, never_prompt, dbname,
         print('Version:', __version__)
         sys.exit(0)
 
+    if not os.path.exists(os.path.expanduser('~/.config/pgcli')):
+        os.mkdir(os.path.expanduser('~/.config/pgcli'))
```
j-bennet: @amjith @inkn : Ok I got it. There should be `os.makedirs` instead of `os.mkdir` here. Travis apparently does not have `~/.config/` dir available.

P.S. @amjith: I'll make sure that when the behavioral tests fail, the output is clearly seen (not just "hey I didn't see what I expected").

amjith: Awesome catch. I had a `.config` folder, so it makes sense that it didn't fail locally. Thank you!

inkn: Nice job. I did not consider the situation that .config dir not exists. I have fixed it now. Thanks very much!

darikg:
Maybe use [appdirs](https://pypi.python.org/pypi/appdirs/1.4.0) instead?

Also what about backwards compatibility? Check for pgclirc in the home folder and print a warning or something?

j-bennet:
Good idea about the warning @darikg. It would be good for the user to know that his "old" config is not working anymore and he better port his settings over.

amjith:
I'd like to take this a step further. I'd like to move the relevant files into `~/.config/pgcli/*`. Then print a warning that such a thing has happened. That way we're not forcing the user to port it manually.

By moving it (instead of copying) we avoid confusion where a user updates his `~/.pgclirc` and wonders why it hasn't taken effect. 

`appdirs` module is tempting, but I wonder if it'll make it harder to provide support for customers in the future. We can't easily tell the user change this value in the config file which is located in `~/.config/pgcli/config` or if you're on Windows check `%USERPROFILE%\AppData\Local\dbcli\pgcli\config`. With `appdirs` we can't reliable say where that file will be located. 

What do you think?

amjith:
@inkn I haven't forgotten about this PR. I'll work on this today to see if I can put together a good migration path before merging this in.

amjith:
Superseded by #356

PR #352, opened 2015-09-15
"Referenced by" formatting fix

TamasNo1:
I realised when using `\d+` on a table, there's a bug in formatting the `Referenced by` section:

Using current implementation it shows:
```
Foreign-key constraints:
    ...
Referenced by:
    TABLE "constraint_name" CONSTRAINT "table_name" FOREIGN KEY ...
Has OIDs: no
```

After fix:
```
Foreign-key constraints:
    ...
Referenced by:
    TABLE "table_name" CONSTRAINT "constraint_name" FOREIGN KEY ...
Has OIDs: no
```

### DIFF
--- a/pgcli/packages/pgspecial/dbcommands.py
+++ b/pgcli/packages/pgspecial/dbcommands.py
@@ -721,7 +721,9 @@ def describe_one_table_details(cur, schema_name, relation_name, oid, verbose):
             if (cur.rowcount > 0):
                 status.append("Referenced by:\n")
             for row in cur:
-                status.append("    TABLE \"%s\" CONSTRAINT \"%s\" %s\n" % row)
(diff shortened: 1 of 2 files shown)


amjith:
Do you think this is a bug in psql as well? Because that sql is taken straight out of psql code base. 

https://github.com/postgres/postgres/blob/master/src/bin/psql/describe.c#L2003

TamasNo1:
They mixed the result indexes instead of sorting the columns in the query which for me seems a bit odd. See it from line 2020: https://github.com/postgres/postgres/blob/master/src/bin/psql/describe.c#L2020

amjith:
You're right. That is odd. Thanks for tracking that down and fixing it in pgcli. 

Cheers!
:fireworks:

PR #356, opened 2015-09-20
Move config file to ~/.config/pgcli

amjith:
This is PR #349 by @inkn to move the config file from ~/.pgclirc to ~/.config/pgcli/config. 

I've added one more commit to this PR which will auto-migrate the config file at launch time and print a message at the top. If there is already a config file under ~/.config/pgcli/config then we won't overwrite it. Pgcli will simply print a message saying the user has to merge their ~/.pgclirc to ~/.config/pgcli/config file. 

I've also added default paths that are compatible with Windows. 

@dbcli/pgcli-core Can you please review this? @darikg I'm interested to know if this works as expected in Windows. My Windows VMs have expired, so I'm still downloading a new Windows VM to test this out. 

/cc @inkn I'm going to close #349 since this PR has all of your original commits.

### DIFF
--- a/pgcli/config.py
+++ b/pgcli/config.py
@@ -1,6 +1,14 @@
 import shutil
+import os
+import platform
 from os.path import expanduser, exists
 from configobj import ConfigObj
+
+def config_location():
+    if platform.system() == 'Windows':
+        return os.getenv('USERPROFILE') + '\AppData\Local\dbcli\pgcli\config'
+    else:
+        return expanduser('~/.config/pgcli/config')
 
 def load_config(usr_cfg, def_cfg=None):
     cfg = ConfigObj()
@@ -16,3 +24,7 @@ def write_default_config(source, destination, overwrite=False):
         return
 
     shutil.copyfile(source, destination)
+
+def upgrade_config(config, def_config):
+    cfg = load_config(config, def_config)
+    cfg.write()
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -475,6 +475,9 @@ def cli(database, user, host, port, prompt_passwd, never_prompt, dbname,
         print('Version:', __version__)
         sys.exit(0)
 
+    if not os.path.exists('~/.config/pgcli'):
+        os.mkdir('~/.config/pgcli')
+
     pgcli = PGCli(prompt_passwd, never_prompt, pgclirc_file=pgclirc)
 
     # Choose which ever one has a valid value.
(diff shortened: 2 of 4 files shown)


darikg:
```
[pgcli2] C:\Users\dg>pgcli postgresql://postgres@localhost/test
  File "C:\Users\dg\Anaconda3\envs\pgcli2\Scripts\pgcli-script.py", line 9, in <module>
    load_entry_point('pgcli==0.19.0', 'console_scripts', 'pgcli')()
  File "C:\Users\dg\Anaconda3\envs\pgcli2\lib\site-packages\click\core.py", line 664, in __call__
    return self.main(*args, **kwargs)
  File "C:\Users\dg\Anaconda3\envs\pgcli2\lib\site-packages\click\core.py", line 644, in main
    rv = self.invoke(ctx)
  File "C:\Users\dg\Anaconda3\envs\pgcli2\lib\site-packages\click\core.py", line 837, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "C:\Users\dg\Anaconda3\envs\pgcli2\lib\site-packages\click\core.py", line 464, in invoke
    return callback(*args, **kwargs)
  File "c:\users\dg\documents\python\pgcli\pgcli\main.py", line 484, in cli
    shutil.move(os.path.expanduser('~/.pgclirc'), config_location())
  File "C:\Users\dg\Anaconda3\envs\pgcli2\lib\shutil.py", line 302, in move
    copy2(src, real_dst)
  File "C:\Users\dg\Anaconda3\envs\pgcli2\lib\shutil.py", line 130, in copy2
    copyfile(src, dst)
  File "C:\Users\dg\Anaconda3\envs\pgcli2\lib\shutil.py", line 83, in copyfile
    with open(dst, 'wb') as fdst:
IOError: [Errno 2] No such file or directory: 'C:\\Users\\dg\\AppData\\Local\\dbcli\\pgcli\\config'
```

amjith:
Of course. I forgot to update the section for creating folders if they don't exist. I'll update them tonight when I get back home. Thanks for checking it.

amjith:
I've update the PR to check and create the config directory if it doesn't exist. Can you try it again?

darikg:
Ok, it worked correctly. It's a little awkward that the history and log files don't auto-migrate as well, but that's probably more trouble than it's worth.

amjith:
@darikg That's a good point. I can migrate those over as well. I think moving the history file at the very least would be useful.

darikg:
It's just a pain because you would want to also edit the config file to update the settings for the log and history filepath, unless the user has non-default paths there, in which case don't migrate the files I guess?

amjith:
Oh yeah. Didn't think about that. 

I won't migrate them automatically, it's too much work. Let the new generations enjoy the log and history inside the `.config` folder. 

So then no more changes necessary. If you're satisfied with the changes feel free to merge.

PR #357, opened 2015-09-22
Don't hide functions from pg_catalog

### LINKED ISSUE #353: Misc pg functions not showing on smart autocompletion
When typing things like `select pg_get_viewdef('viewname', true)` with Smart Completion turned on, pgcli will not try to complete the `pg_get_viewdef` function. It works with naive completion.

Other functions lacking on smart completion: http://www.postgresql.org/docs/7.4/static/functions-misc.html

darikg:
Fixes #353


amjith:
Looks good @darikg. Thanks for fixing this. 

:alarm_clock:

Issue #315, opened 2015-08-03
Does not respect \x auto


carymrobbins:
Is there a way to get the `psql` behavior with `\x auto`?

amjith:
Yes. This is a known issue. I think it is possible to detect the size of the terminal using prompt_toolkit library. 

So it is possible to add this feature. But I just haven't gotten around to doing it. I'll take this into consideration.

Abscond63:
\x  is not good to use, as it should be working as xml on a page handler.


Extended mark-up language. For web publishing.  Some one is trying to get back at 'X' as extended an issue???

What next rotten apples.


Sharon L Doherty 

On Aug 3, 2015, at 09:32, Cary Robbins <notifications@github.com> wrote:

Is there a way to get the psql behavior with \x auto?

stuartquin:
I think this issue is resolved and can be closed with the next release

amjith:
Yes. @carymrobbins If you'd like to give it a shot now you can always install pgcli from master as follows: 

```
pip install https://github.com/dbcli/pgcli/archive/master.zip
```

carymrobbins:
Looks good, thanks!

PR #359, opened 2015-09-22
Issue #315 Add support for \x auto

stuartquin:
Adds support for `\x auto` option.

* I'm not totally sure about returning `rows` from `tabulate` but it was the most straightforward way I could think of to reuse the data from `cur`
* The `get_terminal_width` function has been borrowed from Stack Overflow and is not extensively tested
(works well on GNOME Terminal 3.6.2)

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -25,7 +25,7 @@ from pygments.token import Token
 
 from .packages.tabulate import tabulate
 from .packages.expanded import expanded_table
-from .packages.pgspecial.main import (PGSpecial, NO_QUERY, is_wider_than_terminal)
+from .packages.pgspecial.main import (PGSpecial, NO_QUERY, content_exceeds_width)
 import pgcli.packages.pgspecial as special
 from .pgcompleter import PGCompleter
 from .pgtoolbar import create_toolbar_tokens_func
@@ -348,10 +348,15 @@ class PGCli(object):
                                 click.secho("Aborted!", err=True, fg='red')
                                 break
 
+                        if self.pgspecial.auto_expand:
+                            max_width = self.cli.output.get_size().columns
+                        else:
+                            max_width = None
+
                         formatted = format_output(title, cur, headers, status,
                                                   self.table_format,
                                                   self.pgspecial.expanded_output,
-                                                  self.pgspecial.auto_expand)
+                                                  max_width)
                         output.extend(formatted)
                         end = time()
                         total += end - start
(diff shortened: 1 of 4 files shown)


darikg:
Thanks for tackling this! I've played around with it a little bit and the console width detection seems to be working correctly on Windows. However I'm seeing two issues, demonstrated in the screenshot below:

1) `\x on` and `\x off` toggle the expanded output display regardless of the second parameter. (I.e., if expanded output is already on, `\x on` should basically do nothing
2) Column values are not correctly vertically aligned.

temp

stuartquin:
Hi @darikg Thanks for the feedback.

I've added support for `\x on` and `\x off`.

I was unable to re-produce the weird alignment issue but have added an explicit `strip` to try and prevent it. If the issue persists let me know the steps to recreate and I'll have another look.

darikg:
Great! That seems to have fixed my alignment issue. I have two further suggestions: instead of relying on that `get_terminal_width` implementation, use the python prompt toolkit api directly, maybe something like this:

```
# self is the main pgcli object
if self.pgspecial.auto_expand:
    size = self.cli.output.get_size()  # returns a [rows, columns] namedtuple
    max_width = size.columns
else:
    max_width = None
```

And then pass that max_width parameter to format_output.

Secondly, the implementation of `is_wider_than_terminal` is a little hard to grok at first -- can you add comments to explain the arithmetic?

stuartquin:
I've updated to use the prompt toolkit method, way cleaner!
Also renamed the magic function to `content_exceeds_width` and removed/explained some of the magic numbers

darikg:
:+1: I'm pretty happy with this. I'll let @amjith have a look before I merge it. Thanks for the quick turnaround.

amjith:
Change looks great! :ship: it. 

Thank you for taking the time to implement the feature @stuartquin. It is much appreciated.

el-ethan:
I have been using `\x auto` in pgcli for a while. I had been trying to set that option by default in my `.psqlrc`, without any luck, but just learned that pgcli uses a different config file. In pgcli's config file, is it possible to set `\x auto` by default? Thank you!

stuartquin:
Hi @el-ethan 

We currently have support for `expand = True/False` in the pgcli config file (default `~/.config/pgcli/config`) but not for AUTO.
Support for this would be worth adding, I'll take a look into it and update here.

Thanks

el-ethan:
Thanks @stuartquin! That did the trick! It took me a little while to get it working though. I was on version `1.3.1` of pgcli before, and had to upgrade through brew before this option worked. Works as I had hoped on `1.5.1`!

PR #360, opened 2015-09-23
Add CONCURRENTLY to keyword completion

### LINKED ISSUE #358: CONCURRENTLY keyword not in suggestion list
When creating an index concurrently like so

    CREATE INDEX CONCURRENTLY ...

the word `CONCURRENTLY` gets syntax highlighting as expected, but is not in the drop-down list of suggested keywords.

I'd be happy to pair on this issue if you want :)

johshoff:
See e.g. http://www.postgresql.org/docs/9.4/static/sql-createindex.html

Fixes #358


amjith:
Looks good @johshoff. Thanks for the PR. 

:melon:

PR #361, opened 2015-09-23
Suggest functions as tables

### LINKED ISSUE #189: Suggest set-returning functions after the FROM keyword
http://www.postgresql.org/docs/9.1/static/functions-srf.html

Set-returning functions should be suggested after the FROM keyword.

darikg:
Fixes #189. Suggest set-returning functions in the appropriate places.

This enhances the amount of function meta-data stored: the function argument list, the function return type, and boolean flags `is_aggregate`, `is_window`, and `is_set_returning`. This is a little bit overkill, in that all we need for this PR is `is_set_returning`, but I plan on using most of this in the future. In particular, after (if) this PR is accepted, I'll have follow up that suggests column names from set-returning functions used as tables. Ultimately after that I'd like to implement function tooltips/hints.

### DIFF
--- a/pgcli/pgexecute.py
+++ b/pgcli/pgexecute.py
@@ -4,9 +4,11 @@ import psycopg2
 import psycopg2.extras
 import psycopg2.extensions as ext
 import sqlparse
+from collections import namedtuple
 from .packages import pgspecial as special
 from .encodingutils import unicode2utf8, PY2, utf8tounicode
 import click
+
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


pgcli/pgexecute.py
```diff
@@ -24,6 +26,10 @@
 # See http://initd.org/psycopg/articles/2014/07/20/cancelling-postgresql-statements-python/
 ext.set_wait_callback(psycopg2.extras.wait_select)
 
+FunctionMetadata = namedtuple('FunctionMetadata',
+                              ['schema_name', 'func_name', 'arg_list', 'result',
```
amjith: What is the `result` attribute here? Is it the return_type? If so, should we call it `return_type`?

amjith:
This is a pretty big change. I think I'm able to follow the changes so far. I'm going to sleep on it and play around with it more tomorrow. But it looks good so far. 

Thanks for undertaking this task. I appreciate it very much. :)

pgcli/pgexecute.py
darikg: Yes, it's the literal `returns` clause as specified in the `create function` definition. `return_type` is a better name than `result`

amjith:
I'm satisfied with this PR.

You're welcome to change the attr_name to `return_type` in a future PR. I'm merging this for now. 

Thank you!

:bullettrain_side:

PR #362, opened 2015-09-23
Upgrade to prompt-toolkit-0.52

jonathanslenders:
Upgrade to the latest prompt toolkit. Mouse support enabled by default. (You can make this configurable if you want to.)

There were a few more changes I needed to make, because some defaults have changed to "do nothing". However, the ``prompt_toolkit.shortcuts.get_input`` method still has the same defaults. So, later I'd like to have a look at using that instead of creating the ``Application`` manually. I think that makes the integration more simple.

### DIFF
--- a/pgcli/key_bindings.py
+++ b/pgcli/key_bindings.py
@@ -15,6 +15,8 @@ def pgcli_bindings(get_vi_mode_enabled, set_vi_mode_enabled):
     key_binding_manager = KeyBindingManager(
         enable_open_in_editor=True,
         enable_system_bindings=True,
+        enable_search=True,
+        enable_abort_and_exit_bindings=True,
         enable_vi_mode=Condition(lambda cli: get_vi_mode_enabled()))
 
     @key_binding_manager.registry.add_binding(Keys.F2)
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -15,8 +15,10 @@ import sqlparse
 from prompt_toolkit import CommandLineInterface, Application, AbortAction
 from prompt_toolkit.enums import DEFAULT_BUFFER
 from prompt_toolkit.shortcuts import create_default_layout, create_eventloop
+from prompt_toolkit.buffer import AcceptAction
 from prompt_toolkit.document import Document
 from prompt_toolkit.filters import Always, HasFocus, IsDone
+from prompt_toolkit.layout.lexers import PygmentsLexer
 from prompt_toolkit.layout.processors import (ConditionalProcessor,
                                         HighlightMatchingBracketProcessor)
 from prompt_toolkit.history import FileHistory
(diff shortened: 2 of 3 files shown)


pgcli/key_bindings.py
```diff
@@ -15,6 +15,8 @@ def pgcli_bindings(get_vi_mode_enabled, set_vi_mode_enabled):
     key_binding_manager = KeyBindingManager(
         enable_open_in_editor=True,
         enable_system_bindings=True,
+        enable_search=True,
```
amjith: Does this option enable the C-r incremental reverse search?

pgcli/main.py
```diff
@@ -286,12 +287,15 @@ def prompt_tokens(cli):
         with self._completer_lock:
             buf = PGBuffer(always_multiline=self.multi_line, completer=self.completer,
                     history=FileHistory(os.path.expanduser(history_file)),
-                    complete_while_typing=Always())
+                    complete_while_typing=Always(),
+                    accept_action=AcceptAction.RETURN_DOCUMENT)
```
amjith: What does this option do?

pgcli/main.py
```diff
@@ -286,12 +287,15 @@ def prompt_tokens(cli):
...
 
             application = Application(style=style_factory(self.syntax_style, self.cli_style),
                                       layout=layout, buffer=buf,
                                       key_bindings_registry=key_binding_manager.registry,
                                       on_exit=AbortAction.RAISE_EXCEPTION,
+                                      on_abort=AbortAction.RETRY,
```
amjith: What is the purpose of this RETRY?

amjith:
How do I enable auto-suggestions?

pgcli/key_bindings.py
jonathanslenders: Yes it does. These bindings are now disabled by default, because it could be that the layout doesn't contain a user control that displays the search, and that would be very confusing.

pgcli/main.py
jonathanslenders: When pressing ``Enter``, the "accept action" of a buffer is executed. In the case that the search-buffer has the focus, enter causes it to go the the match in the previous buffer. When the system buffer has the focus, the accept action will execute the system command. For the default buffer you usually want it to return the content. The default here also changed to "do nothing", because it was very confusing in some other cases.

pgcli/main.py
jonathanslenders: This defines what happens if Control-C has been pressed. By default it's ignored.

jonathanslenders:
For auto suggestion, the ``Buffer`` class accept an ``AutoSuggest`` instance.


If I have time later on, I can create another pull request that does everything through the ``prompt_toolkit.shortcuts.get_input``. That has the most sane defaults for cases where ``prompt-toolkit`` is used as a readline replacement.

amjith:
I've tested this PR and the changes look good to me.

@dbcli/pgcli-core Can you please take this for a spin. 

One notable difference, you can use the mouse to move your cursor. This is especially useful if you're writing a long multi-line query and you want to go back to the middle of the 2nd line. You can just click with your mouse and the cursor will be teleported there. 

If you click and drag you can select the query. But this is quite sluggish for me. I'm wondering if there is a way to disable the drag to select feature. 

@jonathanslenders Your thoughts on disabling drag to select?

jonathanslenders:
Right now there is no way yet to disable mouse selection without disabling all mouse handling.
I will work on better mouse support for a next release. Basically, handling mouse move events and probably double click as well (for selecting a word.)

amjith:
Superseded by #365

PR #363, opened 2015-09-24
Using pgspecial as a separate module.

j-bennet:
This branch of pgcli uses pgspecial as an independent module.

It adds the dependency in setup.py.

All of the code, including tests, is now here:

https://github.com/dbcli/pgspecial


j-bennet:
The dockerized setup that installs pgcli from this branch and lets you play with it is available here:

https://github.com/j-bennet/dockerized/tree/master/pgcli-pgspecial

amjith:
@j-bennet This change makes me incredibly happy. 

Especially this: 
using_pgspecial_as_a_separate_module__by_j-bennet_ _pull_request__363_ _dbcli_pgcli

I know that we've just moved the code to a different library, but it makes me quite happy. :)

I think there is a merge conflict because we just got a PR yesterday to implement `\x` with arguments to implement `\x auto`. PR #359. 

I think that feature is worth porting to pgspecial and doing a new release.

j-bennet:
@amjith This is now ported to pgspecial. Do I need to merge master into this branch?

amjith:
Thanks for the ridiculously quick turn around. :)

I had expected there would be a review before the release of 1.1.0 of pgspecial. But I guess we never explicitly discussed that.

I'm interested to discuss the development process for pgspecial. Should we hold it to the same standards as pgcli where every change to master is reviewed or should we have a more relaxed approach? 

If we adopt a relaxes approach to pgspecial, we can iterate faster and then when we open a pr to pgcli to upgrade the version we can scrutinize the changes more thoroughly. Thoughts? 

/cc @darikg @j-bennet

j-bennet:
@amjith I'd rather have the same standards as pgcli. This change was an exception because it was already reviewed within pgcli and only ported to here. There were no additions to the port whatsoever.

amjith:
I'll pull this down and run a manual merge.

amjith:
`\x auto` isn't working as expected. 

2__pgcli___users_amjith_dropbox_code_python_pgcli__python2_7_

amjith:
This branch has been rebased against master and then force pushed. So your local branch will have to be force updated as well. 

```
git pull -f origin j-bennet/extract-pgspecial
```

amjith:
There was a missing import other than that it was good. 

Thank you for finishing this task. :sake:

j-bennet:
Ow. How did this get in there?

https://github.com/dbcli/pgcli/blob/master/pgcli/packages/pgspecial/tmp.sql

amjith:
Some people are terrible. 

Sorry that was me. :disappointed:

PR #365, opened 2015-09-26
Amjith/prompt toolkit upgrade

amjith:
This supersedes #362 and disables the mouse support. 

Reason: 

Enabling mouse support means users can't select the table output to copy/paste. So until that is resolved we're not going to enable mouse support, since copying the output is something I do frequently.

### DIFF
--- a/pgcli/key_bindings.py
+++ b/pgcli/key_bindings.py
@@ -15,6 +15,8 @@ def pgcli_bindings(get_vi_mode_enabled, set_vi_mode_enabled):
     key_binding_manager = KeyBindingManager(
         enable_open_in_editor=True,
         enable_system_bindings=True,
+        enable_search=True,
+        enable_abort_and_exit_bindings=True,
         enable_vi_mode=Condition(lambda cli: get_vi_mode_enabled()))
 
     @key_binding_manager.registry.add_binding(Keys.F2)
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -15,8 +15,10 @@ import sqlparse
 from prompt_toolkit import CommandLineInterface, Application, AbortAction
 from prompt_toolkit.enums import DEFAULT_BUFFER
 from prompt_toolkit.shortcuts import create_default_layout, create_eventloop
+from prompt_toolkit.buffer import AcceptAction
 from prompt_toolkit.document import Document
 from prompt_toolkit.filters import Always, HasFocus, IsDone
+from prompt_toolkit.layout.lexers import PygmentsLexer
 from prompt_toolkit.layout.processors import (ConditionalProcessor,
                                         HighlightMatchingBracketProcessor)
 from prompt_toolkit.history import FileHistory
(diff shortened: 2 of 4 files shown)


amjith:
@j-bennet Can you help me figure out why the behave tests are failing? It's failing locally as well but I'm not sure why.

j-bennet:
I'm not sure if this is something that's changed in prompt toolkit, but it seems that the way the colored prompt is output now is messing things up.

This is the output that's being captured:

https://gist.github.com/j-bennet/9b26709ec4e210e5ae99

As you can see, after "Home: http://pgcli.com\r\n", it all looks like ANSI color codes. This is also why you don't see in exception being printed out: I'm stripping color codes from the output to not clutter the exception with them. But if you run this script, you'll see that it actually prints out the colored prompt string.

Before, the output was different. It was more like:

```
ansi codes, ansi codes, ansi codes [TEXT WE ARE EXPECTING] ansi codes, ansi codes, ansi codes
```

Not sure what to do about it. :( I did ask @jonathanslenders before if it's possible to run the cli in "non-colored" mode for the tests, but it seems like there's no way.

jonathanslenders:
Do you have unit tests that have assertions on how prompt-toolkit renders to the terminal?

I haven't looked at the code, but if that's the case, and it's important to test this part, I'd suggest to use a custom instance of the ``Output`` class and pass that to ``CommandLineInterface``. There you can record all terminal events.

Actually, I'm also not aware of any change in the output rendering. It's designed to send as few bytes to stdout as possible. So, if suddenly we receive much more output, that's likely a bug.

j-bennet:
@jonathanslenders They are end to end, functional tests. They run pgcli as a user would, using pexpect, enter commands and check for strings in output. I'm not using any api or creating any classes. Maybe we'll have to drop this method of testing.

jonathanslenders:
I'm not sure how reliable this is, because somehow this relies on the rendering algorithm that's used. More correct (but much, much more complex to set up) is to use a real in memory terminal emulator (like https://github.com/selectel/pyte ) to parse the rendering characters and check the output cell by cell.

Anyway, when I have some time, I'll see whether there is something that could actually have cause some changes.

j-bennet:
@amjith Let's remove functional tests from travis for now?

amjith:
But it actually found a bug in the config location change. :disappointed: 

We can disable it on Travis for now. But I'd like to adjust the tests and have it running again. I think it is a worthwhile suite.

jonathanslenders:
After looking into it. I see that this is definitely a "bug" in prompt-toolkit caused by a recent change. It does much more cursor movements than required in order to render the output. For every character that changes, it moves the cursor back to the left margin by outputting ``\r`` and then it moves it to the correct ``x`` position. I'll try to push a new fix later today. Thanks for pointing this out!

(FYI: it shouldn't cause any issue, but it's just not optimal.)

jonathanslenders:
Probably this pull request will fix it: https://github.com/jonathanslenders/python-prompt-toolkit/pull/167

amjith:
Nice!! 

Kudos to @j-bennet for adding those tests. 

We're not going to disable the tests. They did their job well. :)

I'm going to wait on merging this PR until the fix is out in Prompt Toolkit. Because merging it now means the integration tests will keep failing. Let's not do that. 

Thank you @jonathanslenders for looking into it and fixing it so quickly (as usual). We'll wait till the next version is out in pypi.

amjith:
@jonathanslenders I finally got around to trying it on Windows and the enter key is broken. 

After I launch pgcli, I can't press enter. So there is no way for me to execute a command.

Did you try it on Windows?

amjith:
Closing this as we've already upgraded to the newer prompt_toolkit rendering this PR obselete.

Issue #281, opened 2015-07-22
support \h


keturn:
or some other help key to show the syntax of a SQL command.

amjith:
This is a valid request. Unfortunately I don't have an easy solution here. 

There is no way to get the documentation out of postgres programmatically. At least I haven't found a way to do it. 

So I'm open to ideas.

stuartquin:
We could get pretty good results using the sgml from the open source psql client.
For example the `SELECT` command: https://github.com/postgres/postgres/blob/aa9eac45ea868e6ddabc4eb076d18be10ce84c6a/doc/src/sgml/ref/select.sgml#L34

We could parse the synopsis and description sections.

Happy to take a look at this if it seems like a good idea.

amjith:
@stuartquin I'd love to see a solution for this. So if you have an idea I'm not going to stop you from exploring it. 

In fact, @j-bennet had a similar idea where, we can check if psql is installed on the system, if it is, then proxy all the `\h` calls to psql and read the output and display it into pgcli. Which is better than nothing, but I think there might be some challenges to get it right on all platforms. 

So please feel free to explore and see what you can come up with.

amjith:
Fix released in 0.20.0. 

Please upgrade:

```
pip install -U pgcli
```

PR #366, opened 2015-09-27
Issue #281 Support for \h and listing of command help

stuartquin:
Show command based help:

```
\h    -- List of commands with help entries
\h SELECT   -- Synopsis and description of select command
\h FUBAR    -- No command found message
```

A few notes:

* The command index lives in a static file in the projecct - `helpcommands.py` and is generated from https://github.com/postgres/postgres/tree/master/doc/src/sgml/ref using https://gist.github.com/stuartquin/5cff044157d2e5491308
@amjith does this solution feel OK? Maybe the above gist should live somewhere in the project?
* The help index is a static file based of the current master of the postgres project, I'm not sure how `psql` deals with difference in help content when connecting to different versions of postgres?

Thanks


amjith:
Very nice!

Yes that gist should live in a folder called 'scripts' in the project root. We can potentially move other scripts into that folder which are not bundled with pgcli.

I like the clever use of `json.tool` to generate the python file from json output. :)

I'm not really sure how psql differentiates between the versions of Postgres server either. 

I want to play around with this branch a little more before merging. Thanks again for the PR.

@j-bennet @darikg Thoughts?

j-bennet:
Nice work, needs a slightly more detailed readme and perhaps be organized into its own module or subfolder.

I looked at postgresql source code. It seems like psql is not trying to be smart and determine server version or anything. SQL help is compiled into a C file from those SGMLs and included into psql distribution. See here:

https://fossies.org/dox/postgresql-9.4.4/sql__help_8c_source.html

```
* generated by
*     /usr/bin/perl create_help.pl ../../../doc/src/sgml/ref sql_help
*
```

amjith:
@j-bennet I'm not sure I understand what you mean by separate module. 

I think this PR would effectively be moved to https://github.com/dbcli/pgspecial since pgspecial is now a separate package.

The scripts/docparser.py file will also move to that repo. Is that what you mean? 

I do agree that an updated README that shows an example of how to run the docparser would be useful. Right now that information is only in the script's usage help.

amjith:
@stuartquin Just to bring you up to speed, we just finished extracting `pgspecial` as it's own package and included it as a dependency. (See #363 for details).

So the contents of this PR will have to be moved to that repo (https://github.com/dbcli/pgspecial). 

I understand it is frustrating to make the changes and then have to redo them all in a different repo. I'm willing to move the changes to the pgspecial repo if you prefer. Just let me know what you prefer.

stuartquin:
@amjith I noticed the new repo and thought this PR would cause issues, I'll have a go at adding it to the pgspecial repo along with the script and README, will let you know if I have any issues.

@j-bennet Interesting that psql uses an almost identical process, I'll have a look at `create_help.pl`, see if we missed anything

j-bennet:
What I meant was something like this (should have explained myself better):

```
/pgcli
    /packages
        /pgspecial
            /help
                commands.py
                generate.py
                README.md
```

So all help-related things would be in the same place. But it's fine as-is also.

stuartquin:
@j-bennet I think it makes sense to have `help/commands.py` but I think we should keep `generate.py` in a scripts folder as it has a dependency (beautifulsoup4) which is not required by the main project.

I've moved this stuff out to pgspecial and opened a new PR -  https://github.com/dbcli/pgspecial/pull/2
so I think the changes here are redundant now.

Is `pgcli/packages/pgspecial/main.py` still used?

Thanks

j-bennet:
Ouch. This whole directory should not be present in pgcli anymore.

https://github.com/dbcli/pgcli/tree/master/pgcli/packages/pgspecial

@amjith It was a halfway merge, eh? :-)

amjith:
Looks like I merged some stuff that left things in half way. I won't be able to clean it up until I get home tonight.

@j-bennet  If you'd like to clean it up, please feel free.

j-bennet:
See PR #367.

darikg:
I get an error when calling`\h` (no argument) with extended formatting `\x on`: `max() arg is an empty sequence`

stuartquin:
Good spot @darikg 
Since the move to pgspecial the majority of this PR no longer makes sense, I'll close it and move over to 
https://github.com/dbcli/pgcli/pull/369 and https://github.com/dbcli/pgspecial/pull/2

PR #368, opened 2015-09-29
Suggest columns from functions

darikg:
This is a follow-up to #361. See also discussion in #189.

Here's the upshot:

temp

### DIFF
--- a/pgcli/packages/function_metadata.py
+++ b/pgcli/packages/function_metadata.py
@@ -75,6 +75,9 @@ class TypedFieldMetadata(object):
         self.default = []
         self.unknown = []
 
+    def __getitem__(self, attr):
+        return getattr(self, attr)
+
 
 def parse_typed_field_list(tokens):
     """Parses a argument/column list, yielding TypedFieldMetadata objects
@@ -106,7 +109,7 @@ def parse_typed_field_list(tokens):
             elif parens == 0 and tok.value == '=':
                 parse_state = 'default'
             else:
-                getattr(field, parse_state).append(tok)
+                field[parse_state].append(tok)
                 if tok.value == '(':
                     parens += 1
                 elif tok.value == ')':
@@ -124,9 +127,9 @@ def parse_typed_field_list(tokens):
                 # note that `ttype in Name` would also match Name.Builtin
                 field.name = tok.value
             else:
-                getattr(field, parse_state).append(tok)
+                field[parse_state].append(tok)
         else:
-            getattr(field, parse_state).append(tok)
+            field[parse_state].append(tok)
 
     # Final argument won't be followed by a comma, so make sure it gets yielded
     if field.type:
--- a/pgcli/packages/function_metadata.py
+++ b/pgcli/packages/function_metadata.py
@@ -1,0 +1,146 @@
+import re
+import sqlparse
+from sqlparse.tokens import Whitespace, Comment, Keyword, Name, Punctuation
+
+
+table_def_regex = re.compile(r'^TABLE\s*\((.+)\)$', re.IGNORECASE)
+
+
+class FunctionMetadata(object):
+
+    def __init__(self, schema_name, func_name, arg_list, return_type, is_aggregate,
+                 is_window, is_set_returning):
+        """Class for describing a postgresql function"""
+
+        self.schema_name = schema_name
+        self.func_name = func_name
+        self.arg_list = arg_list.strip()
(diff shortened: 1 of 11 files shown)


amjith:
This is going to take me a while to review, but I just wanted to say, this is AWESOME! :smile:

pgcli/packages/function_metadata.py
```diff
@@ -0,0 +1,146 @@
...
+            if parens == 0 and tok.value == ',':
+                # End of the current field specification
+                if field.type:
+                    yield field
+                # Initialize metadata holder for the next field
+                field, parse_state = TypedFieldMetadata(), 'type'
+            elif parens == 0 and tok.value == '=':
+                parse_state = 'default'
+            else:
+                getattr(field, parse_state).append(tok)
```
amjith: Is there a reason for using `getattr(field, parse_state).append(tok)` instead of `field[parse_state].append(tok)`. 

The second one reads better (personal opinion, so feel free to disagree).

tests/test_function_metadata.py
```diff
@@ -0,0 +1,57 @@
...
+                IN c double precision = 9.99",
+                OUT d double precision[]            '''
+    tokens = sqlparse.parse(sql)[0].flatten()
+    args = list(parse_typed_field_list(tokens))
+    assert [arg.name for arg in args] == ['a', 'b', 'c', 'd']
+    assert [arg.mode for arg in args] == ['IN', 'IN', 'IN', 'OUT']
+
+
+def test_parse_typed_field_list_no_arg_names():
+    #waiting on sqlparse/169
```
amjith: Looks like this is merged. I presume this is released, since the tests aren't failing.

tests/test_parseutils.py
```diff
@@ -8,97 +8,140 @@ def test_empty_string():
...
+
+@pytest.mark.parametrize('arg_list', ['', 'arg1', 'arg1, arg2, arg3'])
+def test_simple_function_as_table(arg_list):
+    tables = extract_tables('SELECT * FROM foo({0})'.format(arg_list))
+    assert tables == [(None, 'foo', None, True)]
+
+
+@pytest.mark.parametrize('arg_list', ['', 'arg1', 'arg1, arg2, arg3'])
+def test_simple_schema_qualified_function_as_table(arg_list):
+    tables = extract_tables('SELECT * FROM foo.bar({0})'.format(arg_list))
```
amjith: Is the assert statement missing on this test?

amjith:
It works great and the code looks good. 

Can you add a few more keywords to the completion list such as `RETURNS` and `LANGUAGE`? 

I've sprinkled some questions. Once they're addressed this is good to merge. 

I'm traveling for the next 24hours, so response might not be swift. I'll check it out once I land.

pgcli/packages/function_metadata.py
darikg: The second one totally reads better, thanks

darikg:
I think I addressed all your inline comments. I'll add keywords in a separate PR -- I'm getting sick of keyword changes showing up as one giant git diff so I'd like to break them out into a separate file where they can be newline-separated

darikg:
Oh, there's one other thing I forgot to mention. Right now, every time function columns are suggested, they need to be parsed by the `FunctionMetadata.fieldnames()` method. That's a little bit inefficient, since they only really need to be parsed once. My preference is to have `fieldnames()` operate lazily, and memoize its results after the first call. Alternatively, since completion refreshing is done in a separate thread, we could parse all fieldnames in a single pass during a refresh. What do you think?

amjith:
Changes look good. It is ready to be merged. I presume you'll do the `fieldnames()` optimization in a separate PR. 

Regarding the `fieldnames()`, I'm fine with either approach. Doing it all at startup would mean we can keep things the same and it'll fit in the existing structure. Doing it lazily and memoizing the results doesn't mean it'll be complicated either. So I'm fine with either approach. You can pick your favorite. 

I'll let you merge this in, since I'm not sure if you'd like to optimize it in this PR or separately. 

:rocket:

amjith:
@darikg I liked your idea from https://github.com/dbcli/mycli/issues/128#issuecomment-129549454. 

Having the keywords in a separate config file is a very nice idea. So I'd welcome your plan to do a separate PR for splitting out the completion candidates to a file.

darikg:
Ok, sounds good. I'm going to merge this and handle `fieldnames` a little later

PR #371, opened 2015-10-03
Fix suggestions in compound join clauses

darikg:
Previously, this worked correctly, suggesting columns etc. from `a`
`SELECT * FROM abc a JOIN def d ON a.`

But this suggested only keywords:
`SELECT * FROM abc a JOIN def d ON a.id = d.id AND a.`


amjith:
This looks good. Thanks!

:microscope:

PR #373, opened 2015-10-03
Fix need_completion_refresh

darikg:
Was only actually checking the first command in multiple commands


amjith:
Looks good. :rooster:

Issue #370, opened 2015-10-01
I can't actually send commands, enter does nothing.


jmunson:
When running pgcli, it launches without errors and the autocomplete appears to work fine, but when I hit enter it does not actually send the command. If I enable multi-line mode, enter does create a new line, but if I press enter after a ; it does nothing. If I backspace over the ; and hit enter again, it inserts another newline.

I don't see any error output.
I'm running Arch Linux and installed pgcli and installed it with https://aur.archlinux.org/packages/pgcli-git/

$pgcli --version
Version: 0.19.2
$python --version
Python 3.4.3
$pip freeze      
click==5.1
configobj==5.0.6
namcap==3.2.5
packaging==15.3
pcp==1.0
pgcli==0.19.2
pgspecial==1.1.0
prompt-toolkit==0.52
psycopg2==2.6.1
py==1.4.30
pyalpm==0.7
pyelftools==0.23
Pygments==2.0.2
pygobject==3.16.2
pytest==2.8.0
pyxdg==0.25
six==1.9.0
sqlparse==0.1.16
team==1.0
wcwidth==0.1.5

jmunson:
This was fixed by doing pip install prompt_toolkit==0.46 as suggested by amjith on gitter. Thanks!

PR #374, opened 2015-10-04
Move literal definitions to standalone files

darikg:


darikg:
It looks like the tests are failing because the .txt files aren't being copied to the site-packages directory:

```
IOError: [Errno 2] No such file or directory: '/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/pgcli/packages/pgliterals/keywords.txt'
```

amjith:
I've modified setup.py file to include the txt files from pgliterals folder. 

Should we make the flat files into json files? They are quite simple and straight forward right now, but I feel like making them JSON files would give them a bit of a structure. We can make them into an array: 

```
['ACCESS',
'ADD',
'ALL,
...
]
```

We can use the `json` [library](https://docs.python.org/2/library/json.html) that is available in the standard library to read the file.

j-bennet:
What is the reason to make them text and not a python structure (array or dict) in a separate file?

amjith:
I speak for myself and not for @darikg's intentions on this matter. I think the advantage of a text file is that it'll be easy for someone to add new entries without understanding Python. Not that python code is hard to read or understand, but adding things to a json file or a txt file is less intimidating to a non-python user. 

I also envision a day when we can allow users to supply their own custom files which can live anywhere (such as ~/.pgliterals for example). In which case it is better to have them as flat files instead of python source.

@darikg Your thoughts?

darikg:
@j-bennet Mostly just habit to try to separate code and data. Like @amjith said, long-term we'd like users to be able to override literal definitions somewhere, and having arbitrary executable code in config files tends to lead to trouble.

@amjith Thanks! I didn't realize setup.py was responsible for packaging data like that, but it makes sense. json makes sense too -- I guess that would suggest moving all the literals into a single file, like

```
{
    'keywords': [
        'ACCESS',
        'ADD',
        'ALL,
        ...
    ],
    'functions': [
        'AVG',
        'COUNT',
        ...
```

amjith:
@darikg I didn't think about collapsing it into a single file. That's not a bad idea. It'll be easier in the future to have a single file to override instead of multiple files.

j-bennet:
I see. I kind of think that pgcli's audience being developers, they would figure out a python array. Just like moving styles into pgcli config was not a problem. But yes, json in single file sounds like the best option, especially if it eventually goes into user's home directory.

darikg:
Ok, I moved everything to a single json file, but now I'm getting timeout errors from the behavioral tests...

j-bennet:
Do the tests run locally?

darikg:
I don't know - I haven't been able to get behave to work on Windows. But the rest of the tests pass, and the CLI works normally, so I can't even guess what else broke.

darikg:
It looks like pexpect 4.0 just got released today so I'm suspecting that's the culprit

j-bennet:
@darikg I think it's best to remove those tests from travis temporarily. Too much breakage lately. I may need to rethink  / find other ways to do functional testing.

@amjith Thoughts?

j-bennet:
@darikg Functional tests fail for me locally, which is a good sign. I will try to figure out what's going on.

amjith:
Please don't remove the behave tests from Travis. In the past month every failure on behave has caught a bug.

eg:

1. `~/.config` folder had to be created using the `makedirs` calls.
2. prompt_toolkit 0.52 had a broken prompt, which caused this bug #370.

Both of them are real issues. I do NOT want to disable the tests. I know it adds delay to our process but I think it is a small delay compared to the person-cycles we'll have to waste if we released something buggy and had to answer github issues all day.

amjith:
Obligatory caveat, if everyone feels it does more good by disabling, please feel free to ignore my objections.

j-bennet:
@darikg I made a couple of small fixes in tests and it seems to help. Would that be a problem if I check that in to this branch?

j-bennet:
@amjith Those are good points. I'm probably too sensitive to tests breaking because they were my addition in the first place, so each time this happens I feel like I put a spoke in other people's wheel. They do have their use though, so I'd only disable them if I can't fix them.

j-bennet:
This will be problematic though:

https://travis-ci.org/dbcli/pgcli/jobs/83729711#L166

It seems like new `pexpect` requires `ptyprocess` and that's not available in 2.6. I may need to pin the version of `pexpect`.

darikg:
Yeah I'm with @amjith - I'd hate to lose the behavioral tests. Can we pin everything to the last set of working versions?

j-bennet:
@darikg I think so, I'll look at it tonight. But I'd still want to include my current fix.

j-bennet:
This PR includes integration tests fixes: https://github.com/dbcli/pgcli/pull/375

darikg:
@j-bennet sorry, you meant also you had fixes specific to this PR? feel free to push them up

j-bennet:
@darikg No, it's all specific to pexpect, and included in #375.

darikg:
Ok, @j-bennet fixed everything. This PR is reporting as still failing tests even after retriggering a build, but that's just because of https://github.com/travis-ci/travis-ci/issues/1485.

j-bennet:
Pretty. It should certainly make adding new keywords less annoying.

amjith:
The behave tests are passing for me locally. 

I'm going to merge this PR.

j-bennet:
Thanks @amjith.

Issue #287, opened 2015-07-24
Order keyword autocomplete by some form of frequency of use


MattOates:
At the moment it looks like keywords are ordered alphabetically this is usually not what you want. Instead it might be good to order them by their frequency of appearance in actual SQL. For example WH gives WHEN and WHENEVER before WHERE. This could even be learnt as people use pgcli so its custom to their use case.

amjith:
This would be a fantastic feature. :)

But it'll require some plumbing on the completion algorithm to get it right. So it might not be a quick addition, but I'll definitely take a stab.

Thank you!

mknapik:
:+1:

amjith:
This is now available in v1.0. Please upgrade. 

```
pip install -U pgcli
```

PR #377, opened 2015-10-06
Don't sort keyword suggestions alphabetically

darikg:
Keywords now maintain the same order they were listed in the pgliterals definition, which allows us to prioritize `FROM` over `FREEZE` and `WHERE` over `WHEN`

See #287

### DIFF
--- a/pgcli/pgcompleter.py
+++ b/pgcli/pgcompleter.py
@@ -341,10 +341,11 @@ class PGCompleter(Completer):
                 completions.extend(dbs)
 
             elif suggestion['type'] == 'keyword':
-                keywords = self.find_matches(word_before_cursor, self.keywords,
-                                             start_only=True,
-                                             fuzzy=False,
-                                             meta='keyword')
+                # Suggest keywords without sorting so they're suggested in the
+                # same order as listed in pgliterals
+                keywords = self.find_matches(
+                    word_before_cursor, self.keywords, start_only=True,
+                    fuzzy=False, sort_ranked=False, meta='keyword')
                 completions.extend(keywords)
 
             elif suggestion['type'] == 'special':
--- a/pgcli/pgcompleter.py
+++ b/pgcli/pgcompleter.py
@@ -163,7 +163,7 @@ class PGCompleter(Completer):
         self.all_completions = set(self.keywords + self.functions)
 
     def find_matches(self, text, collection, start_only=False, fuzzy=True,
-                     meta=None, meta_collection=None):
+                     sort_ranked=True, meta=None, meta_collection=None):
         """Find completion matches for the given text.
 
         Given the user's input text and a collection of available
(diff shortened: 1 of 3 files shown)


pgcli/pgcompleter.py
```diff
@@ -218,8 +220,11 @@ def _match(item):
 
                 completions.append((sort_key, item, meta))
 
+        if sort_ranked:
```
j-bennet: Perhaps `if sort_ranked and not fuzzy`?

darikg: I don't think so. Note that the following call to `sorted` is sorting the completions based on the output of the `_match` function, which in the case of fuzzy matching is a numerical score. Does that make sense?

j-bennet: Oh I see. Thanks.

amjith:
I have a couple of thoughts: 

1. The number of arguments we pass to `find_matches()` is getting unruly. There are some options that are mutually exclusive. For example you can't have both fuzzy and start_only, neither can we have fuzzy and unsorted. I think we should consolidate all those three args into a single argument. That argument will take a state (kinda like an enum) and based on that we choose whether to do fuzzy or start_only or unsorted. 

2. I feel like this is a stop gap measure before implement https://github.com/dbcli/pgcli/issues/287. I feel like the frequency based suggestions will learn over time and provide a much more accurate suggestion than our our ordering based on our opinions. 

So the question is should we bother doing the stop gap or should we leap ahead and do the frequency based suggestion? or Implement a hybrid where the frequency based suggestion will kick in once we have a history and until then we use the manually ordered list.

I just had a (questionably) brilliant idea. We should not sort the suggestions at all. We either order them via fuzzy or we order them based on their frequency. If there is no frequency (because there is no history to learn from), then the weight for frequency will be zero in which case it'll fallback to the manual ordering in the pgliterals file. 

I guess I had more than a couple of thoughts. :smiley:

darikg:
Moving discussion of the learning algorithm to #379. You're entirely correct that this PR can serve as a fallback strategy, improving the out-of-box experience.

> neither can we have fuzzy and unsorted

That's not true. Nothing *calls* `find_matches` with fuzzy and unsorted, but it's certainly an intelligible combination. I agree with combining `fuzzy` and `start_only` though. What if we just have one boolean `fuzzy`, where `fuzzy=False` means start_only?

amjith:
I'm cool with having just one boolean called fuzzy to determine fuzzy or start_only. 

We get rid of the sorting and rely on the ordering in pgliterals. That will definitely improve the immediate out of box experience.

amjith:
@darikg Are you planning to update this PR or should I implement the changes?

darikg:
I'm playing around with `find_matches` in #379. I'll update that PR soon and you can see what you think

amjith:
@darikg Do we still need this PR or is completely superseded by #379? I'm preparing for the release.

darikg:
Superceded by #379

PR #379, opened 2015-10-07
[experimental] learn keyword preferences

darikg:
Spinning this off from discussion in #377 since they're (mostly) orthogonal.  I was originally a little skeptical of using any learning algorithm because 

1. It might be annoying to have different suggestions on different installations
2. It might preclude improving the base suggestion engine. Right now sqlcomplete suggests only the very vague `keyword` category when it could take greater advantage of syntactical restrictions. Given `SELECT * F`, `FREEZE` is a pretty silly suggestion because it's just not valid sql.

On the other hand, basing suggestions on syntax requires syntactically valid input, and a simple learning algorithm could be much more reliable in the middle of editing temporarily invalid queries. In the long run, the learning approach could move to estimating second- or third-order keyword transitions and be pretty powerful.

So this PR offers a basic experiment of the learning approach. Measure zeroth-order keyword probabilities, and rank keywords thereby. 

Open questions:

1. How should learning be shared between concurrent pgcli sessions? One global state or one state per pgcli instance?  
2. Should we (and if so, how do) we save keyword preferences between sessions?

### DIFF
--- a/pgcli/completion_refresher.py
+++ b/pgcli/completion_refresher.py
@@ -7,6 +7,7 @@ except ImportError:
 from .pgcompleter import PGCompleter
 from .pgexecute import PGExecute
 
+
 class CompletionRefresher(object):
 
     refreshers = OrderedDict()
@@ -15,7 +16,7 @@ class CompletionRefresher(object):
         self._completer_thread = None
         self._restart_refresh = threading.Event()
 
-    def refresh(self, executor, special, callbacks):
+    def refresh(self, executor, special, callbacks, history=None):
         """
         Creates a PGCompleter object and populates it with the relevant
         completion suggestions in a background thread.
@@ -31,9 +32,10 @@ class CompletionRefresher(object):
             self._restart_refresh.set()
             return [(None, None, None, 'Auto-completion refresh restarted.')]
         else:
-            self._completer_thread = threading.Thread(target=self._bg_refresh,
-                                                      args=(executor, special, callbacks),
-                                                      name='completion_refresh')
+            self._completer_thread = threading.Thread(
+                target=self._bg_refresh,
+                args=(executor, special, callbacks, history),
+                name='completion_refresh')
             self._completer_thread.setDaemon(True)
             self._completer_thread.start()
             return [(None, None, None,
@@ -42,7 +44,7 @@ class CompletionRefresher(object):
     def is_refreshing(self):
         return self._completer_thread and self._completer_thread.is_alive()
 
-    def _bg_refresh(self, pgexecute, special, callbacks):
+    def _bg_refresh(self, pgexecute, special, callbacks, history=None):
         completer = PGCompleter(smart_completion=True, pgspecial=special)
 
         # Create a new pgexecute method to popoulate the completions.
@@ -68,8 +70,15 @@ class CompletionRefresher(object):
             # break statement.
             continue
 
+        # Load history into pgcompleter so it can learn user preferences
+        n_recent = 100
+        if history:
+            for recent in history[-n_recent:]:
+                completer.extend_query_history(recent, is_init=True)
+
         for callback in callbacks:
             callback(completer)
+
 
 def refresher(name, refreshers=CompletionRefresher.refreshers):
     """Decorator to populate the dictionary of refreshers with the current
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -9,8 +9,10 @@ import traceback
 import logging
 import threading
 import shutil
+import functools
 from time import time
 from codecs import open
+
 
 import click
 try:
@@ -118,16 +120,19 @@ class PGCli(object):
 
     def register_special_commands(self):
 
-        self.pgspecial.register(self.change_db, '\\c',
-                              '\\c[onnect] database_name',
-                              'Change to a new database.',
-                              aliases=('use', '\\connect', 'USE'))
-        self.pgspecial.register(self.refresh_completions, '\\#', '\\#',
-                              'Refresh auto-completions.', arg_type=NO_QUERY)
-        self.pgspecial.register(self.refresh_completions, '\\refresh', '\\refresh',
-                              'Refresh auto-completions.', arg_type=NO_QUERY)
+        self.pgspecial.register(
+            self.change_db, '\\c', '\\c[onnect] database_name',
+            'Change to a new database.', aliases=('use', '\\connect', 'USE'))
+
+        refresh_callback = lambda: self.refresh_completions(
+            persist_priorities='all')
+
+        self.pgspecial.register(refresh_callback, '\\#', '\\#',
+                                'Refresh auto-completions.', arg_type=NO_QUERY)
+        self.pgspecial.register(refresh_callback, '\\refresh', '\\refresh',
+                                'Refresh auto-completions.', arg_type=NO_QUERY)
         self.pgspecial.register(self.execute_from_file, '\\i', '\\i filename',
-                              'Execute commands from file.')
+                                'Execute commands from file.')
 
     def change_db(self, pattern, **_):
         if pattern:
@@ -270,10 +275,15 @@ class PGCli(object):
     def run_cli(self):
         logger = self.logger
         original_less_opts = self.adjust_less_opts()
+        
+        history_file = self.config['main']['history_file']
+        if history_file == 'default':
+            history_file = config_location() + 'history'
+        history = FileHistory(os.path.expanduser(history_file))
+        self.refresh_completions(history=history, 
+                                 persist_priorities='none')
 
-        self.refresh_completions()
-
-        self.cli = self._build_cli()
+        self.cli = self._build_cli(history)
 
         print('Version:', __version__)
         print('Chat: https://gitter.im/dbcli/pgcli')
@@ -335,9 +345,11 @@ class PGCli(object):
                     # Check if we need to update completions, in order of most
                     # to least drastic changes
                     if query.db_changed:
-                        self.refresh_completions(reset=True)
+                        with self._completer_lock:
+                            self.completer.reset_completions()
+                        self.refresh_completions(persist_priorities='keywords')
                     elif query.meta_changed:
-                        self.refresh_completions(reset=False)
+                        self.refresh_completions(persist_priorities='all')
                     elif query.path_changed:
                         logger.debug('Refreshing search path')
                         with self._completer_lock:
@@ -345,6 +357,10 @@ class PGCli(object):
                                 self.pgexecute.search_path())
                         logger.debug('Search path: %r',
                                      self.completer.search_path)
+                        
+                # Allow PGCompleter to learn user's preferred keywords, etc.
+                with self._completer_lock:
+                    self.completer.extend_query_history(document.text)
 
                 self.query_history.append(query)
 
@@ -354,7 +370,7 @@ class PGCli(object):
             logger.debug('Restoring env var LESS to %r.', original_less_opts)
             os.environ['LESS'] = original_less_opts
 
-    def _build_cli(self):
+    def _build_cli(self, history):
 
         def set_vi_mode(value):
             self.vi_mode = value
@@ -383,14 +399,11 @@ class PGCli(object):
                    filter=HasFocus(DEFAULT_BUFFER) & ~IsDone()),
             ])
 
-        history_file = self.config['main']['history_file']
-        if history_file == 'default':
-            history_file = config_location() + 'history'
         with self._completer_lock:
             buf = PGBuffer(
                 always_multiline=self.multi_line,
                 completer=self.completer,
-                history=FileHistory(os.path.expanduser(history_file)),
+                history=history,
                 complete_while_typing=Always())
 
             application = Application(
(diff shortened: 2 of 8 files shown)


amjith:
First of all this is great. I'm impressed at such a quick turn around. 

Here are some possible solutions to the open questions:

> How should learning be shared between concurrent pgcli sessions? One global state or one state per pgcli instance?

Let's keep the learning separate between concurrent pgcli sessions, which is what your implementation does.

Reason: The same user might start two sessions, one as a root user performing maintenance tasks and the other session as a regular user to do exploration.

One suggestion: Initially the KeywordCounter start empty and then learns during the session. But we have a corpus of query strings in our history file. Why not take the latest 100 entries and pre-populate the KeywordCounter? 

> Should we (and if so, how do) we save keyword preferences between sessions?

Yes! Our history files are the way we save the keyword preferences. At the start of a pgcli session we'll read the last 100 or 1000 entries in the history file and use it to train our model (KeywordCounter). 

I have a few more thoughts, I'll leave them in a bit once I've had a chance to formulate them in a coherent manner.

pgcli/pgcompleter.py
```diff
@@ -30,6 +31,7 @@ def __init__(self, smart_completion=True, pgspecial=None):
         super(PGCompleter, self).__init__()
         self.smart_completion = smart_completion
         self.pgspecial = pgspecial
+        self.keyword_counter = KeywordCounter(self.keywords)
```
amjith: Instead of calling it `self.keyword_counter` let's call it `self.recommendation_engine` or something more generic. 

The idea is that in the future we might have different learning algorithms, one based on counting, one based on weighted frequency, or neural networks or something equally crazy etc. :)

We should also make `KeywordCounter()` take in a initial corpus of input strings train the model with the query history.

amjith:
Right now we're only using this learning algorithm for Keywords. Why not use it for tables/columns/views etc? 

The fuzzy matcher sorts the list based on a tuple with two values, length of the matching group and the starting position of the match. Let's add a third value to the tuple which will be the frequency of a table/column/view. This means when the user types `SELECT * FROM ` the list that pops up will have the table names listed in the most commonly used names. As soon as they start typing it'll reorder the list based on their typing and use the frequency item in the tuple to break ties. 

Only the keywords should be pre-populated from the history file. The table/column/view names should only learn from the start of a session and should not persist between sessions. The reason being, each session will be a specific task which requires a set of tables to be accessed but it'll vary from session to session. We should also reset these frequency counters for tables/cols/views whenever we change the database.

The end. :)

darikg:
> Right now we're only using this learning algorithm for Keywords. Why not use it for tables/columns/views etc?

Good call. This would also be a good time to tweak how sorting is done. Right now `find_matches` sorts within suggestion categories, but not across them. So the completion menu lists all matching tables, then views, then schemas, etc. Instead, sorting should be done after all matches have been found, based on the three-element tuple you just described. So this is kinda dove-tailing with discussion in #377 that we should remove sorting from `find_matches`

darikg:
Ok wow this got complicated quickly. Quick overview of the most recent set of changes:

- We have a new package `priorization`, with one class `PrevalenceCounter`
    - PrevalanceCounter counts keyword and identifier name prevalences separately

- `PGCompleter` has a member `prioritizer`, which is initialized to an instance of `PrevalenceCounter`, but like @amjith said, this could become some other type of prioritizer like a neural network

- `pgcompleter` has a new class `Match`, which is a simple namedtuple (`completion`, `priority`)

- `find_matches` now returns a list of matches, instead of a list of completions.
    - The list is unsorted.
    - The priority is a nested tuple of `(fuzzy_match_tuple, prevalance)`
    - `find_match` arguments `start_only` and `fuzzy` are collapsed into a single parameter `mode`
        - `mode` = 'keyword' ==> start_only matching and keyword prevalence
        - `mode` = 'name' ==> fuzzy matching and name prevalence
        - TODO there's a small ambiguity here with hardcoded functions and datatypes, which use start_only matching but are names, not keywords. 

  - matches are collapsed across all suggestion types, and then sorted by priority

  - Around this point I got sick of the linebreaks in `pgcompleter.get_completions()` and broke out everything into submethods `get_table_matches`, `get_schema_matches` etc. and use dictionary lookup to dispatch on suggestion type. This gives back two levels of indentation and was a big improvement in readability imho.

- On pgcli startup, we read the most recent 100 lines from file history and learn keyword (but not name) prevalence. This is done in the completion refresher background thread so it shouldn't hurt startup time

- Finally, learned prevalences are persisted across completion refreshes, but changing databases wipes out the learned name prevalences

j-bennet:
@darikg Wow, this looks great. I have a question about this part:

https://github.com/dbcli/pgcli/blob/darikg/learn-keyword-prefs/pgcli/packages/prioritization.py#L40

The comment says that we can't rely on sqlparse to recognize keywords, because sqlparse is database agnostic. But we're only parsing keywords out of our own history, right? so that would be database specific. Am I missing something?

darikg:
@j-bennet  So in order to count identifier usages I'm iterating over the input like this:

```
for parsed in sqlparse.parse(text):
    for token in parsed.flatten():
        if token.ttype in Name:
            self.name_counts[token.value] += 1
```
For symmetry & simplicity it would be nice to be able to also do:
```
        elif token.ttype in Keyword:
            self.keyword_counts[token.value] += 1
```

But that won't catch all keywords properly unless sqlparse knows to tokenize them as keyword. So prioritization.py has its own ad hoc keyword tokenization machinery.

pgcli/completion_refresher.py
```diff
@@ -68,6 +70,13 @@ def _bg_refresh(self, pgexecute, special, callbacks):
             # break statement.
             continue
 
+        # Load history into pgcompleter so it can learn user preferences
+        n_max_recent = 100
+        n_recent = history and min(n_max_recent, len(history))
+        if n_recent:
+            for recent in history[-n_recent:]:
```
amjith: This could be rewritten as:

```
n_recent = history and 100
for recent in history[-n_recent:]:
```

If you ask for last 100 items and history is only 10 items, python will give you only the 10 items. No need to be cautious about array out of bounds. 

Proof: 

```
>>> a = [1,2,3]
>>> a[-5:]
[1, 2, 3]
>>> a[-2:]
[2, 3]
>>> a[-1:]
[3]
```

pgcli/pgcompleter.py
```diff
@@ -43,6 +49,20 @@ def __init__(self, smart_completion=True, pgspecial=None):
 
         self.all_completions = set(self.keywords + self.functions)
 
+        self.suggestion_matchers = {
```
amjith: This should be made into a class var, since this is a static mapping of keywords to functions and it won't change between objects.

pgcli/pgcompleter.py
```diff
@@ -181,9 +209,18 @@ def find_matches(self, text, collection, start_only=False, fuzzy=True,
 
         text = last_word(text, include='most_punctuations').lower()
 
+        if mode == 'name':
+            fuzzy = True
+            priority_func = lambda x: self.prioritizer.name_count(x)
```
amjith: I think this could be made: 

```
priority_func = self.prioritizer.name_count
```

pgcli/pgcompleter.py
```diff
@@ -181,9 +209,18 @@ def find_matches(self, text, collection, start_only=False, fuzzy=True,
...
+        else:
+            fuzzy = False
+            priority_func = lambda x: self.prioritizer.keyword_count(x)
```
amjith: Same as above.

tests/utils.py
```diff
@@ -14,16 +14,8 @@ def db_connection(dbname=None):
...
-    CAN_CONNECT_TO_DB = True
-    SERVER_VERSION = conn.server_version
-    json_types = register_json_typecasters(conn, lambda x: x)
-    JSON_AVAILABLE = 'json' in json_types
-    JSONB_AVAILABLE = 'jsonb' in json_types
-except:
-    CAN_CONNECT_TO_DB = JSON_AVAILABLE = JSONB_AVAILABLE = False
-    SERVER_VERSION = 0
+CAN_CONNECT_TO_DB = JSON_AVAILABLE = JSONB_AVAILABLE = False
+SERVER_VERSION = 0
```
amjith: :fearful:

pgcli/main.py
```diff
@@ -425,25 +439,37 @@ def adjust_less_opts(self):
 
         return less_opts
 
-    def refresh_completions(self):
-        self.completion_refresher.refresh(self.pgexecute, self.pgspecial,
-                                          self._on_completions_refreshed)
+    def refresh_completions(self, history=None, persist_priorities='all'):
+        callback = functools.partial(self._on_completions_refreshed,
+                                     persist_priorities=persist_priorities)
```
amjith: Clever! :+1:

pgcli/main.py
```diff
@@ -425,25 +439,37 @@ def adjust_less_opts(self):
...
+            old_completer = self.completer
             self.completer = new_completer
+
+            if persist_priorities == 'all':
+                new_completer.prioritizer = old_completer.prioritizer
+            elif persist_priorities == 'keywords':
+                new_completer.prioritizer = old_completer.prioritizer
+                new_completer.prioritizer.clear_names()
+            elif persist_priorities == 'none':
+                pass
```
amjith: I don't follow the logic behind this if/elif/else block. Can you add a comment block?

amjith: Never mind, I'm able to follow it now. I still think a comment would be nice, since the function name says it's swapping completer objects and we actually do more than that inside the function.

pgcli/main.py
```diff
@@ -553,6 +580,19 @@ def need_completion_refresh(queries):
 
     return False
 
+
+def has_change_db_command(queries):
```
amjith: Can you change https://github.com/dbcli/pgcli/pull/379/files#diff-38f3aeda57742756afa677ca09dcea43R575 to use this new function? So we don't duplicate the same logic twice?

pgcli/pgcompleter.py
```diff
@@ -162,7 +190,7 @@ def reset_completions(self):
                            'datatypes': {}}
         self.all_completions = set(self.keywords + self.functions)
 
-    def find_matches(self, text, collection, start_only=False, fuzzy=True,
+    def find_matches(self, text, collection, mode='name',
```
amjith: Wouldn't it be better to have `mode` take a string called `'fuzzy'` instead of `'name'`? When `mode` is set to `'name'` is it not self-documenting, it adds a bit of cognitive load.

amjith:
This is quite the PR. :)

Nice job. I've left a bunch of suggestions inline. I'm not sure I understand why the utils was changed to disable all the database tests. 

Thanks for taking the time to tackle this. The next release of pgcli is gonna kick butt. :)

pgcli/pgcompleter.py
darikg: That's true. But having class vars defined after method definitions looks weird to me though. Makes sense though

tests/utils.py
darikg: WOW that should not have been committed. Thank god for code review. My meagre excuse is I was working on a computer without postgresql installed, and for some reason psycopg2 just hangs trying to connect to it instead of timing out

pgcli/pgcompleter.py
darikg: Sure. I switched the mode names to fuzzy/strict instead of name/keyword

darikg:
>I'm not sure I understand why the utils was changed to disable all the database tests.
Ugh I'm really sorry about that.

> Thanks for taking the time to tackle this. The next release of pgcli is gonna kick butt. :)
I was thinking actually we might want to do a new release before merging this, so it has some time to live in master before being released to the wild

amjith:
> I was thinking actually we might want to do a new release before merging this, so it has some time to live in master before being released to the wild. 

That's a valid request. I can get started with the release process unless one of you is interested in doing.

j-bennet:
@darikg Oh I see. I thought sqlparse covers all possible keywords, but I guess it only covers the rather general subset and it would not catch ones specific to postgres. Makes sense.

amjith:
Suggestions after the `WHERE` clause is not suggesting columns names anymore. 

1__2_0_python_-__pgcli___users_amjith_dropbox_code_python_pgcli___tmux_

The even more surprising aspect of this find is that we don't have a test for this simple case `SELECT * FROM table_name WHERE `.

amjith:
Sorry about the false alarm. The column names are available but they're buried deeper in the list.

This makes me wonder if mixing all the completions together and then sorting based on priority is the right way to go.

darikg:
That's happening because the prioritizer is loading keyword prevalence from your history and not names, so they're showing up with higher priority. There's a couple ways to handle this. 

1) Like I mentioned at the very top of this PR, it would be nice to improve keyword suggestion in general -- none of the suggestions in that screenshot are valid SQL. Keywords are suggested in where clauses because of some issues with `LIKE` dbcli/mycli#135 and `INTERVAL` #340. So we could try to figure out a more sophisticated way of suggesting them. (Possibility: a new suggestion type 'operator' which would suggest and, or, not, like, ilike, interval, etc., separate from keyword). 

2) Force keywords to the bottom of the list, allowing all other completion types to intermingle. Fuzzy matching returns a two-tuple for sorting. Strict matching also returns a two-tuple, but the second element is always zero, because it's unused. So we could just specify that second element to be negative infinity instead.

I went ahead and tried option 2 because it's so easy.

amjith:
The new implementation works much better. I think we can use this solution for now while we implement the second order suggestions. 

I did notice one weird bug which is in master as well as this PR: 

```
SELECT * FROM table_name WHERE abs
```

Until I type `abs` it showed a ton of suggestions that started with `abs` such as `abs, abstime, abstimeeq` etc. But as soon as typed `s` in `abs` the completion menu went away which is odd. I'm guessing it's a parsing bug. I haven't had time to dig in yet.

darikg:
One weird thing with function suggestions as of #357 is that some functions are double listed -- once in the hardcoded functions list, and once in the database metadata. Not sure if it's related but I think `abs` is one of those functions so it'd be worth checking to see that PR introduced that bug.

amjith:
I checked `abs` it's not listed in the functions in pgliterals. I think it is probably listed in sqlparse as a reserved word.

amjith:
Just found this article: http://nicolewhite.github.io/2015/10/05/improving-cycli-autocomplete-markov-chains.html

Markov chain based suggestion.

darikg:
That's awesome. It looks like she's looking only at `keyword -> keyword` transitions. It would be cool to also include 

- `keyword -> identifier` transitions 
    - Suggest columns foo and bar in the SELECT clause, but columns baz and qux in the WHERE clause

- `keyword -> identifier type`
    - E.g. if the user tends to alias qualify column names, suggest table aliases in the SELECT clause before column names

- `identifier -> identifier`
     - E.g. if you tend to select columns foo and bar together, `select foo, ` suggests bar

amjith:
Now that 0.20.0 is out, lets get this merged into master and give it a thorough testing. 

@darikg Can you rebase this branch to bring it up to date?

darikg:
I'll finish rebasing this soon

amjith:
Ping!

@darikg If you're busy I can take care of rebasing it. Just let me know.

darikg:
squashed & rebased

amjith:
Thanks @darikg. 

:bullettrain_front:

amjith:
We've had this merged into master for about 2 weeks now. How does everyone feel? 

Is it working out? 

When I type `SELECT * FROM table WHERE ` I get the list of columns at the top, but as soon as I start typing the columns gets drowned out by the list of keywords, which is jarring. 

Here's an example: 

1__0_2_python2_7_-__pgcli___users_amjith_dropbox_code_python_pgcli___tmux_

I don't have a great suggestion to counter this yet. But I'd like to get feedback from the team as well as some hardcore pgcli users about the current behavior. 

@dbcli/pgcli-core 

I'll start collecting a list of users who have been active in reporting issues or sending occasional PRs. If you have user suggestions, please leave their user name here and we'll contact them seeking feedback.

darikg:
@amjith that definitely seems wrong. I'll work on getting a test set up for it. But yeah, in general, I'd love more feedback.

PR #380, opened 2015-10-09
Consolidate the timing information.

### LINKED ISSUE #378: Timing does not display Command Time
My Pgcli install (0.19.1) always displays 0.000s for the Command Time. Format Time correctly displays times.

amjith:
Fixes #378.

Reviewer: @stuartquin


amjith:
Here is a solid way to test this: 

```
SELECT pg_sleep(2)
```

That should come back with a timing value of approximately 2 seconds.

stuartquin:
Managed to find some naturally long queries and all seems to work well, nice and tidy.

PR #381, opened 2015-10-10
Makes "enter" key function as the "tab" key only when the completion menu is displayed

mdsrosa:
New users get frustrated when they select a SQL command, press `Enter` and the command gets executed. So this PR makes the enter key work like the tab, but only when the completion menu is visible.

### DIFF
--- a/pgcli/filters.py
+++ b/pgcli/filters.py
@@ -1,0 +1,12 @@
+from prompt_toolkit.filters import Filter
+
+class HasSelectedCompletion(Filter):
+    """Enable when the current buffer has a selected completion."""
+
+    def __call__(self, cli):
+        complete_state = cli.current_buffer.complete_state
+        return (complete_state is not None and
+                complete_state.current_completion is not None)
+
+    def __repr__(self):
+        return "HasSelectedCompletion()"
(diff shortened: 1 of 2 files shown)


amjith:
:+1: I like this PR. 

I'll let someone else merge this PR. 

@j-bennet ?

amjith:
Actually it would be nice to check this on Windows before merging.

@darikg Would you mind giving it whirl? You've mentioned in the past that there is a subtle difference between the tab key and ctrl-space key in windows. So I want to make sure we didn't mess with anything by overriding the enter key.

j-bennet:
The PR looks good to me. I don't have a Windows machine to verify this, so I'll leave it to @darikg.

darikg:
this is great, thanks @mdsrosa 

@amjith the tab/control-space distinction isn't important to me any more, it was a stop-gap measure while prompt-toolkit on windows caught up

mdsrosa:
@amjith @j-bennet @darikg Thank you very much!
Landscape-bot is complaining about 3 problems that were found in the code. But I can't see what is wrong. Should I be worried?

amjith:
The problems reported by landscape are false positives. You don't have to worry about it.

PR #382, opened 2015-10-10
Improve error handling in pgexecute

darikg:
I don't like how errors are handled in pgexecute.run:

1. The default behavior with multiple statements, with errors in intermediate statements, is to continue executing the remaining statements. I'd like to be able to override that behavior to stop immediately on errors. 

2. Handling all exceptions makes it more difficult to reuse pgexecute outside of pgcli -- consumers may want to catch their own exceptions

3. \special commands are run outside the try clause. Because named queries can include arbitrary sql, they should be run like any other query

4. The `except` clause is too narrow. Currently, it catches `psycopg2.ProgrammingError`, which includes syntax errors, but not, for example, division by zero errors. See [here](http://initd.org/psycopg/docs/module.html#exceptions) for the psycopg2 exception hierarchy. I think ideally we'd widen all the way to catching `psycopg2.DatabaseError`, but `pgcli.main` relies on catching `psycopg2.OperationalError` in order to detect connection failures

This PR is one attempt at addressing these issues. It gives `pgexecute.run` and extra argument to specify how you want exceptions handled -- continue, stop, or reraise. I haven't done anything about point 4 though. Does catching `psycopg2.DatabaseError`, and always reraising `psycopg2.OperationalError` regardless of the `on_error` mode make sense?

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -31,7 +31,7 @@ import pgspecial as special
 from .pgcompleter import PGCompleter
 from .pgtoolbar import create_toolbar_tokens_func
 from .pgstyle import style_factory
-from .pgexecute import PGExecute
+from .pgexecute import PGExecute, ON_ERROR_RESUME, ON_ERROR_STOP
 from .pgbuffer import PGBuffer
 from .completion_refresher import CompletionRefresher
 from .config import write_default_config, load_config, config_location
@@ -81,6 +81,10 @@ class PGCli(object):
         self.syntax_style = c['main']['syntax_style']
         self.cli_style = c['colors']
         self.wider_completion_menu = c['main'].as_bool('wider_completion_menu')
+
+        on_error_modes = {'STOP': ON_ERROR_STOP, 'RESUME': ON_ERROR_RESUME}
+        self.on_error = on_error_modes[c['main']['on_error'].upper()]
+
         self.completion_refresher = CompletionRefresher()
 
         self.logger = logging.getLogger(__name__)
@@ -130,7 +134,7 @@ class PGCli(object):
         except IOError as e:
             return [(None, None, None, str(e))]
 
-        return self.pgexecute.run(query, self.pgspecial)
+        return self.pgexecute.run(query, self.pgspecial, on_error=self.on_error)
 
     def initialize_logging(self):
 
@@ -330,7 +334,8 @@ class PGCli(object):
                     res = []
                     # Run the query.
                     start = time()
-                    res = pgexecute.run(document.text, self.pgspecial)
+                    res = pgexecute.run(document.text, self.pgspecial,
+                                        on_error=self.on_error)
                     successful = True
                     output = []
                     total = 0
--- a/pgcli/pgclirc
+++ b/pgcli/pgclirc
@@ -44,6 +44,12 @@ syntax_style = default
 # for end are available in the REPL.
 vi = False
 
+# Error handling
+# When one of multiple SQL statements causes an error, choose to either
+# continue executing the remaining statements, or stopping
+# Possible values "STOP" or "RESUME"
+on_error = RESUME
+
 
 # Custom colors for the completion menu, toolbar, etc. 
 [colors]
--- a/pgcli/pgexecute.py
+++ b/pgcli/pgexecute.py
@@ -258,11 +258,14 @@ class PGExecute(object):
 
                 # Not a special command, so execute as normal sql
                 yield self.execute_normal_sql(sql)
-            except psycopg2.ProgrammingError as e:
+            except psycopg2.DatabaseError as e:
                 _logger.error("sql: %r, error: %r", sql, e)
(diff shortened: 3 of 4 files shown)


pgcli/main.py
```diff
@@ -81,6 +81,10 @@ def __init__(self, force_passwd_prompt=False, never_passwd_prompt=False,
         self.syntax_style = c['main']['syntax_style']
         self.cli_style = c['colors']
         self.wider_completion_menu = c['main'].as_bool('wider_completion_menu')
+
+        on_error_modes = {'STOP': ON_ERROR_STOP, 'RESUME': ON_ERROR_RESUME}
```
j-bennet: RAISE is missing here.

amjith: Nah. RAISE won't be required by main.py. RAISE is a state for pgexecute, because reraising the exception in main.py will crash everything.

amjith: Nope. I was mistaken. I think RAISE could be a valid case but I'm not sure how it'll be dealt with.

amjith:
> Does catching psycopg2.DatabaseError, and always reraising psycopg2.OperationalError regardless of the on_error mode make sense?

Yes. That makes perfect sense.

pgcli/main.py
amjith: I was right originally. The RAISE state is not used by pgcli's main module. It is designed to address point number 2, which is to use pgexecute in a context outside of pgcli.

j-bennet: Main uses this state to pass on to pgexecute, doesn't it? Also, if the user adds this to config:

```
[main]

on_error = RAISE
```

then pgcli will crash. Is that the behavior we want?

amjith: We don't advertise `RAISE` as a valid option in the config file. So if a user decides to change it to a string that is neither `STOP` nor `RESUME`, pgcli will crash at startup, which I think is a valid behavior. 

The reason it will crash at startup is because the line below this tries to use the string to look up a value from the `on_error_modes` dictionary, which won't have a valid entry for that key. So it will throw a `KeyError` and crash pgcli.

darikg:
Updated to do that then

amjith:
Thanks @darikg. 

I'm satisfied with the PR and I'm ready to merge. 

One more point of reference. Here's how psql does the equivalent of this: 

http://stackoverflow.com/questions/4480381/postgres-sql-fail-on-script-error

We don't support any of the `\set` commands and I don't think we can go down that path. 

But the idea of having a command line switch to enable/disable this feature is intriguing. At the very least we should provide a way to change the behavior at runtime, so people don't have to quit pgcli, change the config file and restart just so they can run a script (via `\i`). 

But that's an enhancement to this feature. So I'll create a new issue for that.

amjith:
@darikg I'd like to resurrect the discussion on this PR. 

Why do we have three states RESUME, STOP and RAISE? 

I think there should be only two states. RESUME and RAISE. Because as soon as we RAISE we're going to STOP.

Right now STOP swallows the error and colors it using Click returns the status message. I don't see the value in that. 

I think pgexecute shouldn't be in the business of coloring error messages.

The reason I'm bringing this up is because we've unintentionally broke https://github.com/dbcli/pgcli/blob/master/pgcli/main.py#L342. Even if the command fails, it'll always set the successful flag to `True`. 

I noticed this when @j-bennet brought up this corner case scenario in the PR https://github.com/dbcli/pgcli/pull/388#issuecomment-149317244.

darikg:
I agree with moving error message coloring out of pgexecute. But in the case of RESUME, how will the pgexecute user (in this case, pgcli) know whether or not to color the output? Check for missing headers?

Another way to do it would be to change pgexecute.run to accept only single sql statements, and always raising on error. This would put the responsibility of splitting sql and handling errors on the caller.

amjith:
Maybe add a new field to the return tuple which denotes success/failure. 

What do you think about getting rid of RAISE? When the option is set to STOP we always raise the error which acts as a stop.

darikg:
Alternatively pgexecute.run could have an additional argument `exception_formatter` which would be a callable that converts an exception into formatted output. 

But yeah ultimately we should collapse STOP and RAISE like you say

amjith:
Yes, `exception_formatter` is a nice idea. I'd be up for that.

j-bennet:
I would support the idea of pgexecute always raising an error and the caller handling it. Callable is the next best thing though, it gives the caller almost as much flexibility. But I'd rather make the callable optional, and raising an error default.

amjith:
@j-bennet Setting on_error to  `STOP` will always raise an exception. But setting on_error to `RESUME` will take a callable that will format the error. 

@darikg Are you doing this, or should I do it?

darikg:
I'll take a stab at it in a day or two if that's cool

amjith:
Yup. If I can release before Saturday, I'll be happy. 

Saturday is when I'm presenting pgcli at a meetup and I'd like to have the latest version out in PyPI.

Issue #355, opened 2015-09-17
Hide command line passwords from ps


joewalnes:
It's super convenient to be able to run pgcli like this:

```bash
$ pgcli 'host=somehost user=someuser password=somepassword'
# or
$ pgcli postgresql://someuser:somepassword@somehost
```

However, it enables other users on the same system to see the password by running `ps`, `top`, etc.

Some other clients (like the regular `mysql` client) have a cunning way to hide this: http://unix.stackexchange.com/questions/88665/how-does-ps-know-to-hide-passwords

And there's a Python package that wraps this up to make it easy: https://github.com/dvarrazzo/py-setproctitle

Although neither of these are foolproof (there's a timing attack due to a small window of opportunity between the process being started and the args being modified), it can help reduce the attack vector.

amjith:
Looks like you have a great solution to this. Would you be willing to take a stab at it and send a PR? I'll be happy to help in anyway possible.

joewalnes:
Sure! Looking at the code, I see two options for implementing this:

* Option 1: In the main `cli()` function, read `sys.argv` and look for anything that looks like a connection string in the entire cmd line, and hide it. This is the quickest solution.

* Option 2: Maybe this actual arg hiding should go inside the `click` library. Something like `@click.argument('database', obscure=lambda arg: do_something(arg), ...)`. This will be useful to more people.

Thoughts?

amjith:
That's a good observation. I think solution 1 is good for the short term and file an issue with click for the longer term solution. 

You're right it'll be useful for others.

jayzeng:
:+1: on the feature, I (personally) don't see the need to make it a cli argument.

amjith:
Fix released in 0.20.0. 

Please upgrade:

```
pip install -U pgcli
```

PR #387, opened 2015-10-18
Issue #355 Use setproctitle to hide command line passwords

stuartquin:
Hi Guys

First draft at solving https://github.com/dbcli/pgcli/issues/355
Would be good to get some feedback on the approach, particularly the regex's

One thing this doesn't cover is entries in shell history file, will do some research on this.

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -57,6 +57,7 @@ from collections import namedtuple
 Query = namedtuple('Query', ['query', 'successful', 'mutating'])
 
 
+
 class PGCli(object):
 
     def __init__(self, force_passwd_prompt=False, never_passwd_prompt=False,
@@ -522,15 +523,15 @@ def cli(database, user, host, port, prompt_passwd, never_prompt, dbname,
             '\thost: %r'
             '\tport: %r', database, user, host, port)
 
-    obscure_process_password()
+    obfuscate_process_password()
     pgcli.run_cli()
 
-def obscure_process_password():
+def obfuscate_process_password():
     process_title = setproctitle.getproctitle()
     if '://' in process_title:
         process_title = re.sub(r":(.*):(.*)@", r":\1:xxxx@", process_title)
     elif "=" in process_title:
-        process_title = re.sub(r"password=\S+", "password=xxxx", process_title)
+        process_title =  re.sub(r"password=(.+?)((\s[a-zA-Z]+=)|$)", r"password=xxxx\2", process_title)
 
     setproctitle.setproctitle(process_title)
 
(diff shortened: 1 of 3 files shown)


pgcli/main.py
```diff
@@ -520,8 +522,18 @@ def cli(database, user, host, port, prompt_passwd, never_prompt, dbname,
...
 
+    obscure_process_password()
     pgcli.run_cli()
 
+def obscure_process_password():
+    process_title = setproctitle.getproctitle()
+    if '://' in process_title:
+        process_title = re.sub(r":(.*):(.*)@", r":\1:xxxx@", process_title)
+    elif "=" in process_title:
+        process_title = re.sub(r"password=\S+", "password=xxxx", process_title)
```
amjith: This will break if the password has spaces in them. I tried it by changing my password locally with spaces in them and here's the output: 

```
amjith          67478   0.1  0.2  2499852  27260 s010  S+   11:10PM   0:00.34 /Users/amjith/.virtualenvs/pgcli/bin/python2.7 /Users/amjith/.virtualenvs/pgcli/bin/pgcli host=localhost user=amjith password=xxxx check'
```

I had set the password as `amjith check` the first portion was masked and the portion after the space is left alone.

pgcli/main.py
```diff
@@ -520,8 +522,18 @@ def cli(database, user, host, port, prompt_passwd, never_prompt, dbname,
             '\thost: %r'
             '\tport: %r', database, user, host, port)
 
+    obscure_process_password()
     pgcli.run_cli()
 
+def obscure_process_password():
```
amjith: I prefer `obfuscate_process_password` over `obscure_process_password` since it is hard to discern whether obscure is used as a noun or a verb at first glance.

pgcli/main.py
```diff
@@ -520,8 +523,18 @@ def cli(database, user, host, port, prompt_passwd, never_prompt, dbname,
...
 
+    obfuscate_process_password()
     pgcli.run_cli()
 
+def obfuscate_process_password():
+    process_title = setproctitle.getproctitle()
+    if '://' in process_title:
+        process_title = re.sub(r":(.*):(.*)@", r":\1:xxxx@", process_title)
+    elif "=" in process_title:
+        process_title =  re.sub(r"password=(.+?)((\s[a-zA-Z]+=)|$)", r"password=xxxx\2", process_title)
```
amjith: Nice! I was trying to think of a way to enhance the regex to address the concern I raised. But I couldn't come up with something reasonable. So I'm glad you were able to find an alternate regex that works.

amjith:
Thanks for addressing this. :cherries:

PR #388, opened 2015-10-19
Reset completions when database is changed.

amjith:
Reviewer: @j-bennet 

This is a corner case scenario in the background completion refresh. When a user changes to a new database using `\c` and the completion is happening in the background, we still retain the tables and views from the old database and the completion suggestions show old database entries. 

This PR will address this by resetting the completion data-structure before triggering the completion refresh.

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -566,7 +566,8 @@ def need_completion_reset(queries):
     for query in sqlparse.split(queries):
         try:
             first_token = query.split()[0]
-            return first_token.lower() in ('use', '\\c', '\\connect')
+            if first_token.lower() in ('use', '\\c', '\\connect'):
+                return True
         except Exception:
             return False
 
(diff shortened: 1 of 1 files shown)


j-bennet:
Thinking about a corner case for your corner case: user says "\c non_existent_db_name". Sqlparse won't hiccup because it's valid sql. Database, however, won't be changed.

Am I overthinking it?

amjith:
Nope. That's a valid case. I'll fix that.

j-bennet:
Looks good. Going to merge.

amjith:
I should have left a comment that it is incomplete.

It doesn't fix the issue because pgexecute is not raising exceptions for failed queries anymore. 

I started a discussion in one of the closed PRs about what to do about it. 

ref: #382

j-bennet:
Yikes.

Shall I revert?

amjith:
Nope. No need to revert. We'll roll with it. :smiley:

PR #390, opened 2015-10-19
Change the on_error default value to STOP.

amjith:
@darikg Can you review and merge? 

I'd like to set the default to STOP since that is the conservative (aka "sane") option.


darikg:
:+1: for sane defaults

PR #391, opened 2015-10-19
Check for headers in \x auto mode.

amjith:
Reviewer: @stuartquin 

Related to #369. 

Check for the headers before using expanded mode when `\x auto` is enabled.


stuartquin:
Good spot :+1:

PR #395, opened 2015-10-22
Tweak on error options

darikg:
As discussed in #382. Split the RESUME, STOP, RAISE on_error enum into two separate input arguments: `exception_formatter`, a callable that accepts an Exception and returns a (title, rows, headers, status) tuple as if it were a valid query result, and `on_error_resume`, a boolean. If you want `pgexecute.run` to always raise an exception, simply don't supply an `exception_formatter`.

It occurs to me now that the input arguments `on_error_resume` and `exception_formatter` could be moved to instance variables of `pgexecute` instead being being function arguments but that seems like mostly a matter of taste so I'll leave it open for discussion first.

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -36,7 +36,7 @@ import pgspecial as special
 from .pgcompleter import PGCompleter
 from .pgtoolbar import create_toolbar_tokens_func
 from .pgstyle import style_factory
-from .pgexecute import PGExecute, ON_ERROR_RESUME, ON_ERROR_STOP
+from .pgexecute import PGExecute
 from .pgbuffer import PGBuffer
 from .completion_refresher import CompletionRefresher
 from .config import write_default_config, load_config, config_location
@@ -88,8 +88,7 @@ class PGCli(object):
         self.cli_style = c['colors']
         self.wider_completion_menu = c['main'].as_bool('wider_completion_menu')
 
-        on_error_modes = {'STOP': ON_ERROR_STOP, 'RESUME': ON_ERROR_RESUME}
-        self.on_error = on_error_modes[c['main']['on_error'].upper()]
+        self.on_error = c['main']['on_error'].upper()
 
         self.completion_refresher = CompletionRefresher()
 
@@ -340,8 +339,9 @@ class PGCli(object):
                     res = []
                     # Run the query.
                     start = time()
+                    on_error_resume = self.on_error == 'RESUME'
                     res = pgexecute.run(document.text, self.pgspecial,
-                                        on_error=self.on_error)
+                                        exception_formatter, on_error_resume)
                     output = []
                     total = 0
                     for title, cur, headers, status in res:
@@ -616,6 +616,12 @@ def quit_command(sql):
             or sql.strip().lower() == 'quit'
             or sql.strip() == '\q'
             or sql.strip() == ':q')
+
+
+def exception_formatter(e):
+    msg = click.style(utf8tounicode(str(e)), fg='red')
+    return None, None, None, msg
+
 
 if __name__ == "__main__":
     cli()
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -619,8 +619,7 @@ def quit_command(sql):
 
 
 def exception_formatter(e):
-    msg = click.style(utf8tounicode(str(e)), fg='red')
-    return None, None, None, msg
+    return click.style(utf8tounicode(str(e)), fg='red')
 
 
 if __name__ == "__main__":
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -57,8 +57,18 @@ from psycopg2 import OperationalError
 from collections import namedtuple
 
 # Query tuples are used for maintaining history
-Query = namedtuple('Query', ['query', 'successful', 'mutating'])
-
+MetaQuery = namedtuple(
+    'Query',
+    [
+        'query',            # The entire text of the command
+        'successful',       # True If all subqueries were successful
+        'total_time',       # Time elapsed executing the query
+        'meta_changed',     # True if any subquery executed create/alter/drop
+        'db_changed',       # True if any subquery changed the database
+        'path_changed',     # True if any subquery changed the search path
+        'mutated',          # True if any subquery executed insert/update/delete
+    ])
+MetaQuery.__new__.__defaults__ = ('', False, 0, False, False, False, False)
 
 
 class PGCli(object):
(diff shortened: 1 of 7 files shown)


pgcli/pgexecute.py
```diff
@@ -263,15 +267,14 @@ def run(self, statement, pgspecial=None, on_error=ON_ERROR_RESUME):
                 _logger.error("traceback: %r", traceback.format_exc())
 
                 if (isinstance(e, psycopg2.OperationalError)
-                        or on_error == ON_ERROR_RAISE):
+                        or not exception_formatter):
```
amjith: Let's change this to `if (isinstance(e, psycopg2.OperationError) or or not on_error_resume):`. 

Reason being it is possible for us to provide an `exception_formatter()` but not use it because the option is set to `STOP`. A user might change this at runtime which will merely change the option to `RESUME` without having to worry about the `exception_formatter()`. 

Also using the presence of `exception_formatter` to determine whether or not to `raise` is not very explicit and it's not self documenting.

pgcli/pgexecute.py
```diff
@@ -263,15 +267,14 @@ def run(self, statement, pgspecial=None, on_error=ON_ERROR_RESUME):
...
                     # Always raise operational errors, regardless of on_error
                     # specification
                     raise
 
-                result = click.style(utf8tounicode(str(e)), fg='red')
-                yield None, None, None, result
+                yield exception_formatter(e)
```
amjith: exception_formatter() should only format the exception, so this should be `yield None, None, None, exception_formatter(e)`.

pgcli/pgexecute.py
```diff
@@ -263,15 +267,14 @@ def run(self, statement, pgspecial=None, on_error=ON_ERROR_RESUME):
...
 
-                if on_error == ON_ERROR_STOP:
+                if not on_error_resume:
```
amjith: I dont' think this break is necessary, since the raise up above will break us out of the loop.

pgcli/pgexecute.py
```diff
@@ -226,12 +220,22 @@ def _json_typecaster(self, json_data):
         else:
             return json_data
 
-    def run(self, statement, pgspecial=None, on_error=ON_ERROR_RESUME):
+    def run(self, statement, pgspecial=None, exception_formatter=None,
```
amjith: The default argument for `exception_formatter` should be `repr`. The exception_formatter is an optional function that does coloring, when one is not provided we should do the bare minimum which is convert the exception to it's default string representation.

amjith:
This is very close to perfection. 

Summary of the comments I've left inline: 

The `exception_formatter` is an optional function that let's you go above and beyond with the exception format, in the default case we should format the exception using `repr` so it returns the default representation of exception.

We should not overload the meaning of `exception_formatter` to decide whether to raise an exception or not. That behavior is controlled by the `on_error_resume` boolean.

darikg:
I don't agree with making on_error_resume=False raise exceptions for a couple of reasons. The main one is that flipping the boolean requires completely rewriting the call site.

Compare:

```results = list(executor.run(sql, on_error_resume=True, ...))```

vs

```
results = []
try:
    for r in executor.run(sql, on_error_resume=False, ...):
        results.append(r)
except psycopg2.DatabaseError as e:
    results.append((None, None, None, str(e))
````

Doesn't that seem extreme? It's even weirder with the exception formatter argument.

```
results = []
try:
    for r in executor.run(sql, on_error_resume=False,
                                    exception_formatter=my_exception_formatter):
        results.append(r)
except psycopg2.DatabaseError as e:
    # Even though we supplied an exception formatter to the run function, it's
    # not used and we need to apply it ourselves
    results.append((None, None, None, my_exception_formatter(e))
````


> Also using the presence of exception_formatter to determine whether or not to raise is not very explicit and it's not self documenting.

The default value of exception_formatter is None, so the default behavior is 'if an error occurs, raise an exception' which I think is very much in the spirit of the zen of python (Errors should never pass silently.    Unless explicitly silenced.) The user has to explicity choose how to silence them. We could certainly rename `exception_formatter` to `exception_handler` instead if that seems more self-documenting.

**Comment by amjith** [comment]
> The main one is that flipping the boolean requires completely rewriting the call site.

I actually think the other way around. Take our existing scenario in pgcli. We currently do the following in main.py (which is the call site). 

```
res = pgexecute.run(document.text, self.pgspecial, exception_formatter, on_error_resume)
```

We're going to provide the user with an option to switch the on_error behavior at runtime. So when they switch from STOP to RESUME, we have to change two variable `on_error_resume` to `False` and `exception_formatter` to `None`. 

The examples you showed me were extreme cases. The second snippet with the `try .. except` clause will be the normal use case if we want to provide the user with the ability to switch from one mode to another. 

I prefer to have one variable to be the deciding authority of mode.

**Comment by darikg** [comment]
> We're going to provide the user with an option to switch the on_error behavior at runtime. So when they switch from STOP to RESUME, we have to change two variable on_error_resume to False and exception_formatter to None.

Ok, we have a miscommunication somewhere, because this is not correct. Only on_error_resume has to change -- exception_formatter remains the same. Regardless of the value of on_error_resume, exception_formatter is used to make a 'nice' result. The only difference in behavior is whether subsequent queries are run.

**Comment by amjith** [comment]
You're right there is a miscommunication (or a misunderstanding). 

I'm operating under the assumption that I would like pgexecute to only have two modes of behavior. Those modes are 'RESUME' and 'STOP'. In the 'STOP' mode whenever an error is encountered it will raise an exception, which will serve as a way to stop execution and the rest of the statements won't be executed because an exception is thrown. 

In the 'RESUME' mode whenever an error is encountered, we will use an exception_formatter to prettify the error and continue executing the rest of the commands. If exception_formatter is not provided then we will use 'repr' as the default exception_formatter.

But I believe you're working under the assumption that even during the STOP mode pgexecute will format the exception and return a predictable tuple, so main.py doesn't have to deal with exceptions. Is that correct? If so, that doesn't solve our problem that I raised originally where the `success` variable in main.py is always set to true since we swallow the exception in pgexecute.

**Comment by darikg** [comment]
That's a fair assessment, thanks.

But in either my version or yours I don't see how the `success` variable is handled properly in the on_error_resume=True case. Really the semantics of a single boolean representing the result of a multi-query are ill-defined. It might be better to have an array of booleans so we could distinguish `all(successful)` versus `any(not successful)`

**Comment by amjith** [comment]
You are right. We can't do it with a single boolean. 

But it also requires a much bigger refactor because we're currently saving off the entire multi-statment if it was successful. https://github.com/dbcli/pgcli/blob/master/pgcli/main.py#L420

So I want to step back and analyze the use cases for this. 

1. This is used by the ipython integration to decide whether to save the query or not. We need a single boolean for that (perhaps `all(successful)`).

2. We need to decide if we have to do a completion refresh. This needs a more granular control, where we check to see if an `ALTER` or `DROP` or `CREATE` statement was successful or not and kick off the completion refresh. Although one could argue that kicking off the completion even if the statement is unsuccessful doesn't hurt anything. 

The only scenario in which we will run into a problem is if the user decides to do `select * from table1; \c database2; select * from table2`. In that case we absolutely must know whether the `\c` command in the middle was successful or not, because if it is successful, then we need to drop the existing completion data-structures and then kick off a new completion refresh in the background.

Resetting the existing data-structures is only done when the user changes databases, if we don't reset it, pgcli will continue to suggest older database completions while the new refresh is happening in the background. If the command was not successful and we preemptively reset the datastructures, then until the newly kicked off refresh is done, there won't be any suggestions.

So it is safe to say this needs a fair bit of overhauling to get it right.

**Comment by darikg** [comment]
Yeah. Let's try to get *#*2 right first, and then figure out the ipython stuff. 

My first thought is increase the tuple returned by `run` to include a `success` boolean. (I think you suggested this already earlier on in the PR). Since `run` is splitting the sql, we might as well also return the split statement, so we don't need to resplit outside of `run`. So as a gross sketch we'd have something like

```
db_changed = False
needs_refresh = False
results = executor.run(document.text, ...)
for title, cur, headers, status, success, sql in results:
    db_changed = db_changed or (success and has_db_change_cmd(sql))
    needs_refresh = needs_refresh or (success and has_mutating_cmd(sql))
    ...
```

Does that make sense?

**Comment by amjith** [comment]
Yup. That sounds good to me.

**Comment by darikg** [comment]
I've updated this to reflect our latest round of discussion. I ended up refactoring 'pgcli.run_cli' into a couple extra submethods which I think was an improvement in readability.

File: pgcli/main.py [hunk 5]
```diff
@@ -423,6 +349,133 @@ def prompt_tokens(cli):
...
+
+            cli = CommandLineInterface(
+                application=application,
+                eventloop=create_eventloop())
+
+            return cli
+
+    def _evaluate_command(self, text):
+        """Used to run a command entered by the user during CLI operation
+        (Puts the E in REPL)
```
amjith: LOL!!

pgcli/main.py
```diff
@@ -54,8 +54,18 @@
...
+    [
+        'query',            # The entire text of the command
+        'successful',       # True If all subqueries were successful
+        'total_time',       # Time elapsed executing the query
+        'meta_changed',     # True if any subquery executed create/alter/drop
+        'db_changed',       # True if any subquery changed the database
+        'path_changed',     # True if any subquery changed the search path
+        'mutated',          # True if any subquery executed insert/update/delete
+    ])
+MetaQuery.__new__.__defaults__ = ('', False, 0, False, False, False, False)
```
amjith: Good choice on using named tuple for this.

pgcli/main.py
```diff
@@ -423,6 +349,133 @@ def prompt_tokens(cli):
...
+        db_changed = False
+        path_changed = False
+        output = []
+        total = 0
+
+        # Run the query.
+        start = time()
+        on_error_resume = self.on_error == 'RESUME'
+        res = self.pgexecute.run(text, self.pgspecial,
+                                 exception_formatter, on_error_resume)
```
amjith: I noticed that pgexecute is not updated, it still relies on the presence of `exception_formatter` to decide whether to raise an exception or not. 

Which means `pgexecute` will never raise an exception because we're always setting the `exception_formatter`.

amjith:
Big thumbs up on breaking run_cli into smaller functions.

pgcli/main.py
darikg: I've grown to really, really like namedtuples

pgcli/main.py
darikg: correct.

amjith: If we're not going to change this behavior we should remove the `except Exception:` clause from main.py. Since that is misleading.

amjith:
Can you rebase? Looks like there are merge conflicts.

pgcli/main.py
amjith: Removal of except clause can wait. As I see it, this PR adds value and I don't want to hold up any longer. As soon as you rebase, I'll do one final test and we can merge. We'll deal with the except clause another time.

darikg: To clarify: pgexecute only catches and formats psycopg2.DatabaseError exceptions. The main except Exception is there to prevent runtime exceptions originatin from our own codebase from crashing pgcli in the name of user friendliness.

darikg:
rebased

amjith:
Thank you!!

:lollipop:

j-bennet:
Congratulations @amjith @darikg, this is huge!

PR #398, opened 2015-10-24
Fix indentation problems with "explain". Seems to solve #313.

d33tah:
It's possible I broke something, I don't know the codebase.

### DIFF
--- a/pgcli/packages/tabulate.py
+++ b/pgcli/packages/tabulate.py
@@ -441,10 +441,10 @@ def _align_column(strings, alignment, minwidth=0, has_invisible=True):
 
     """
     if alignment == "right":
-        strings = [s.strip() for s in strings]
+        strings = [s for s in strings]
         padfn = _padleft
     elif alignment == "center":
-        strings = [s.strip() for s in strings]
+        strings = [s for s in strings]
         padfn = _padboth
     elif alignment == "decimal":
         decimals = [_afterpoint(s) for s in strings]
@@ -455,7 +455,7 @@ def _align_column(strings, alignment, minwidth=0, has_invisible=True):
     elif not alignment:
         return strings
     else:
-        strings = [s.strip() for s in strings]
+        strings = [s for s in strings]
         padfn = _padright
 
     if has_invisible:
--- a/pgcli/packages/tabulate.py
+++ b/pgcli/packages/tabulate.py
@@ -441,10 +441,8 @@ def _align_column(strings, alignment, minwidth=0, has_invisible=True):
 
     """
     if alignment == "right":
-        strings = [s for s in strings]
         padfn = _padleft
     elif alignment == "center":
-        strings = [s for s in strings]
         padfn = _padboth
     elif alignment == "decimal":
         decimals = [_afterpoint(s) for s in strings]
@@ -455,7 +453,6 @@ def _align_column(strings, alignment, minwidth=0, has_invisible=True):
     elif not alignment:
         return strings
     else:
-        strings = [s for s in strings]
         padfn = _padright
 
     if has_invisible:
--- a/tests/test_tabulate.py
+++ b/tests/test_tabulate.py
@@ -1,0 +1,14 @@
+from pgcli.packages.tabulate import tabulate
+from textwrap import dedent
+
+
+def test_dont_strip_leading_whitespace():
+    data = [['    abc']]
+    headers = ['xyz']
+    tbl, _ = tabulate(data, headers, tablefmt='psql')
+    assert tbl == dedent('''
+        +---------+
+        | xyz     |
+        |---------|
+        |     abc |
+        +---------+ ''').strip()


darikg:
Do you have any reproducible test cases? 

Here's a simple one:

Before:
```
postgres> select '        abc' abc
+-------+
| abc   |
|-------|
| abc   |
+-------+
SELECT 1
```

After:
```
postgres> select '        abc' abc
+-------------+
| abc         |
|-------------|
|         abc |
+-------------+
```

So that's an improvement!

@amjith Why is tabulate.py vendored in again? If we start improving our own version we're going to need to start testing it

amjith:
It was vendored in because I had to implement the `psql` format for it. It was merged in upstream but it wasn't released for a long time.

Since then we've made changes to it to accommodate various things such as replacing `None` with `<null>` which are pgcli specific and doesn't make sense to merge upstream. 

I'm open to adding tests for it.

d33tah:
Would the tests be necessary for the change to be merged in?

amjith:
@d33tah That's a good question. I don't think it is necessary, but its definitely a nice to have. 

But before we merge this, I'd at least like to test this PR with a bunch of different datatypes such as JSON, integers, floats, strings, strings with newline etc. 

So give us a day to go through the tests. 

@darikg might have a set to tests or items to check as well.

amjith:
@darikg You had asked for a reproducible example for indented output. 

You can use any join statement to see this behavior:

```
explain SELECT * FROM django_migrations m JOIN django_content_type c ON c.id = m.id;
```

psql:
```
                                     QUERY PLAN
------------------------------------------------------------------------------------
 Hash Join  (cost=1.61..3.25 rows=27 width=1702)
   Hash Cond: (m.id = c.id)
   ->  Seq Scan on django_migrations m  (cost=0.00..1.27 rows=27 width=1044)
   ->  Hash  (cost=1.27..1.27 rows=27 width=658)
         ->  Seq Scan on django_content_type c  (cost=0.00..1.27 rows=27 width=658)
(5 rows)
```

pgcli:
```
+----------------------------------------------------------------------------+
| QUERY PLAN                                                                 |
|----------------------------------------------------------------------------|
| Hash Join  (cost=1.61..3.25 rows=27 width=1702)                            |
| Hash Cond: (m.id = c.id)                                                   |
| ->  Seq Scan on django_migrations m  (cost=0.00..1.27 rows=27 width=1044)  |
| ->  Hash  (cost=1.27..1.27 rows=27 width=658)                              |
| ->  Seq Scan on django_content_type c  (cost=0.00..1.27 rows=27 width=658) |
+----------------------------------------------------------------------------+
EXPLAIN
```

Notice the absence of indentation on last line.

pgcli/packages/tabulate.py
```diff
@@ -441,10 +441,10 @@ def _align_column(strings, alignment, minwidth=0, has_invisible=True):
 
     """
     if alignment == "right":
-        strings = [s.strip() for s in strings]
+        strings = [s for s in strings]
```
darikg: Just for readability's sake, I think you should outright delete these lines. Since `strings` is already a list, `[s for s in strings]` just creates an identical list, which is redundant.

d33tah:
@darikg: Added a commit related to your comment.

darikg:
@d33tah A test would be great. It will help ensure that future changes to tabulate don't break this behavior. Here's one to get started -- just copy it to a file `test_tabulate.py` in the `tests` directory.

```
from pgcli.packages.tabulate import tabulate
from textwrap import dedent


def test_dont_strip_leading_whitespace():
    data = [['    abc']]
    headers = ['xyz']
    tbl, _ = tabulate(data, headers, tablefmt='psql')
    assert tbl == dedent('''
        +---------+
        | xyz     |
        |---------|
        |     abc |
        +---------+ ''').strip()
```

I think that would be more than enough for now, since its our first test for tabulate.

d33tah:
@darikg: Test added, see commit message.

darikg:
Thanks @d33tah! You can see from the green "All checks have passed" indicator that all the tests passed.

Also, it looks like explain plans are printing correctly now:

```
test> explain select * from foo join bar on foo.xx = bar.z
+--------------------------------------------------------------------+
| QUERY PLAN                                                         |
|--------------------------------------------------------------------|
| Merge Join  (cost=285.12..606.19 rows=20758 width=20)              |
|   Merge Cond: (foo.xx = bar.z)                                     |
|   ->  Sort  (cost=135.34..140.19 rows=1940 width=12)               |
|         Sort Key: foo.xx                                           |
|         ->  Seq Scan on foo  (cost=0.00..29.40 rows=1940 width=12) |
|   ->  Sort  (cost=149.78..155.13 rows=2140 width=8)                |
|         Sort Key: bar.z                                            |
|         ->  Seq Scan on bar  (cost=0.00..31.40 rows=2140 width=8)  |
+--------------------------------------------------------------------+
EXPLAIN
```

I'll leave this open for the rest of the day in case anyone has some other ideas for tabulate behavior to test for. It seems good to me though.

d33tah:
@darikg: https://travis-ci.org/dbcli/pgcli/jobs/87343667 - where in this build log do we have a confirmation that the test was run?

darikg:
Scroll down to this:
```
tests/test_completion_refresher.py ....
tests/test_expanded.py .
tests/test_function_metadata.py ......
tests/test_fuzzy_completion.py ..
tests/test_main.py ...
tests/test_naive_completion.py ....
tests/test_parseutils.py ................X........................................
tests/test_pgexecute.py ................ssss...................................................................
tests/test_pgspecial.py ...........
tests/test_smart_completion_multiple_schemata.py ..................
tests/test_smart_completion_public_schema_only.py ...............................
tests/test_sqlcompletion.py ............................................................x..................................................................
tests/test_tabulate.py .
```

You can see a single dot after test_tabulate.py. That's the single test in that file passing

d33tah:
Ah, pardon, missed it. Thanks.

amjith:
:+1: 

Tested it with various datatypes and satisfied with the change. 

:mushroom:

d33tah:
Why :mushroom: ? :D

amjith:
I like leaving random emojis to show my appreciation for your contribution. 

Mushroom == PowerUp

Super Mairo reference. :smile:

PR #400, opened 2015-10-26
Setproctitle should be optional in tests.

j-bennet:
Right now our tests fail in tox because of setproctitle.


amjith:
Looks good. 

:seedling:

PR #402, opened 2015-10-28
Pin codecov version in travis.yml

darikg:
A new version of codecov (1.6.0) was released to PyPi today and our tests on travis starting failing on python 2.6. The log shows some import error in behave, so I have no idea how pinning the codecov version to 1.5.1 fixes it, but it does.


amjith:
:octopus:

PR #404, opened 2015-10-29
Update README.rst

vinotheassassin:
Changed auto-completion to auto-completes in the features because it flows better.


amjith:
Thank you!

:rooster:

PR #407, opened 2015-11-05
Switch the location of log/history file based on OS.

amjith:
@j-bennet @darikg I made a reference implementation of what I had in mind. 

You're welcome to test this and merge or write your own implementation if you're not satisfied. 

I haven't tested this yet, sorry.

### DIFF
--- a/pgcli/config.py
+++ b/pgcli/config.py
@@ -6,9 +6,9 @@ from configobj import ConfigObj
 
 def config_location():
     if platform.system() == 'Windows':
-        return os.getenv('USERPROFILE') + '\AppData\Local\dbcli\pgcli\config'
+        return os.getenv('USERPROFILE') + '\\AppData\\Local\\dbcli\\pgcli\\'
     else:
-        return expanduser('~/.config/pgcli/config')
+        return expanduser('~/.config/pgcli/')
 
 def load_config(usr_cfg, def_cfg=None):
     cfg = ConfigObj()
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -555,15 +555,16 @@ def cli(database, user, host, port, prompt_passwd, never_prompt, dbname,
         os.makedirs(config_dir)
 
     # Migrate the config file from old location.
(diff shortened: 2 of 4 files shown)


j-bennet:
Wow you're fast @amjith. How about I apply this into my appveyor branch and we'll see if it fixes it.

j-bennet:
Yes it fixes the appveyor run problem - it's back to prompt_toolkit problem, just as expected:

https://ci.appveyor.com/project/j-bennet/pgcli/build/job/2gqkdnkftguytn6x

j-bennet:
@amjith This one should be updated too:

https://github.com/dbcli/pgcli/blob/amjith/fix_log_file_windows/pgcli/pgcompleter.py#L21

Otherwise I think the fix is good to go.

j-bennet:
I added another commit that:
* Updates the code that migrates old config with new path.
* Updates named queries loading from the new path.
@darikg or @amjith could you take a look for sanity check and merge?

darikg:
So `config_location()` returns something like `c:\users\username\AppData\Local\dbcli\pgcli\`. But everything that uses it writes `config_location() + 'config'`. What's the point of that `'config'` subfolder?

j-bennet:
@darikg So in new implementation `config_location` does NOT return full path to file. It returns directory name to where the file should be written. To that, we add `config`, `log`, or `history` to get full absolute path.

darikg:
Ahh, so 'config' isn't a subfolder name, it's the filename. Sorry for the misunderstanding

amjith:
I'm replying from my phone on 2G. I couldn't check the commit. So I'll
defer to @darikg to review.

@j-bennet Can you release the next version once this is merged?

Amjith

j-bennet:
Yes. It should be 0.20.1, correct?

amjith:
Yes.

darikg:
I tested it by renaming my current config folder, or recreating the original .pgclirc file to test the migration, and everything looks good

Issue #399, opened 2015-10-26
Pager breaks cli history with wide output? 


pik:
I'm not sure if there is a configurable setting to fix this but the default is insanely bad. Scrolling up I have dozens of screens that look like this (when using arrow keys to horizontal scroll the warped columns, which essential destroys the ability to scroll up and see query history). 

<img width="1680" alt="screen shot 2015-10-26 at 4 59 20 pm" src="https://cloud.githubusercontent.com/assets/10236920/10724939/27786b16-7c03-11e5-94c5-2d07bb6594a6.png">

amjith:
It is configurable. 

`pgcli` will honor your `$PAGER` env variable. So you can launch pgcli as follows: 

`PAGER=cat pgcli`

That sets the PAGER env var to `cat`. Which will pipe the output through `cat` instead of `less`, which would solve the history scrollback. 

Alternatively, you can set PAGER to `more` to still keep some of the functionality of scrolling but it won't do horizontal scrolling.

pik:
Thanks for the response. The problem isn't the horizontal scrolling in less, it's that the less output is written to the same display as pgcli instructions. If you run `psql` with the same settings (using PAGER=less) the output is contained in a separate display and is hidden after the user exists less (likewise ipython %edit magic etc.).

amjith:
Yes. I chose a specific set of `less` [options](https://github.com/dbcli/pgcli/blob/master/pgcli/main.py#L432) to make it behave the way it does right now. I completely understand that this won't be desirable to everyone. 

I'll take that into consideration for the future release and provide a way to override that behavior.

pik:
@amjith aren't the less defaults (`SXRF`) not being applied there? Would something like this fix it (let me know if I'm misunderstanding something): 

        less_opts = os.environ.get('LESS', '-SRXF')
        self.logger.debug('Original value for LESS env var: %r', less_opts)

        return less_opts

amjith:
The code as it is now, reads the current value set in `LESS` env var and stores it away. Then overrides the `LESS` env var by setting it to `-SRXF`. The reason I store it away is because I want to restore it to the original value chosen by the user upon exit. Done here: https://github.com/dbcli/pgcli/blob/master/pgcli/main.py#L427

The reasoning behind this decision is that an average user is unlikely to have set their `LESS` env var to a specific value.

So pgcli tries to set some sane values for LESS options. This has served well up until now.

[...]

PR #409, opened 2015-11-10
Enable user set LESS options and \pset pager config toggling

pik:
Here is an initial implementation for fixing #399 (in tandem with https://github.com/dbcli/pgspecial/pull/6).

This should provide correct correct behaviour for \pset pager off/on/always options - without breaking the format_output api and maintaining correct behaviour when used with `\x` options. 

I've extracted the content_exceeds_width calculation out of format_output to avoid doing it twice. Perhaps the very nice way to do would be to check every row during formatting and raise if it breaks on a condition (rather than just checking the first-row as in the previous implementation) but I don't know how easy this would be since table-formatting is done by a different library. 

I will add some additional tests if this seems like a reasonable implementation. (Travisci currently fails because `pgspecial.pager_config` does not exist).

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -60,7 +60,6 @@ from psycopg2 import OperationalError
 from collections import namedtuple
 
 # Query tuples are used for maintaining history
-
 MetaQuery = namedtuple(
     'Query',
     [
@@ -75,9 +74,7 @@ MetaQuery = namedtuple(
 MetaQuery.__new__.__defaults__ = ('', False, 0, False, False, False, False)
 
 # Default less opts
-LESS_DEFAULTS = '-SRXF'
-
-
+LESS_DEFAULTS = '-~SR'
 
 class PGCli(object):
 
@@ -320,7 +317,7 @@ class PGCli(object):
                 query = MetaQuery(query=document.text, successful=False)
 
                 try:
-                    output, query = self._evaluate_command(document.text)
+                    output, query, output_fits_screen = self._evaluate_command(document.text)
                 except KeyboardInterrupt:
                     # Restart connection to the database
                     self.pgexecute.connect()
@@ -342,7 +339,7 @@ class PGCli(object):
                     click.secho(str(e), err=True, fg='red')
                 else:
                     try:
-                        click.echo_via_pager('\n'.join(output))
+                        self.display_output(output, output_fits_screen)
                     except KeyboardInterrupt:
                         pass
 
@@ -382,6 +379,13 @@ class PGCli(object):
             # Don't persist pgcli less defaults when session is terminated
             if less_opts_adjusted:
                 os.environ['LESS'] = ''
+
+    def display_output(self, output, output_fits_screen=True):
+        if self.pgspecial.pager_config == special.main.PAGER_ALWAYS or \
+            (self.pgspecial.pager_config == special.main.PAGER_LONG_OUTPUT and not output_fits_screen):
+            click.echo_via_pager("\n".join(output))
+        else:
+            click.echo("\n".join(output))
 
     def _build_cli(self, history):
 
@@ -455,6 +459,7 @@ class PGCli(object):
         on_error_resume = self.on_error == 'RESUME'
         res = self.pgexecute.run(text, self.pgspecial,
                                  exception_formatter, on_error_resume)
+        screen_size = self.cli.output.get_size()
 
         for title, cur, headers, status, sql, success in res:
             logger.debug("headers: %r", headers)
@@ -470,7 +475,7 @@ class PGCli(object):
                     break
 
             if self.pgspecial.auto_expand:
-                max_width = self.cli.output.get_size().columns
+                max_width = screen_size.columns
             else:
                 max_width = None
 
@@ -495,7 +500,7 @@ class PGCli(object):
         meta_query = MetaQuery(text, all_success, total, meta_changed,
                                db_changed, path_changed, mutated)
 
-        return output, meta_query
+        return output, meta_query, output_fits_screen(output, screen_size)
 
     def _handle_server_closed_connection(self):
         """Used during CLI execution"""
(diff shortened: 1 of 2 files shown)


pgcli/main.py
```diff
@@ -366,9 +372,17 @@ def run_cli(self):
...
+            # Don't persist pgcli less defaults when session is terminated
+            if less_opts_adjusted:
+                os.environ['LESS'] = ''
+
+    def display_output(self, output, output_fits_screen=True):
+        if self.pgspecial.pager_config == special.main.PAGER_ALWAYS or \
+            (self.pgspecial.pager_config == special.main.PAGER_LONG_OUTPUT and not output_fits_screen):
+            click.echo_via_pager(output)
+        else:
+            print(output)
```
amjith: I'd prefer to use `click.echo(output)` since it tries to be a robust replacement for print(). http://click.pocoo.org/5/quickstart/#echoing

pgcli/main.py
```diff
@@ -456,16 +470,27 @@ def _evaluate_command(self, text):
                     click.secho("Aborted!", err=True, fg='red')
                     break
 
-            if self.pgspecial.auto_expand:
-                max_width = self.cli.output.get_size().columns
+            lines, columns = self.cli.output.get_size()
+            expanded = self.pgspecial.expanded_output
+            # Checking the width of only the first content row is fast but my be faulty
+            if cur and not expanded and content_exceeds_width(cur.fetchone(), columns):
```
amjith: This completely break `\x auto`. We can't decide based on the first row. That's why we use the return value from `tabluate` to check if it exceeds the width.

amjith:
@pik Thanks for taking a stab at this change. I like the addition of `\pset pager` to pgspecial. But I think it should only have two options `off/on`. 

If it is `on` then we check whether the env var `LESS` is set or not. If it is set then we don't mess with anything and we use it as is. If it is not set then we use `-RXF` if the output fits the screen width, and use `-SRF` if the output doesn't fit the screen width.

What do you think?

pik:
@amjith Thanks for the response and sorry for the long hiatus on this. I've updated to fix the issue with determining `output_fits_screen` for `\x auto`; it's a little weird because I didn't really want to alter the number of return values from format_output (not sure which is more preferable). 
At least currently there are checks in two places one as previously in format_output (determine whether it should be tabulated for `\x auto`) and the other after the output has been formatted (since formatting can increase the number of rows to not fit on screen, even if `\x auto` was triggered). 

I'd also like to add tests for display_output and a few of the functions called on PGCli class - but that would probably require a fair amount of stubbing. I'm not sure if there is a stubbing library or approach you would prefer? (I left it empty for now.)

> If it is on then we check whether the env var LESS is set or not. If it is set then we don't mess with anything and we use it as is. If it is not set then we use -RXF if the output fits the screen width, and use -SRF if the output doesn't fit the screen width

This would cover all of my personal use cases, my only hesitation would be if there are actually users which want to use custom LESS opts and have more control over that particular behaviour.

amjith:
@pik I somehow never got notified of the new commits. I'll take a look at them now.

pgcli/main.py
```diff
@@ -372,9 +375,17 @@ def run_cli(self):
 
         except EOFError:
             print ('Goodbye!')
-        finally:  # Reset the less opts back to original.
-            logger.debug('Restoring env var LESS to %r.', original_less_opts)
-            os.environ['LESS'] = original_less_opts
+        finally:
+            # Don't persist pgcli less defaults when session is terminated
+            if less_opts_adjusted:
+                os.environ['LESS'] = ''
```
amjith: It seems like you're clearing the LESS env var, but if the user had customized their LESS env var, we want to restore it when we exit not overwrite it with empty string.

pik: It's because `adjust_less_opts` sets defaults only on `if not os.environ.get('LESS'):` - so `less_opts_adjusted` is always false unless they had no LESS options set.

pgcli/main.py
```diff
@@ -73,6 +73,8 @@
     ])
 MetaQuery.__new__.__defaults__ = ('', False, 0, False, False, False, False)
 
+# Default less opts
+LESS_DEFAULTS = '-~SR'
```
amjith: What is this ~S option? I can't find it in the man pages.

pgcli/main.py
amjith: I see what you did there. Never mind my comment.

pgcli/main.py
```diff
@@ -653,6 +664,19 @@ def obfuscate_process_password():
...
+def output_fits_screen(output, screen_size):
+    for entry in output:
+        new_line_index = entry.find("\n")
+        if new_line_index == -1:
+            if len(entry) > screen_size.columns:
+                return False
+        else:
+            if new_line_index > screen_size.columns:
+                return False
+            if entry.count("\n") >= screen_size.rows:
```
amjith: Shouldn't this be len(output) >= screen_size.rows?

pgcli/main.py
pik: These are separate options: 
```
-S or --chop-long-lines
```
```
-~ or --tilde Normally lines after end of file are displayed as a single tilde (~).  This option causes lines after end of file to be displayed as blank lines.
```

pgcli/main.py
amjith: I believe `entry` is each row in the result. So counting the number of `\n` in each row to determine if we have exceeded the screen_size.rows won't work. Is that correct? 

That's why I'm suggesting `if len(output) >= screen_size.rows:` as an alternative.

pik: Usually an output looks like this after it has been formatted: 
` ['+------+-------+\n|   id | key   |\n|------+-------|\n|    1 | foo   |\n+------+-------+', 'SELECT 1']`

 (I check after it has been tabulated because tabulation may alter it based on padding or `\x auto`)

Now it occurs to me that this check will not work if output has more than the standard two entries though.. In that case `count("\n")` per entry should be added to a total and checked at the end of the for loop to see whether it exceeds `screen_size.rows`.

amjith:
We've carried over most of the changes from the PR in subsequent releases. I'm going to close this PR. Let us know if you have more suggestions.

PR #413, opened 2015-11-16
Use only immutable datatypes for suggestions

darikg:
The main goal here is to make it easier to write regression tests using set-based logic. Before, suggest_type returned lists of dictionaries, whose values were lists. Because mutable data structures aren't hashable, it wasn't possible to create sets of suggestions, and mandated the ugly `sorted_dicts` kludge. Here, I've reworked everything to return tuples instead of lists, and namedtuples instead of dicts. For the most part, I think this leads to more readable code, although admittedly one-element tuples are a bit of an eyesore.

### DIFF
--- a/pgcli/packages/sqlcompletion.py
+++ b/pgcli/packages/sqlcompletion.py
@@ -2,6 +2,7 @@ from __future__ import print_function
 import sys
 import re
 import sqlparse
+from collections import namedtuple
 from sqlparse.sql import Comparison, Identifier, Where
 from .parseutils import last_word, extract_tables, find_prev_keyword
 from pgspecial.main import parse_special_command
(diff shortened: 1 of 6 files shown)


pgcli/packages/sqlcompletion.py
```diff
@@ -106,13 +126,13 @@ def suggest_special(text):
 
     if cmd == text:
         # Trying to complete the special command itself
-        return [{'type': 'special'}]
+        return Special(),
```
amjith: I'd still prefer to have parens around single value tuples. It's way too easy to miss the trailing comma.

pgcli/packages/sqlcompletion.py
```diff
@@ -129,28 +149,27 @@ def suggest_special(text):
...
-            return [{'type': 'schema'},
-                    {'type': rel_type, 'schema': []}]
+            return Schema(), rel_type(schema=None)
 
     if cmd in ['\\n', '\\ns', '\\nd']:
-        return [{'type': 'namedquery'}]
+        return NamedQuery(),
 
-    return [{'type': 'keyword'}, {'type': 'special'}]
+    return Keyword(), Special()
```
amjith: Surrounding parens here as well. I know it's not a single value tuple but having the parens emphasizes that it's a tuple.

amjith:
I absolutely love this PR. :heart: 

The resulting code is not only succinct but way more clear in it's intent. I've always felt our list of dictionaries cobbled together with arbitrary fields was a bit icky. This is a very nice solution. 

Thank you for thinking of the future generations. :smiley:

Parenthesis around tuples is a personal preference, if you're not comfortable with that suggestion, just let me know and I'll merge this in as is.

darikg:
Thanks! I personally slightly prefer it without the parens, but it's not a strong preference by any means so I went ahead and changed them. While I was doing that I found two cases where I was still returning lists instead of tuples so it paid off.

amjith:
Looks great.

:fork_and_knife:

Issue #417, opened 2015-11-18
\x auto: list index out of range 


carymrobbins:
It seems that empty result sets with `\x auto` enabled cause a `list index out of range` exception.

```
crobbins> select 1 where false;         
+------------+                          
| ?column?   |                          
|------------|                          
+------------+                          
SELECT 0                                       
          
crobbins> \x                            
Expanded display is on.                            

crobbins> select 1 where false;         
                                        
SELECT 0                                
             
crobbins> \x auto                       
Expanded display is used automatically. 
                 
crobbins> select 1 where false;         
list index out of range                 
```

amjith:
Thanks for reporting the bug. It is indeed an oversight. I've created a PR (#419) to fix it. It should be fixed in the next release.

amjith:
This is now available in v1.0. Please upgrade. 

```
pip install -U pgcli
```

PR #419, opened 2015-11-22
Fix the bug in auto-expand mode when there are no rows to display.

amjith:
Reviewer: @j-bennet 

This is a fix for #417. 

I've also added some tests to prevent future failure.


j-bennet:
Fix verified. Nice of you to add the test @amjith.

Issue #415, opened 2015-11-17
completion for count and explain functions


lookforit:
I noticed that these 2 functions are not listed by `pgcli`.
How to add it by users?

amjith:
Yes `EXPLAIN` is not yet added to the list. We're in the process of refactoring how we supply these keywords. 

You can add it to this file and send a PR: https://github.com/dbcli/pgcli/blob/master/pgcli/packages/pgliterals/pgliterals.json

lookforit:
Thanks.
Looking forward these completions be included in next release.

amjith:
@lookforit I've added the EXPLAIN keyword in PR #420, but what do you mean by `COUNT` because it is already listed as a suggestion.

j-bennet:
Closed via https://github.com/dbcli/pgcli/pull/420.

PR #420, opened 2015-11-22
Add EXPLAIN keyword to the completion list.

amjith:
Reviewer: @stuartquin 

Trivial change to add EXPLAIN keyword to the suggestions. 

This is to address issue #415.


stuartquin:
Simple, and it works :+1:

PR #421, opened 2015-11-22
Handle manually entered double-quote gracefully

darikg:
Surprisingly, manually entering a double quote, as in `SELECT * FROM "` breaks suggestions. This just allows the suggestion system to ignore it.

### DIFF
--- a/pgcli/packages/parseutils.py
+++ b/pgcli/packages/parseutils.py
@@ -3,7 +3,7 @@ import re
 import sqlparse
 from collections import namedtuple
 from sqlparse.sql import IdentifierList, Identifier, Function
-from sqlparse.tokens import Keyword, DML, Punctuation, Token
+from sqlparse.tokens import Keyword, DML, Punctuation, Token, Error
 
 cleanup_regex = {
         # This matches only alphanumerics and underscores.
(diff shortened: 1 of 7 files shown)


amjith:
Good catch. Your fix works as expected.

I think there is more room for improvement. For example, this is a valid sql `SELECT * FROM "custom"."abc"`. In fact that is the default sql generated by most ORMs. 

But currently in pgcli if you type `SELECT * FROM "custom".` we get all the tables and schemas, not just the tables from the `custom` schema. 

I'm fine merging this as is if you'd like to handle the above mentioned case in a separate PR. Let me know what you prefer.

darikg:
Huh. I'm surprised that doesn't work. I vote to merge this in as in and handle that case down the road.

darikg:
wait sorry, don't merge this in, that's a new bug introduced by this pr not an outstanding issue

darikg:
Ok, I updated this to handle double quotes in a slightly more sophisticated way, so now `SELECT * FROM "custom"."` suggests things correctly. It's probably still a little user-surprising that the quotes get automatically erased unless they're strictly needed, but we can leave that for the future I think.

pgcli/packages/parseutils.py
```diff
@@ -254,6 +256,29 @@ def _parsed_is_open_quote(parsed):
...
+
+    :param word: string representing a (partially complete) identifier
+    :return: sqlparse.sql.Identifier, or None
+    """
+
+    p = sqlparse.parse(word)[0]
+    n_tok = len(p.tokens)
+    if n_tok == 1 and isinstance(p.tokens[0], Identifier):
+        return p.tokens[0]
+    elif (n_tok in (2, 3)) and p.token_next_match(0, Error, '"'):
```
amjith: Can you elaborate a little bit on what's going on here?

You're checking if the number of tokens is either 2 or 3 and the next match is an Error? 

Why not check if `n_tok` is greater than 1? Is there a specific reason for 2 or 3?

darikg:
I'd explain that token count check thing but I'm pretty sure it was incorrect anyway

amjith:
Looks good. :gem:

Issue #396, opened 2015-10-22
Humanize time display?


d33tah:
Here's what I just noticed:

```
+----------+
|    count |
|----------|
| 12237922 |
+----------+
SELECT 1
Time: 152.868s
```

Making it `humanize.time.naturaldelta(t) + (miliseconds) + "miliseconds"` would be much more readable. What do you think about it?

amjith:
I decided to settle on seconds as the default units because changing it between milliseconds, seconds and minutes will make it hard to compare them. 

For example, when someone runs a query and sees that it takes 1.5seconds then they optimize it somehow to make it run in 0.9s it'll be good to see 0.9 instead of 900ms. 

It adds a bit of cognitive overload to convert them between units.

So it was a conscious decision to keep it consistent. Does that answer your question?

d33tah:
You could keep both. I once wrote a private patch for psql that adds "humanized" time information next to the milisecond count. Having a uniform unit is nice unless you're going to read how many hours the query took based on the number of seconds. I tend to run long-running queries in PostgreSQL quite often.

j-bennet:
This was implemented and merged.

PR #422, opened 2015-11-24
Added humanized time display. Connect #396.

j-bennet:
This is a small change to add humanized time display in parentheses.

Examples:

```
Time: 0.006s (a moment)
Time: 10.005s (10 seconds)
Time: 120.011s (2 minutes)
```

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -341,8 +341,12 @@ class PGCli(object):
                         pass
 
                     if self.pgspecial.timing_enabled:
-                        print('Time: %0.03fs (%s)' % (query.total_time,
-                              humanize.time.naturaldelta(query.total_time)))
+                        # Only add humanized time display if > 1 second
+                        if query.total_time > 1:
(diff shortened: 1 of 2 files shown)


amjith:
The change works as expected. But I'm not sure I like the fact it says 'a moment' for time deltas under a second. 

I suspect that most of the sql commands execute in under a second and it's going to keep showing `a moment` in parenthesis for most of them. 

My preference would be to fork humanize and vendor it in inside our packages folder and make necessary edits to support sub-second formatting. 

What do you think?

j-bennet:
@amjith I'm thinking for sub-second queries, this humanized display does not really add any new information, so no point even forking the lib. How about I only show this display for queries that last more than a second?

amjith:
Yes. I think that's a good idea.

j-bennet:
Examples after new change:

```
Time: 0.502s
Time: 1.002s (a second)
Time: 10.009s (10 seconds)
```

amjith:
Looks good. :koala:

PR #423, opened 2015-11-25
Fix suggestions in multi-way joins

darikg:
`extract_tables` was stopping as soon as it hit the `ON` keyword in the first join clause

Thanks to @jamiely for reporting this issue


jamiely:
:+1:

amjith:
Works well. Thanks for adding comprehensive tests. 

Cheers!
:leaves:

PR #425, opened 2015-11-26
Really sort keywords after everything else

darikg:
As @amjith pointed out in #379 keywords don't sort properly after starting to type an identifier


amjith:
Looks good to me. Nice work.

:newspaper:

Issue #426, opened 2015-11-30
'ascii' codec can't decode in PostgreSQL function


dcere:
Hi,

I am having an encoding issue while using PostgreSQL functions. It looks like pgcli is trying to decode a string with non-ascii characters using the ascii codec. Steps to reproduce:

## Create test database
```
_$: createdb -O myuser testascii
_$: psql -U myuser testascii
testascii=> CREATE TABLE town
testascii-> (
testascii(>   name character varying(32)
testascii(> );

testascii=> \l
                                      List of databases
       Name        |  Owner   | Encoding |   Collate   |    Ctype    |   Access privileges   
-------------------+----------+----------+-------------+-------------+-----------------------
 postgres          | postgres | UTF8     | en_US.UTF-8 | en_US.UTF-8 | 
 template0         | postgres | UTF8     | en_US.UTF-8 | en_US.UTF-8 | =c/postgres          +
                   |          |          |             |             | postgres=CTc/postgres
 template1         | postgres | UTF8     | en_US.UTF-8 | en_US.UTF-8 | =c/postgres          +
                   |          |          |             |             | postgres=CTc/postgres
 testascii         | myuser   | UTF8     | en_US.UTF-8 | en_US.UTF-8 | 
```

## Insert test data
```
testascii=> INSERT INTO town VALUES ('Güell');
```

## Create function
```
testascii=> CREATE OR REPLACE FUNCTION select_towns(p_limit integer)
RETURNS SETOF record AS
'
SELECT "name"
FROM town
LIMIT p_limit;
'
LANGUAGE SQL
;

testascii=> SELECT select_towns(10);
 select_towns 
--------------
 (Güell)
(1 row)
```

## Test with pgcli
```
_$: pgcli --host dbserver --dbname testascii --user myuser
Password: 
Version: 0.20.1
Chat: https://gitter.im/dbcli/pgcli
Mail: https://groups.google.com/forum/#!forum/pgcli
Home: http://pgcli.com
testascii> SELECT * FROM town;
+--------+
| name   |
|--------|
| Güell  |
+--------+
SELECT 1
Time: 0.001s
testascii> SELECT select_towns(10);
'ascii' codec can't decode byte 0xc3 in position 2: ordinal not in range(128)

testascii> \q
Goodbye!
```

Since it works when using a simple `SELECT * FROM town` my guess is that the problem is because of the function, but could be something else.

[...]

PR #428, opened 2015-12-05
Register a unicode typecaster for record type.

amjith:
This is to address #426. 

Reviewer: @stuartquin 

There are many datatypes in Postgres and we have to register each datatype to be treated as unicode. This PR registers the `record` datatype to be treated as unicode. I wonder what other datatypes have to be explicitly registered.


stuartquin:
Looks good and works :+1:

Issue #424, opened 2015-11-26
TRUNCATE CASCADE notices not shown properly (only one per statement)


ThiefMaster:
First of all, this is the output in `psql`:

``` sql
indico=> TRUNCATE events.contributions, events.subcontributions, events.timetable_entries, events.sessions, events.session_blocks, events.breaks CASCADE;
NOTICE:  truncate cascades to table "session_principals"
NOTICE:  truncate cascades to table "session_block_principals"
NOTICE:  truncate cascades to table "contribution_principals"
NOTICE:  truncate cascades to table "contribution_field_values"
NOTICE:  truncate cascades to table "subcontribution_references"
NOTICE:  truncate cascades to table "contribution_references"
NOTICE:  truncate cascades to table "contribution_person_links"
NOTICE:  truncate cascades to table "legacy_contribution_id_map"
NOTICE:  truncate cascades to table "legacy_session_id_map"
TRUNCATE TABLE
indico=> BEGIN;
BEGIN
indico=> TRUNCATE events.contributions, events.subcontributions, events.timetable_entries, events.sessions, events.session_blocks, events.breaks CASCADE;
NOTICE:  truncate cascades to table "session_principals"
NOTICE:  truncate cascades to table "session_block_principals"
NOTICE:  truncate cascades to table "contribution_principals"
NOTICE:  truncate cascades to table "contribution_field_values"
NOTICE:  truncate cascades to table "subcontribution_references"
NOTICE:  truncate cascades to table "contribution_references"
NOTICE:  truncate cascades to table "contribution_person_links"
NOTICE:  truncate cascades to table "legacy_contribution_id_map"
NOTICE:  truncate cascades to table "legacy_session_id_map"
TRUNCATE TABLE
indico=> ROLLBACK;
ROLLBACK
```

Now, when I run the same thing in `pgcli`:

``` sql
indico> TRUNCATE events.contributions, events.subcontributions, events.timetable_entries, events.sessions, events.session_blocks, events.breaks CASCADE;
NOTICE:  truncate cascades to table "legacy_session_id_map"

TRUNCATE TABLE
Time: 0.185s
indico>
```

or inside a transaction:

``` sql
indico> BEGIN;
BEGIN
Time: 0.003s
indico> TRUNCATE events.contributions, events.subcontributions, events.timetable_entries, events.sessions, events.session_blocks, events.breaks CASCADE;
NOTICE:  truncate cascades to table "legacy_session_id_map"

TRUNCATE TABLE
[...]

PR #429, opened 2015-12-06
Concat and return all available notices

stuartquin:
A quick one, addresses #424

### DIFF
--- a/pgcli/pgexecute.py
+++ b/pgcli/pgexecute.py
@@ -288,12 +288,10 @@ class PGExecute(object):
         cur = self.conn.cursor()
         cur.execute(split_sql)
 
-        try:
-            title = ''
-            while len(self.conn.notices) > 0:
-                title = title + self.conn.notices.pop()
-        except IndexError:
-            title = None
+        # conn.notices persist between queies, we use pop to clear out the list
+        title = ''
+        while len(self.conn.notices) > 0:
+            title = title + self.conn.notices.pop()
 
         # cur.description will be None for operations that do not return
         # rows.
--- a/pgcli/pgexecute.py
+++ b/pgcli/pgexecute.py
@@ -289,7 +289,9 @@ class PGExecute(object):
         cur.execute(split_sql)
 
         try:
-            title = self.conn.notices.pop()
+            title = ''
+            while len(self.conn.notices) > 0:
+                title = title + self.conn.notices.pop()
         except IndexError:
             title = None
 


pgcli/pgexecute.py
```diff
@@ -289,7 +289,9 @@ def execute_normal_sql(self, split_sql):
         cur.execute(split_sql)
 
         try:
-            title = self.conn.notices.pop()
+            title = ''
+            while len(self.conn.notices) > 0:
+                title = title + self.conn.notices.pop()
```
amjith: I'm not sure why I originally decided to used pop() on the notices. But I think we have an opportunity to fix it. Let's make `title = '\n'.join(self.conn.notices)`.

I think that'll be faster and more idiomatic.

amjith: I now remember why I was using `pop`, because we're calling notices on the connection object and not the cursor object. So the notices are kept for the whole session unless we clear them. 

I believe your code does the right thing. My apologies.

pgcli/pgexecute.py
```diff
@@ -289,7 +289,9 @@ def execute_normal_sql(self, split_sql):
         cur.execute(split_sql)
 
         try:
```
amjith: Now that there is a while loop, let's get rid of the try/except clause. The while loop condition takes care of safe access.

stuartquin:
Yup, I originally tried joining but then found the notices following me about. `pop` lets us clear out the list

amjith:
Maybe we should document that with a comment so the future generations
aren't making the same mistake.

stuartquin:
Added a quick comment and force pushed

amjith:
Looks good to me. :goat: 

Cheers!

darikg:
@stuartquin Any idea how hard it'd be to print notices asynchronously while the query's running? Sometimes it's nice to have progress reports on slow running queries, e.g.

```
test> create function slow_warnings() returns void language plpgsql as $$
      begin
          for i in 1..5 loop
              perform pg_sleep(1);
              raise warning 'i = %', i;
          end loop;
      end
      $$; 
```

Running `select slow_warnings() ` in psql will print the warnings out one at a time while the query sleeps, but pgcli has to wait until the query finishes and prints them all at once.

stuartquin:
@darikg Having Async notices actually makes a lot of sense. I'll create a new issue for it and start some investigation

Cheers

Issue #437, opened 2016-01-06
enum type unicode error in py2


GTxx:
Hi, I have some enum type in Chinese. When I query enum value or tables using these enums, unicode encoding error raises.
```sql
my_db> select unnest(enum_range(NULL::"level"));
'ascii' codec can't decode byte 0xe5 in position 0: ordinal not in range(128) 
my_db> select * from book;
'ascii' codec can't decode byte 0xe5 in position 0: ordinal not in range(128) 
```

this is the error log:
```log
2016-01-06 09:50:27,348 (7938/MainThread) pgcli.main ERROR - sql: u'select unnest(enum_range(NULL::"level"));', error: UnicodeDecodeError('ascii', '大', 0, 1, 'ordinal not in range(128)')
2016-01-06 09:50:27,350 (7938/MainThread) pgcli.main ERROR - traceback: 'Traceback (most recent call last):
  File "/home/my_home/.pyenv/versions/work_env/lib/python2.7/site-packages/pgcli/main.py", line 306, in run_cli
    output, query = self._evaluate_command(document.text)
  File "/home/my_home/.pyenv/versions/work_env/lib/python2.7/site-packages/pgcli/main.py", line 453, in _evaluate_command
    self.pgspecial.expanded_output, max_width)
  File "/home/my_home/.pyenv/versions/work_env/lib/python2.7/site-packages/pgcli/main.py", line 615, in format_output
    missingval='<null>')
  File "/home/my_home/.pyenv/versions/work_env/lib/python2.7/site-packages/pgcli/packages/tabulate.py", line 889, in tabulate
    ['\t'.join(map(_text_type, row)) for row in list_of_lists])
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe5 in position 0: ordinal not in range(128)
```
I think this might be similar with #426. In Python2, psycopg2 returns enum type as string object. So tabulate  fails encoding enum type value.

amjith:
Yes I think they are related. Can you try the latest version in master? 

You can install directly from master using `pip`. 

```
pip install https://github.com/dbcli/pgcli/archive/master.zip
```

GTxx:
Hi @amjith, latest version works for me. 
This issue is already discussed in #438. I think we can close this issue.

PR #438, opened 2016-01-06
convert enum type to unicode in tabulate. #437

GTxx:
I try to encode string  before it is passed to tabulate. It works for my issue.

### DIFF
--- a/pgcli/packages/tabulate.py
+++ b/pgcli/packages/tabulate.py
@@ -8,6 +8,7 @@ from collections import namedtuple
 from decimal import Decimal
 from platform import python_version_tuple
 from wcwidth import wcswidth
+from ..encodingutils import utf8tounicode
 import re
 
 
@@ -515,7 +516,7 @@ def _format(val, valtype, floatfmt, missingval=""):
         return "{0}".format(val)
     elif valtype is _binary_type:
         try:
-            return _text_type(val, "ascii")
+            return _text_type(val, "utf-8")
         except TypeError:
             return _text_type(val)
     elif valtype is float:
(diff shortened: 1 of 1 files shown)


amjith:
There are two ways to fix this. One we can register a type caster in psycopg2 to return us unicode values for enums instead of utf-8. Two we can do what you've done which is to handle it in tabulate. 

I've been resisting the urge to fix this issue in tabulate since we want psycopg2 to do the right thing. But I'm getting tired of playing whack a mole with psycopg2 and I think fixing it in tabulate is a much better option. 

@dbcli/pgcli-core Can you please share your thoughts?

amjith:
BTW, here's a way to reproduce the issue: 

```
CREATE TYPE mood AS ENUM ('sad', 'ok', 'happy', '日本語');
CREATE TABLE person (name TEXT, current_mood mood);
INSERT INTO person VALUES ('Moe', '日本語')
SELECT * FROM person;
```

I have a PR that registers a type caster for enum. https://github.com/dbcli/pgcli/pull/440

The question is should we register a type caster as done in PR #440 or should we use this PR to patch it in tabulate?

j-bennet:
@amjith I think that psycopg2 is the right place to fix this, because re-encoding every string in tabulate may affect performance. Why do you prefer the tabulate way?

amjith:
The reason for fixing it in tabulate would be, it'll be a blanket fix that'll handle all the datatypes. Right now we're adding special type casters for each datatype https://github.com/dbcli/pgcli/blob/master/pgcli/pgexecute.py#L16-L21. 

But we still manage to miss corner cases as reported here. But if we handle it in tabulate, then we won't need separate type casters and it'll future proof us from running into corner cases like this.

j-bennet:
@amjith I see what you mean. Do you have any concerns about performance? This is what I measured:
```
$ python -m timeit -s "from encodingutils import utf8tounicode" "utf8tounicode(u'日本語')"
1000000 loops, best of 3: 0.309 usec per loop
```
With 0.3s per conversion, what if we have 100K values to convert in table?

darikg:
@j-bennet I think that's 0.3 microseconds

j-bennet:
@darikg That sounds much better. :)

amjith:
Good point about overhead and thanks for doing the micro-benchmark to show some real numbers. I'm happy with the benchmark and I'm confident that this won't produce a noticeable lag in the print out.

j-bennet:
Agree, with performance concern out of the way, this approach looks good.

amjith:
@GTxx Thanks for the PR. I appreciate it. 

:snowflake:

PR #440, opened 2016-01-07
Register special type casters for enum types.

amjith:


j-bennet:
This needs a test case @amjith.

amjith:
I wasn't sure if we're going to merge this in or patch the issue in tabulate as done here #438. If we go with the other PR, we can add a test for it at that time.

amjith:
Closing this infavor of #438.

PR #449, opened 2016-01-25
fix execute_from_file after #395

### LINKED ISSUE #443: \i: run() got an unexpected keyword argument 'on_error'
Hi,

I was testing a query using the \i option and I got this error. Using psql's \i option work as expected, so I guess this is an error somewhere in pgcli's code. Steps to reproduce:

## Create test database
```
_$: createdb -O myuser testdb
_$: psql -U myuser testdb
testdb=> CREATE table employee
testdb-> (
testdb(>   name character varying(32);
testdb(> );
```

## Insert test data
```
testdb=> INSERT INTO employee VALUES('John Doe');
```

## Create query file
```
_$: echo "SELECT count(*) FROM employee;" > /tmp/test.sql
```

## Test with psql
```
testdb=> \i /tmp/test.sql
 count 
-------
     1
(1 row)
```

## Test with pgcli
```
pgcli --host dbserver --user myuser --dbname testdb
Password: 
Version: 0.20.1
Home: http://pgcli.com
testdb> \i /tmp/test.sql
run() got an unexpected keyword argument 'on_error'
testdb> \q
Goodbye!
```

I've taken a (very) quick look at the code base and I think it fails because of the difference between [this line](https://github.com/dbcli/pgcli/blob/master/pgcli/pgexecute.py#L229) and [this other line] (https://github.com/dbcli/pgcli/blob/master/pgcli/main.py#L160) but maybe you can take a more detailed look at this.

Thank you.

davidszotten:
fixes #443

### DIFF
--- a/pgcli/pgexecute.py
+++ b/pgcli/pgexecute.py
@@ -260,7 +260,11 @@ class PGExecute(object):
                     cur = self.conn.cursor()
                     try:
                         for result in pgspecial.execute(cur, sql):
-                            yield result + (sql, True)
+                            # e.g. execute_from_file already appends these
+                            if len(result) < 6:
+                                yield result + (sql, True)
+                            else:
+                                yield result
                         continue
                     except special.CommandNotFound:
                         pass
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -157,7 +157,10 @@ class PGCli(object):
         except IOError as e:
             return [(None, None, None, str(e))]
 
-        return self.pgexecute.run(query, self.pgspecial, on_error=self.on_error)
+        on_error_resume = (self.on_error == 'RESUME')
+        return self.pgexecute.run(
+            query, self.pgspecial, on_error_resume=on_error_resume
+        )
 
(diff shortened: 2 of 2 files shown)


pgcli/pgexecute.py
```diff
@@ -260,7 +260,11 @@ def run(self, statement, pgspecial=None, exception_formatter=None,
                     cur = self.conn.cursor()
                     try:
                         for result in pgspecial.execute(cur, sql):
-                            yield result + (sql, True)
+                            # e.g. execute_from_file already appends these
```
davidszotten: this is a bit ugly, could we do better?

amjith: I agree this is a bit ugly, but I'm not sure what else can be done here. I'm trying to think of a way to communicate from execute_from_file() that we successfully ran the query and got all the result tuple. But I can't think of a great way to do it.

amjith:
Thanks for taking the time to fix the issue.

Sorry about the late reply, I was really swamped all last week. 

:cake:

PR #452, opened 2016-02-02
Load default pager and its options on load from config file instead of manipulate environment variable

fernandomora:
I purpose to set the defaut pager "less -SRXF" in pgclirc config file in order to allow users have more flexibility in pager configuration.
In my case I want to have different pager configuration for pgcli/psql than for the rest of the system or use a pager filtered to colorize some expressions:
`pager= grcat ~/.grc/conf.psql | less -iMSx4FXRe`

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -75,8 +75,23 @@ MetaQuery = namedtuple(
     ])
 MetaQuery.__new__.__defaults__ = ('', False, 0, False, False, False, False)
 
-
 class PGCli(object):
+
+    def set_default_pager(self, config):
+        configured_pager = config['main'].get('pager')
+        os_environ_pager = os.environ.get('PAGER')
+
+        if configured_pager:
+            self.logger.info('Default pager found in config file: ' + '\'' + configured_pager + '\'')
+            os.environ['PAGER'] = configured_pager
+        elif (os_environ_pager):
+            self.logger.info('Default pager found in PAGER environment variable: ' + '\'' + os_environ_pager + '\'')
+            os.environ['PAGER'] = os_environ_pager
+        else:
+            self.logger.info('No default pager found in environment. Using os default pager')
+        # Always set default set of less recommended options, they are ignored if pager is 
+        # different than less or is already parameterized with their own arguments 
+        os.environ['LESS'] = '-SRXF'
 
     def __init__(self, force_passwd_prompt=False, never_passwd_prompt=False,
                  pgexecute=None, pgclirc_file=None):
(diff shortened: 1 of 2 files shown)


amjith:
@fernandomora That's a clever use of the set_pager() special command. I'm in favor of this idea. 

@j-bennet Can you please take a look at the failing behave tests? I can't reproduce it locally.

fernandomora:
I think the problem is that the initialization of PGCli is overwriting the default parameterization of "cat" pager for tests.

https://github.com/fernandomora/pgcli/blob/417b4fb0296487125925364d3f9a53a595b86879/tests/features/environment.py#L18
https://github.com/fernandomora/pgcli/blob/417b4fb0296487125925364d3f9a53a595b86879/pgcli/main.py#L101

In my local environment is also passing all tests, but in travis using less as pager seems to cause problems to pass them.

I think, despite it was more smarter using  better using set_pager(), we should leave its use to pager switching during execution and set environment "PAGER" environment variable on pgcli load in this preference order:
1. pager set on pgli config file (not configured in default configuration)
2. pager set on "PAGER" os environment variable

In this way, call to to "\pager" will reset properly the pager to that stored by default on pgcli load as pgspecial implementation was doing:
https://github.com/dbcli/pgspecial/blob/v1.2.0/pgspecial/main.py#L143-L148

I will refactor the PR with my proposal

pgcli/main.py
```diff
@@ -75,9 +75,24 @@
...
+            os.environ['PAGER'] = configured_pager
+        elif (os_environ_pager):
+            self.logger.info('Default pager found in PAGER environment variable: ' + '\'' + os_environ_pager + '\'')
+            os.environ['PAGER'] = os_environ_pager
+        else:
+            self.logger.info('No default pager found in environment. Using os default pager')
+        # Always set default set of less recommended options, they are ignored if pager is 
+        # different than less or is already parameterized with their own arguments 
+        os.environ['LESS'] = '-SRXF'
+
```
amjith: An alternate succinct implementation: 

```
def set_default_pager(self, config):
    os.environ['PAGER'] = os.environ.get('PAGER') or config['main'].get('pager')
```

I want to keep the default pager uncommented in the config file. That way if the user has setup a PAGER then we don't mess with it, but if they haven't then we give them a sane value for that option. This gives a more pleasant experience out of the box.

Also this new change might break existing pgcli behavior for people who already have pgcli installed.

fernandomora:
@amjith agree with the more concise implementation one, but the order you purpose is the opposite of what I am purposing, because when `os.environ.get('PAGER')` is defined in the environment (usually it is) pager of configuration file would never be used.

The reason of my PR is exactly the opposite:
`os.environ['PAGER'] =  config['main'].get('pager') or os.environ.get('PAGER')`
Despite having a PAGER defined in the environment for using along all the tasks of my system, I want to use a different pager, maybe with different options, exclusively for pgcli. The only place where I can put that exclusive configuration, to override system PAGER, is the config file.

I purpose tho leave the `pager = less '-SRXF'` line commented in pgclirc only for skeleton or documentation purposes in order to allow users to uncomment and edit it in `~/.config/pgcli/config` file

Current pgcli pager behavior:
1. If `PAGER` is defined in the environment, that pager is used
2. else default OS pager is used

New pgcli pager behaviour I was thinking in:
1. If `pager` is configured in loaded config (by default is commented in pgclirc, but you can have it editing `~/.config/pgcli/config`), that pager is used
2. else if `PAGER` is defined in the environment, that pager is used
3. else default OS pager

New pgcli pager behaviour I understand you are purposing:
1. if `PAGER` is defined in the environment, that pager is used
2. else if `pager` is configured in loaded config (it is always loaded because is in pgclirc), that pager is used

If we do it in the way I say, by default, until you explicitly edit config, this change does not break current pgcli behavior. The current use case of the tests `PAGER='cat'` would continue working properly since pager is not in config file and on the other hand you allow me to a new use case where I can have in pgcli a different pager configuration due its particular use than in the rest of the system.

amjith:
@fernandomora You're right. That makes more sense than my suggestion. 

I'll merge it in. Thanks for the PR. 

:roller_coaster:

PR #453, opened 2016-02-02
Handle a corner case for execute_from_file.

amjith:
Since we added two more parameters to the result tuple, we need to update the error case for execute_from_file() special functions. 

Reviewer: @darikg

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -150,12 +150,12 @@ class PGCli(object):
     def execute_from_file(self, pattern, **_):
         if not pattern:
             message = '\\i: missing required argument'
-            return [(None, None, None, message)]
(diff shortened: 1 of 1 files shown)


darikg:
Is this related to the discussion in #449, where pexecute is appending the extra two tuple elements?

I'm a little confused because I don't see a behavioral difference with or without this patch applied

amjith:
Yes it is related to #449. While I was reviewing the code in the PR I realized that the exception clauses did not set the boolean to false in the tuple that they were returning. This boolean is used for determining whether to save the last executed sql in case we're embedded in an IPython shell. 

So visibly there won't be a change unless we've launched it from IPython.

darikg:
Got it, thanks for clarifying

Issue #450, opened 2016-01-25
NULL values should be displayed as <null> when extended display mode is on


andyjack:
```
Version: 0.20.1
Chat: https://gitter.im/dbcli/pgcli
Mail: https://groups.google.com/forum/#!forum/pgcli
Home: http://pgcli.com
postgres> select NULL as nullval;
+-----------+
|   nullval |
|-----------|
|    <null> |
+-----------+
SELECT 1
Time: 0.001s
postgres> \x
Expanded display is on.
Time: 0.000s
postgres> select NULL as nullval;
-[ RECORD 0 ]-------------------------
nullval | None

SELECT 1
Time: 0.001s
postgres>
```
I'm on 9.4.5, if that's significant.

amjith:
This is now fixed in master. Can you try installing directly from master? 

```
pip install -U https://github.com/dbcli/pgcli/archive/master.zip
```

andyjack:
It is working, thank you!
```
postgres> \x
Expanded display is on.
Time: 0.000s
postgres> select null as nullvar;
-[ RECORD 0 ]-------------------------
nullvar | <null>

SELECT 1
Time: 0.001s
postgres> \x
Expanded display is off.
Time: 0.000s
postgres> select null as nullvar;
+-----------+
|   nullvar |
|-----------|
|    <null> |
+-----------+
SELECT 1
Time: 0.001s
```

PR #454, opened 2016-02-02
Display null values as <null> in expanded output.

amjith:
Reviewer: @stuartquin 

We display the `null` values as `<null>` in the tabular output, this PR makes pgcli do the same for expanded output as well. 

Addresses https://github.com/dbcli/pgcli/issues/450


stuartquin:
Looks good, does the job :+1:

PR #456, opened 2016-02-02
test for recent patch for execute_from_file

davidszotten:
to prevent obvious regressions at least

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -4,7 +4,10 @@ try:
     import setproctitle
 except ImportError:
     setproctitle = None
-from pgcli.main import obfuscate_process_password, format_output
+
+from pgcli.config import config_location
+from pgcli.main import obfuscate_process_password, format_output, PGCli
+from utils import dbtest, run
 
 
 @pytest.mark.skipif(platform.system() == 'Windows',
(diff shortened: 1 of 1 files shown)


amjith:
Nice!!

:lollipop:

amjith:
Hrm. Looks like travis didn't run the tests when the PR was opened. Now it is failing after I've merged it.

https://travis-ci.org/dbcli/pgcli/builds/106639518

tests/test_main.py
```diff
@@ -58,3 +61,15 @@ def test_format_output_auto_expand():
...
     assert expanded_results == expanded
+
+
+@dbtest
+def test_i_works(tmpdir, executor):
+    sqlfile = tmpdir.join("test.sql")
+    sqlfile.write("SELECT NOW()")
+    cli = PGCli(
+        pgexecute=executor,
+        pgclirc_file=config_location(),
```
amjith: I found out why it is failing. 

This config_location is a directory under ~/.config/pgcli, but we create that directory (if it doesn't exist) when pgcli is launched. https://github.com/dbcli/pgcli/blob/master/pgcli/main.py#L611

Since we're not launching pgcli, that directory is never created. 

The easy fix will be to call os.makedirs() in this test, but the better fix will be to do that os.makedirs() in the PGCli's __init__() method. Here: https://github.com/dbcli/pgcli/blob/master/pgcli/main.py#L92

Would you like to make this change?

davidszotten: see #458

Issue #448, opened 2016-01-25
year is out of range


fpietka:
I did it by mistake, but it turns out that some old dates can't be read through pgcli.
ie. `0001-01-01 00:00:00+00:09:21 BC`

now I have to update my field from psql to get pgcli working again :grin:

amjith:
That's odd. Unless psycopg2 has some built-in checks this shouldn't be the case. Can you send me the error message you get? 

Could you also send me the logs? You can find the logs either in ~/.pgcli.log or ~/.config/pgcli/log.

fpietka:
Here is the error log I get:
```
2016-01-26 14:04:16,835 (9567/MainThread) pgcli.main ERROR - sql: u"select to_date('1453734279.4203', 'U');", error: ValueError('year is out of range',)
2016-01-26 14:04:16,839 (9567/MainThread) pgcli.main ERROR - traceback: 'Traceback (most recent call last):
  File "/usr/local/Cellar/pgcli/0.20.1/libexec/lib/python2.7/site-packages/pgcli/main.py", line 306, in run_cli
    output, query = self._evaluate_command(document.text)
  File "/usr/local/Cellar/pgcli/0.20.1/libexec/lib/python2.7/site-packages/pgcli/main.py", line 453, in _evaluate_command
    self.pgspecial.expanded_output, max_width)
  File "/usr/local/Cellar/pgcli/0.20.1/libexec/lib/python2.7/site-packages/pgcli/main.py", line 615, in format_output
    missingval=\'<null>\')
  File "/usr/local/Cellar/pgcli/0.20.1/libexec/lib/python2.7/site-packages/pgcli/packages/tabulate.py", line 884, in tabulate
    list_of_lists, headers = _normalize_tabular_data(tabular_data, headers)
  File "/usr/local/Cellar/pgcli/0.20.1/libexec/lib/python2.7/site-packages/pgcli/packages/tabulate.py", line 583, in _normalize_tabular_data
    rows = list(tabular_data)
ValueError: year is out of range
'
```

stuartquin:
This seems like a known issue with psycopg2, it can be worked around with a custom extension like we do with json.
I'll take a look

amjith:
This is now fixed in master. Can you try installing from master? 

```
pip install -U https://github.com/dbcli/pgcli/archive/master.zip
```

fpietka:
Sorry for the delay. It doesn't seem to have changed.

Here is the log I got:

```
[...]

PR #457, opened 2016-02-03
Issue #448 extension for BC dates

stuartquin:
Addresses BC date issue by typecasting all `DATE` and `TIMESTAMP` results to string.


amjith:
Looks good. 

The test failures are caused by #456. It is now addressed by #458.

Thanks! 

:crossed_flags:

PR #458, opened 2016-02-03
ensure target dir exists when copying config

davidszotten:


amjith:
Close, but the tests are still failing because when we initialize logging it tries to write to `~/.config/pgcli/log` location. This is the default location specified in the `pgclirc` file. 

I think we should really be calling the `os.makedirs()` in the `__init__` method of the PGCli class. That will make sure the necessary setup is done before writing the config and the log file.

amjith:
:+1:

Thanks for sticking with it. :ice_cream:

PR #461, opened 2016-02-15
Use lexical order to break ties when fuzzy matching

drocco007:
Suppose we have a bunch of tables start with the name `user`, e.g. `user`, `user_group`, `user_action`, `user_address`, etc. When searching for the text `user`, where the match group length and position are identical for all of these items, the current master of pgcli produces completion results in arbitrary order:

pgcli arbitrary order

This PR adds a new priority fallback — after the fuzzy match group length, position, and history priority — such that, if the existing ranking fails to produce a total ordering, lexical order is used to break the ties. Here is the previous example with this change in place:

pgcli lexical order

### DIFF
--- a/pgcli/pgcompleter.py
+++ b/pgcli/pgcompleter.py
@@ -212,7 +212,7 @@ class PGCompleter(Completer):
         else:
             fuzzy = False
             priority_func = self.prioritizer.keyword_count
-            
+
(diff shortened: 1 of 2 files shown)


amjith:
Looks good to me. 

Thanks for addressing this case. 

:fountain:

PR #476, opened 2016-03-14
Upgrade to prompt_toolkit 0.60.

jonathanslenders:
Changes (see 0.58 .. 0.60.):
https://github.com/jonathanslenders/python-prompt-toolkit/blob/master/CHANGELOG

I also added a simple example of the continuation prompt. This shows dots in front of every next line, when the input is multiline. Feel free to change this.


amjith:
Works well. I like the continuation prompt. :)

:+1:

Issue #475, opened 2016-03-12
Suggestions after the FROM keyword are jumbled


amjith:
<img width="759" alt="1__1_0_python2_7_-__pgcli___users_amjith_dropbox_code_python_pgcli____tmux_" src="https://cloud.githubusercontent.com/assets/49260/13720553/2dadeb8a-e7c1-11e5-8455-fd6c9d8fa091.png">

The list of items suggested after the `FROM` keyword are not in any specific order. The list has alternating table names and function names. 

@darikg Any thoughts that jump out?  

 @drocco007 I remember you made a change to the fuzzy completion recently, is this behavior expected?

darikg:
Wow that's weird. If you delete your history file, does the effect persist?

amjith:
I deleted the history and started pgcli. Here's a screenshot. 

![1__1_0_python2_7_-__pgcli___users_amjith_dropbox_code_python_pgcli____tmux_](https://cloud.githubusercontent.com/assets/49260/13726589/f1436224-e883-11e5-9060-3dbfb8806712.png)

drocco007:
Interesting. I'll have a look over the PR later today :)

PR #477, opened 2016-03-19
Fix lexical order tiebreaking

ajlai:
Fix for #475.

The lexical priority was originally implemented as the `-index` of a match text within the lexically sorted collection in `find_matches`.
The problem is when `get_completions` combines results from multiple `find_matches` calls.
As the `-index` is the priority only within the scope of a single `find_matches` call, we end up with jumbling (especially in the case `SELECT * FROM `) because `get_completions` has only the lexical priority to do tie breaking, so it orders the completions as the first elem from all the get_X_matches, second elem from all get_X_matches, etc.

The new lexical_priority is properly comparable when we're combining multiple `find_matches` calls.
I've also added a test for making sure the `SELECT * FROM ` results are ordered.

It looks like this for me:
image

### DIFF
--- a/DEVELOP.rst
+++ b/DEVELOP.rst
@@ -12,19 +12,19 @@ the changes and create the commits in your local machine. Then push those
 changes to your fork. Then click on the pull request icon on github and create
 a new pull request. Add a description about the change and send it along. I
 promise to review the pull request in a reasonable window of time and get back
-to you. 
+to you.
 
 In order to keep your fork up to date with any changes from mainline, add a new
 git remote to your local copy called 'upstream' and point it to the main pgcli
 repo.
 
-:: 
+::
 
    $ git remote add upstream git@github.com:dbcli/pgcli.git
 
 Once the 'upstream' end point is added you can then periodically do a ``git
 pull upstream master`` to update your local copy and then do a ``git push
-origin master`` to keep your own fork up to date. 
+origin master`` to keep your own fork up to date.
 
 Local Setup
 -----------
@@ -60,7 +60,7 @@ working folder into the virtualenv. By installing it using `pip install -e`
 we've linked the pgcli installation with the working copy. So any changes made
 to the code is immediately available in the installed version of pgcli. This
 makes it easy to change something in the code, launch pgcli and check the
-effects of your change. 
+effects of your change.
 
 Adding PostgreSQL Special (Meta) Commands
 -----------------------------------------
@@ -90,7 +90,7 @@ Building RPM and DEB packages
 
 You will need Vagrant 1.7.2 or higher. In the project root there is a
 Vagrantfile that is setup to do multi-vm provisioning. If you're setting things
-up for the first time, then do: 
+up for the first time, then do:
 
 ::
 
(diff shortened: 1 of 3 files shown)


amjith:
Thank you for taking the time to dig in. I'm checking it out right now.

drocco007:
Hi @ajlai, thanks for looking at this.

This implementation treats non-alphanumeric characters as significant, so table names that require quoting will always appear at the top of the results (`-ord('"')` is -34):

image

I think we can fix this by tweaking the priority calculation slightly

    lexical_priority = tuple(-ord(c) for c in item if c.isalnum()) + (1,)

In other words, use only alphanumeric characters when computing the lexical tiebreaker. I ran a quick test which looks correct:

image

What do you think?

cc @amjith

drocco007:
Huh. One interesting side effect of my suggestion is that it changes the ordering of underscore-separated names:

image

Notice the `authorization` table, which, without the `if c.isalnum()` clause, ends up at the bottom of the table names that start with `auth`. At the moment I think I actually prefer this new ordering, but I wanted to make sure you're aware of this ramification of my suggestion.

ajlai:
Hi @drocco007,
Thanks for the suggestion! I like the quotes being ignored, but am not a fan of the underscore-separated case since `authorization` ends up in the middle of a bunch of `auth_X` suggestions. I'm currently leaning towards still keeping all chars in the lexical priority for consistency.

FWIW, sublime text with a blank `goto anything` does do their ordering this way:
image
 
I'll defer to @amjith though for what we want the expected behavior to be :smile:

drocco007:
> FWIW, sublime text with a blank goto anything does do their ordering this way:

Sure, but keep in mind that the quotes in the name of a file are significant in a way that the quotes around a PG table name are not: for the `"user"` file in the Sublime Text example, the quote characters are part of the name, whereas in the PG examples the table's name is actually `user`; the quotation marks are simply an artifact of PostgreSQL's handling of the intersection between table naming rules and its reserved keywords.

I'm not too concerned about the underscores, but I do think we need to keep the existing quote handling, which was in place long before the lexical ordering change that this PR is updating. Fortunately the spelling to do that is straightforward:

    lexical_priority = tuple(-ord(c) for c in self.unescape_name(item)) + (1,)

amjith:
I haven't had a chance to review this yet. But I'll get some time tonight. 

Sorry about the delay.

amjith:
Hi @ajlai, Thanks for taking the time to dig into this and thanks @drocco007 for reviewing the change and catching the quoted names bubbling to the top. 

My preference is to not have the quoted names bubble up, but I'm also not happy about ignoring the underscores in names. Having `authorization` show up in the middle of `auth_*` names looks odd. My preference would be to strip the quotes when determining the priority. Fortunately it is very easy to do. We already have a function to do exactly that `unescape_name()` https://github.com/dbcli/pgcli/blob/master/pgcli/pgcompleter.py#L63. 

@ajlai would you mind adding that behavior to your PR? If you're busy, let me know and I'll be happy to merge this and take on that task myself.

ajlai:
Thank you @amjith and @drocco007 for reviewing!

I'll add this to the PR later today.

amjith:
@ajlai When you get a chance to add a commit can you leave a comment in this PR? Github doesn't automatically notify when a new commit is added to a PR.

ajlai:
@amjith, names are now unescaped. PTAL.

amjith:
Yup. It works as expected. Thanks for fixing the bug. :ice_cream:

Issue #478, opened 2016-03-22
Auto-completion breaks for table names with ALL CAPS


amjith:
Here's an example: 

Broken for a table named "INSERT".

![quoted-completion](https://cloud.githubusercontent.com/assets/49260/13954269/4ec5bda4-effd-11e5-84e9-a9b6c260b7eb.gif)

NOT broken for table named "select".

![quoted-completion-not-broken](https://cloud.githubusercontent.com/assets/49260/13954354/9f550f68-effd-11e5-8818-7a80a277ba33.gif)

amjith:
This is now available in v1.0. Please upgrade. 

```
pip install -U pgcli
```

PR #479, opened 2016-03-23
Fix auto-completion breaking for table names with caps

ajlai:
Fix for #478

image


amjith:
Neat! Thanks for the quick turn around. :rocket:

Issue #467, opened 2016-02-24
\i <filename> + tab fails to autocomplete filenames


mikebissell:
When I import a SQL file using the \i command, I expect the tab key to autocomplete filenames for me, however pgcli doesn't know how to do this (PostgreSQL 9.3.3, Python 2.7.10, OS X 10.10.5).

Version: 0.20.1
Chat: https://gitter.im/dbcli/pgcli
Mail: https://groups.google.com/forum/#!forum/pgcli
Home: http://pgcli.com
pgdbdev> \i foo

I hit tab after typing the prefix "foo", so pgcli should have autocompleted "foobar.sql", since the file foobar.sql exists in the working directory. Instead, nothing happened -- no autocompletion. Absolute and relative paths to files outside the working directory should work, too, but they don't. For reference, this autocompletion feature works well in the old psql CLI.

amjith:
Yes. We have this planned but haven't gotten around to it.

ajlai:
@amjith I'd be interested in taking a stab at this one if someone isn't already doing it.

amjith:
Be my guest. :smile:

amjith:
This is now available in v1.0. Please upgrade. 

```
pip install -U pgcli
```

PR #480, opened 2016-03-24
Add \i path completion

ajlai:
Initial prototype of #467 

@amjith feel free to check it out. Looks like this for me:
image

If you select a directory it will start prompting autocomplete for the next characters you input or for everything if you just press tab.

image

Currently also handles `..` and `~`. What do you think?

### DIFF
--- a/pgcli/packages/sqlcompletion.py
+++ b/pgcli/packages/sqlcompletion.py
@@ -35,6 +35,8 @@ NamedQuery = namedtuple('NamedQuery', [])
 Datatype = namedtuple('Datatype', ['schema'])
 Alias = namedtuple('Alias', ['aliases'])
 
+Path = namedtuple('Path', [])
+
 
 def suggest_type(full_text, text_before_cursor):
     """Takes the full_text that is typed so far and also the text before the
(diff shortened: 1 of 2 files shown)


amjith:
@ajlai Thanks for the PR. It works great in my machine.

I was hoping to use the builtin PathCompleter from prompt_toolkit: https://github.com/jonathanslenders/python-prompt-toolkit/blob/master/prompt_toolkit/contrib/completers/filesystem.py

Here's an example of how it is used in ptpython: https://github.com/jonathanslenders/ptpython/blob/master/ptpython/completer.py

The advantage in reusing prompt_toolkit's path completer is we get the benefits of bug fixes in the future for free and it is well tested (in all the platforms we support). 

What do you think?

ajlai:
Hey @amjith,
Thanks for the suggestion. I'll take a look at this later tonight.

amjith:
Hi @ajlai, did you get a chance to look into the PathCompleter?

ajlai:
Hi @amjith,
I did take a look at PathCompleter and agree that we should use that instead. I'll have some more time to work on the PR tomorrow.

ajlai:
@amjith I switched it over to PathCompleter. Good suggestion, it looks a lot simpler :smile:.

Please give it a try, it behaves a bit differently since it's now just using the bare PathCompleter completions (need to manually type `/` to enter dirs, no display_meta).

amjith:
@ajlai Thank you for switching over. I've tested it and it looks great. :+1: 

We can add the meta information at a later time. 

@darikg Would you mind trying this in Windows to see if the path completer will work with dos path names?

darikg:
This is great! It works fine in Windows, although it does seem to break when directory names contain spaces. Probably a bug to report upstream to prompt-toolkit.

amjith:
Thanks for testing it @darikg.

PR #483, opened 2016-04-05
Add support for \o command to redirect query output to a file

Gollum999:


amjith:
@Gollum999 Thank you for the PR. Sorry it took me a while to get back to you. 

I tested the PR locally and it does the basic stuff right. 

I'd like to suggest a small improvement, can you log the query to the file in addition to the output? 

Otherwise it is hard to decipher the results from just the output.

amjith:
Actually I see that you're merely matching the behavior of psql. Can you tell me how you use `\o` in your regular workflow? 

I'm wondering if there is an advantage in preserving the psql behavior.

Gollum999:
@amjith I actually implemented this for a friend since he uses `\o` much more than I do.  He said that he typically uses it to diff different sets of output, e.g. a table before/after a modification, similar tables between development and release databases, etc..  In which case I don't see any harm in adding the query to the output file.

I also thought about giving some sort of feedback after using `\o`, e.g. "Writing to file {filename}" or "File output disabled", but I left it out to try to stick to psql's original functionality.  Is that something you would like added?

amjith:
That's a great idea. I'm all in favor of providing sensible user feedback. 👍 

I think having the queries in the output file will be valuable. Thanks for tackling this.

amjith:
It works great! Well done. 

There is a one corner case that is not handled. If the output has unicode characters then writing to the file fails. This is because you're opening the file without specifying an encoding. This can be easily fixed by adding an encoding as follows:

```
with open(self.output_file, 'a', encoding='utf-8') as f:
```

Once you make that change it is ready to be merged.

amjith:
I'm going to merge the PR and make that change myself. Thank you for taking the time to contribute. 🍨

Gollum999:
Thanks for finishing that up.  I've been busy the past few weeks and haven't had the time.

Glad to help!

PR #487, opened 2016-04-24
Add encoding to the output file.

amjith:
Reviewer: @j-bennet 

This PR changes adds the encoding to the file that is opened to write the output. 

This is a follow-up to the PR #483


j-bennet:
Looks good.

PR #488, opened 2016-04-27
Unquote URI parts prior to using in connection

asfaltboy:
As you may know, [RFC-3986][1] allows passing "disallowed" characters as long as these are percent encoded.
Since PGCLI deals with the url parsing, we should expect this to happen, and use the values after they have been decoded.

(includes a simple test for the parsing and decoding)

[1]: https://tools.ietf.org/html/rfc3986#section-2.1

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -54,9 +54,9 @@ from .__init__ import __version__
 click.disable_unicode_literals_warning = True
 
 try:
-    from urlparse import urlparse
+    from urlparse import urlparse, unquote
 except ImportError:
-    from urllib.parse import urlparse
+    from urllib.parse import urlparse, unquote
 
 from getpass import getuser
 from psycopg2 import OperationalError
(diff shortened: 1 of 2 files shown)


pgcli/main.py
```diff
@@ -238,8 +238,10 @@ def connect_dsn(self, dsn):
     def connect_uri(self, uri):
         uri = urlparse(uri)
         database = uri.path[1:]  # ignore the leading fwd slash
-        self.connect(database, uri.hostname, uri.username,
-                     uri.port, uri.password)
+        arguments = [database, uri.hostname, uri.username,
+                     uri.port, uri.password]
+        # unquote each URI part (they may be percent encoded)
+        self.connect(*list(map(lambda p: unquote(p) if p else p, arguments)))
```
amjith: If we run unquote on the raw uri before we call urlparse(), it should make the code simpler. Would you mind running the unquote on https://github.com/dbcli/pgcli/pull/488/files#diff-38f3aeda57742756afa677ca09dcea43R239

amjith:
Thank you for the PR and the accompanying test. I appreciate the time you took to create the PR. 

I've left a comment to simplify the code a little bit. Can you please revise the PR and leave a comment when you've done so?

pgcli/main.py
asfaltboy: Sorry for slow response. This cannot be done if we use `urlparse`, in fact this results in the error that has prompted me to submit this PR:

```
pgcli/main.py:241: in connect_uri
    uri = urlparse(unquoted)

../../../.pyenv/versions/2.7.11/lib/python2.7/urlparse.py:143: in urlparse
    tuple = urlsplit(url, scheme, allow_fragments)

>               raise ValueError("Invalid IPv6 URL")
E               ValueError: Invalid IPv6 URL
```

This happens if we unquote the URI and then try to parse the strings, which now includes invalid characters according to RFC  3986.

The error message from urllib is slightly misleading, hence this PR :)

amjith:
Good to know. Thank you for the follow through. 👍

PR #491, opened 2016-05-05
Improved config loading

jmunson:
I noticed that when I ran pgcli it was leaving a '~' dir in my current working dir. This is because ensure_dir_exists was not calling expanduser()

As long as I was fixing that I noticed you had hardcoded ~/.config, so I went ahead and made it check $XDG_CONFIG_HOME

You might want to change the Windows path to also use os.path.join for consistency, but since I don't have python on windows I did not want to risk breaking it without being able to check that it still works.

### DIFF
--- a/pgcli/config.py
+++ b/pgcli/config.py
@@ -9,7 +9,7 @@ def config_location():
     if platform.system() == 'Windows':
         return os.getenv('USERPROFILE') + '\\AppData\\Local\\dbcli\\pgcli\\'
     else:
-        return expanduser('~/.config/pgcli/')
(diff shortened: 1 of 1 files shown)


amjith:
Thanks for adding the `expanduser` fix to the `ensure_dir_exists()` function. That is a valid bug. 

I'm going to refrain from using the $XDG_CONFIG_HOME because we want to be able to tell the user where to look for the config file and log files when we try to address support issues. Having them look into $XDG_CONFIG_HOME/pgcli/* or ~/.config/pgcli/* is just cumbersome and adds another layer of confusion. 

So I'm going to cherry pick one of the commits from this PR but not the other. I hope that makes sense.

amjith:
Added to master.

PR #492, opened 2016-05-06
Completion to column list when *  is entered

koljonen:


darikg:
I think this is a fantastic idea. Usually whenever I write `SELECT * FROM foo` it's because I'm too lazy to type out all the columns from `foo`, not because I want a query that changes whenever the columns of `foo` change. 

The major issues I think are 

1.  How does this interact with table aliases? Should `SELECT f.* FROM foo f` suggest `f.bar, f.baz, f.qux` ?

2. How do we indicate to the user what is being suggested? Right now the meta-text is simply 'columns'. This would probably be the most 'magical' suggestion that pgcli makes -- is there any way to suggest that something special is being suggested here?

amjith:
I'm not sure I understand what this PR is doing. Can one of you explain?

koljonen:
amjith: What it does is that e.g. when the user types "insert into sometable(*<tab>", the result is "insert into sometable(cola, colb, colc, ...", and likewise for other places where pgcli suggests columns.

darikg:
1. I'll look into how we can copy the qualifier. I definitely think we should.
2. I don't think this is that big a problem, because there's only one option in the list when * is entered, so it's automatically selected, and in all likelihood it's what the user wanted.

amjith:
@darikg Would you might taking stewardship of this PR? I have a newborn at home and slammed at the moment.

darikg:
@koljonen I'm away from my computer for the next couple of days but when I get back early next week I'll have a closer look

darikg:
Some thoughts:

  - This seems to unnecessarily complicate `find_matches`. I think you could leave that function as is and move the * special case to `get_column_matches` instead.

- Right now you're pulling column names out of the completions, which seems indirect. You can look at the code in `populate_scoped_cols` for how to pull columns out of the metadata dicts. Or if move everything to `get_column_matches` I think you might be able to use the `scope_cols` variable directly. 

- I *think* it shouldn't be necessary to add the extra `parent` metadata column to the `Column` type. Can't you get the table alias out of the `tables` field in `Column`?

- Need some tests. Here's one you can add to `test_smart_completion_public_schema_only.py`:

```
def test_wildcard_column_expansion(completer, complete_event):
    sql = 'SELECT * FROM users'
    pos = len('SELECT *')

    completions = completer.get_completions(
        Document(text=sql, cursor_position=pos), complete_event)

    col_list = 'id, email, first_name, last_name'
    expected = Completion(text=col_list, start_position=-1,
                          display='*', display_meta='columns')

    assert expected in completions
```

koljonen:
Good feedback. I've made those changes.

pgcli/pgcompleter.py
```diff
@@ -308,15 +307,33 @@ def get_column_matches(self, suggestion, word_before_cursor):
         tables = suggestion.tables
         _logger.debug("Completion column scope: %r", tables)
         scoped_cols = self.populate_scoped_cols(tables)
-
+        colit = scoped_cols.items()
+        flat_cols = []
```
darikg: just personal preference but I find something like `flat_cols = itertools.chain(*scoped_cols.items())` a bit clearer in intent

pgcli/pgcompleter.py
```diff
@@ -308,15 +307,33 @@ def get_column_matches(self, suggestion, word_before_cursor):
...
                                    if count > 1 and col != '*']
-
-        return self.find_matches(word_before_cursor, scoped_cols, meta='column')
+        lastword = last_word(word_before_cursor, include='most_punctuations')
+        if lastword == '*':
+            if (lastword != word_before_cursor and len(tables) == 1
+              and word_before_cursor[-len(lastword) - 1] == '.'):
+                # User typed x.*; replicate "x." for all columns
+                sep = ', ' + self.escape_name(tables[0].ref) + '.'
+                collist = (sep).join([c for c in flat_cols if c != '*'])
```
darikg: This must be incorrect because for columns `foo, bar, baz` you get `foo, x.bar, x.baz`, i.e. it forgets the leading qualifier. I think you want
```
qual = self.escape_name(tables[0].ref) + '.'
collist = ', '.join(qual + c for c in flat_cols if c ~= '*')
```
Note also you don't need parens around the string argument to `join`, e.g. `sep.join(` not `(sep).join(`. You also don't need square brackets inside the join call, e.g. `sep.join(c for c in cols)` not `sep.join([c for c in cols])`

tests/test_smart_completion_public_schema_only.py
```diff
@@ -553,3 +553,73 @@ def test_columns_before_keywords(completer, complete_event):
...
+    assert expected == completions
+
+def test_wildcard_column_expansion_with_two_tables(completer, complete_event):
+    sql = 'SELECT * FROM "select" JOIN users u ON true'
+    pos = len('SELECT *')
+
+    completions = completer.get_completions(
+        Document(text=sql, cursor_position=pos), complete_event)
+
+    # The order of the tables is indeterministic, so allwo both possibilities
```
darikg: So I'm pretty sure the reason the table order is indeterministic here is that you now have `populate_scoped_cols` returning a dict, and `get_column_matches` is now iterating over the items of that dict, and of course the order of keys in a dict is non-deterministic. It'd be nice to fix this issue there, instead of having the more flexible test here. At least for the user it'd be nice to have columns suggested in a predictable manner. I think you could have  `populate_scoped_cols` return a ordered dict instead, if such a thing is possible with defaultdict.

pgcli/pgcompleter.py
```diff
@@ -308,15 +307,33 @@ def get_column_matches(self, suggestion, word_before_cursor):
...
+            elif len(scoped_cols) > 1:
+                # Multiple tables; qualify all columns
+                collist = (', ').join([t.ref + '.' + c for t, cs in colit
+                    for c in cs if c != '*'])
+            else:
+                # Plain columns
+                collist = (', ').join([c for c in flat_cols if c != '*'])
+            return [Match(completion=Completion(collist, -1,
```
darikg: Extra blank line would be nice here

pgcli/pgcompleter.py
```diff
@@ -308,15 +307,33 @@ def get_column_matches(self, suggestion, word_before_cursor):
...
+                display_meta='columns', display='*'), priority=(1,1,1))]
+        return self.find_matches(word_before_cursor, flat_cols, meta='column')
```
darikg: Here too

pgcli/pgcompleter.py
```diff
@@ -436,76 +453,33 @@ def get_namedquery_matches(self, _, word_before_cursor):
...
         :param scoped_tbls: list of TableReference namedtuples
-        :return: list of column names
+        :return: dict {TableReference:[list of column names]}
         """
 
-        columns = []
+        columns = defaultdict(lambda: [])
         meta = self.dbmetadata
-
+        def addcols(schema, rel, alias, reltype, cols):
```
darikg: Blank lines before and after function definitions, per pep8

darikg:
I really appreciate you expanding the tests to cover more complicated situations. I left a number of suggestions inline.

pgcli/pgcompleter.py
koljonen: The leading qualifier isn't part of the text that gets replaced, so the qualifier needs to be skipped for the first column in that case.

tests/test_smart_completion_multiple_schemata.py
```diff
@@ -322,4 +322,74 @@ def test_suggest_columns_from_aliased_set_returning_function(completer, complete
...
+    assert expected == completions
+
+def test_wildcard_column_expansion_with_two_tables(completer, complete_event):
+    sql = 'SELECT * FROM public."select" JOIN custom.users ON true'
+    pos = len('SELECT *')
+
+    completions = completer.get_completions(
+        Document(text=sql, cursor_position=pos), complete_event)
+
+    # The order of the tables is indeterministic, so allwo both possibilities
```
darikg: Outdated comment

tests/test_smart_completion_public_schema_only.py
```diff
@@ -553,3 +553,71 @@ def test_columns_before_keywords(completer, complete_event):
...
```
darikg: Outdated comment

darikg:
There's two comments left behind in the tests about indeterministic order, but other than that I think this is good to go :fireworks:

PR #496, opened 2016-05-13
Support for join-condition suggestions after ON

koljonen:
The user types 'SELECT * FROM parenttable p JOIN childtable c on '. We then suggest ['c.parenttableid = p.parenttableid', 'c.createdby = p.createdby', ...] as completions.
The suggestions are FK matches first, then name matches for int columns, then name matches for other columns.

Some things that have been changed in order to make this work:
    meta[reltype]][relname] goes from being a list of columns to being a list of ColumnMetadata namedtuples, which, apart from the column name, also contains its data type and a list of FKs from the column (i.e. where the column is a child). This entails modification of a number of functions dealing with columns.
    get_matches can now be fed with priorities from outside, enabling more intelligent sorting.
    TableReference gets a property, ref, that returns COALESCE(Alias, TableName).
    populate_scoped_cols now returns a dict {TableReference:{colname:ColumnMetadata}} instead of just a list [colname].


darikg:
Looks like a bunch of tests are failing here because of `dict.itervalues()`

darikg:
This is a pretty big change, so it will take me some time to get through. I like the direction it's going in.

koljonen:
I've squashed and rebased onto my drop_asterisk_column_suggestion branch.

darikg:
Still slowly going through this. Do you feel like pulling out the proposed changes to `find_matches` and submitting that as a standalone PR? The more eyes we get on that part the better, because it plays such an important role in everything else.

darikg:
I had an idea today! What if instead of tweaking `find_matches` to allowing injecting higher priority suggestions, we got fuzzy matching to play nice with special characters like `=`? I'm imagining the user would start typing `select * from foo join bar on `, and then even if the join condition suggestion was really low on the list, they could type one extra character `select * from foo join bar on =`, and then the join conditions would rocket to the top of the list?

koljonen:
I think that at least in the case where the user hasn't typed anything after 'ON ', the gain from having the join conditions on top is rather large Being able to just type 'select * from foo join bar on <tab> join baz on <tab>' and being able to trust that you get the right join conditions automatically without having to think about it is a pretty great user experience.

pgcli/packages/function_metadata.py
```diff
@@ -1,7 +1,7 @@
 import re
 import sqlparse
 from sqlparse.tokens import Whitespace, Comment, Keyword, Name, Punctuation
-
+from ..pgcompleter import ColumnMetadata
```
darikg: We have a rule of thumb that stuff in the /packages module should be independent of pgcli. So it'd be better to move the definition of ColumnMetadata here, and have pgcompleter import it, instead of the other way around

pgcli/pgcompleter.py
```diff
@@ -34,8 +34,9 @@
 
 
 Match = namedtuple('Match', ['completion', 'priority'])
-
-
+ColumnMetadata = namedtuple('ColumnMetadata', ['name', 'datatype', 'foreignkeys'])
+ForeignKey = namedtuple('ForeignKey', ['parentschema', 'parenttable',
```
darikg: Might as well move ForeignKey definition into /packages alongside ColumnMetadata

pgcli/pgcompleter.py
```diff
@@ -341,6 +369,55 @@ def get_column_matches(self, suggestion, word_before_cursor):
...
+        refprio = dict((tbl.ref, num) for num, tbl
+            in enumerate(suggestion.tables))
+        # Map (schema, tablename) to tables and ref to columns
+        tbldict = defaultdict(list)
+        for t, cs in scoped_cols.items():
+            tbldict[(t.schema, t.name)].append(t)
+        refcols = dict((t.ref, cs) for t, cs in scoped_cols.items())
+        # For each fk from the left table, generate a join condition if
+        # the other table is also in the scope
+        conds = [(make_cond(lefttable.ref, rtbl.ref, lcol.name, rcol[2]),
```
darikg: This list comprehension is a beast! Should probably switch to a more traditional nested for-loop structure

pgcli/pgcompleter.py
```diff
@@ -341,6 +369,55 @@ def get_column_matches(self, suggestion, word_before_cursor):
...
+            if rtbl and rtbl.ref != lefttable.ref]
+
+        # For name matching, use a {(colname, coltype): TableReference} dict
+        col_table = defaultdict(lambda: [])
+        for tbl, col in ((t, c) for t, cs in scoped_cols.items() for c in cs):
+            col_table[(col.name, col.datatype)].append(tbl)
+        # Find all name-match join conditions
+        found = set(cnd[0] for cnd in conds)
+        conds.extend((cond, 'name join',
+            1000 if c.datatype[:3] == 'int' else 0 + refprio[rtbl.ref])
```
darikg: This one too

tests/test_parseutils.py
```diff
@@ -1,14 +1,14 @@
...
+    find_prev_keyword, is_open_quote)
 
 def test_empty_string():
     tables = extract_tables('')
     assert tables == ()
 
 def test_simple_select_single_table():
     tables = extract_tables('select * from abc')
-    assert tables == ((None, 'abc', None, False),)
+    assert tables == (TableReference(None, 'abc', None, False),)
```
darikg: These changes shouldn't strictly be necessary, since namedtuples can equal bare tuples. Do you find being more explicit here helpful? If so, I'm totally fine with these changes, I'm just curious.

darikg:
I've finally finished reading everything. I left some suggestions inline. The only other thing that's missing would be to add a test to test_pgexecute that creates two tables with an FK relationship, and asserts that the `executor.foreignkeys()` returns the correct ForeignKey

pgcli/pgexecute.py
```diff
@@ -6,10 +6,11 @@
 import psycopg2.extensions as ext
 import sqlparse
 import pgspecial as special
+from collections import namedtuple
 from .packages.function_metadata import FunctionMetadata
 from .encodingutils import unicode2utf8, PY2
 
-
+ForeignKey = namedtuple('ForeignKey', ['parentschema', 'childschema', 'parenttable', 'childtable', 'parentcolumns', 'childcolumns'])
```
darikg: Is ForeignKey defined twice, once here and once in pgcompleter?

tests/test_parseutils.py
koljonen: That was just me being confused by the output of the tests, so I've reverted it now.

pgcli/pgexecute.py
koljonen: They're slightly different in that FKs from the DB can contains multiple columns, whereas the FKs stored in ColumnMetadata are a relationship between only two columns (for simplicity).

tests/test_smart_completion_public_schema_only.py
```diff
@@ -32,7 +32,7 @@ def completer():
     tables, columns = [], []
     for table, cols in metadata['tables'].items():
         tables.append(('public', table))
-        columns.extend([('public', table, col) for col in cols])
+        columns.extend([('public', table, col, 'text') for col in cols])
```
darikg: Maybe add a comment here to point out that we're hardcoding the dummy column types to 'text'?

tests/test_sqlcompletion.py
```diff
@@ -1,13 +1,14 @@
...
+    Function, Datatype, Alias, JoinCondition)
+from pgcli.packages.parseutils import TableReference
 import pytest
 
 
 def test_select_suggests_cols_with_visible_table_scope():
     suggestions = suggest_type('SELECT  FROM tabl', 'SELECT ')
     assert set(suggestions) == set([
-        Column(tables=((None, 'tabl', None, False),)),
+        Column(tables=(TableReference(None, 'tabl', None, False),)),
```
darikg: Revert these TableReference changes too?

pgcli/pgcompleter.py
```diff
@@ -34,8 +35,6 @@
 
 
 Match = namedtuple('Match', ['completion', 'priority'])
-
-
 class PGCompleter(Completer):
```
darikg: Can you put back the whitespace before the class definition? It looks really weird without it

tests/test_smart_completion_public_schema_only.py
```diff
@@ -41,7 +41,7 @@ def completer():
     views, columns = [], []
     for view, cols in metadata['views'].items():
         views.append(('public', view))
-        columns.extend([('public', view, col) for col in cols])
+        columns.extend([('public', view, col, 'text') for col in cols])
```
darikg: ditto

darikg:
This is really shaping up. I just merged the find_matches PR, so you can rebase this one on top of master, and fix a couple nitpicks I had, and I think we're good to go

koljonen:
Fixed, squashed and rebased.

darikg:
Sorry, one more issue I just found: the `join_cond_pat` regex fails with new lines. Here's a test for `test_sqlcompletion` that's currently failing:

```
def test_on_suggests_join_multilines():
    sql = '''   SELECT *
                FROM foo
                    JOIN bar ON '''
    suggestions = suggest_type(sql, sql)
    expected = JoinCondition(
        tables=((None, 'foo', None, False),
                (None, 'bar', None, False)),
        parent=None)
    assert expected in suggestions
```

In general, `join_cond_pat` seems more complicated than it needs to be. Can you explain it a bit more?

koljonen:
Good catch.

darikg:
Ahh thanks `join_cond_pat` finally clicked for me. I have a trick [here](https://github.com/dbcli/pgcli/commit/c3e0edc592bed56a181bddd5cd0643aa53cfb2ab) that uses the `sqlparse` parsed statement to check the preceding token instead of a regex, which I think will work even with table names containing spaces (as long as they're escaped, of course.) Does that seem reasonable?

koljonen:
Yeah, using the parser is obviously better.

darikg:
:boom: Merged! Thanks for your patience & hard work. This is a killer feature!

PR #497, opened 2016-05-13
Improve filtering of functions for FROM clause

koljonen:
Change from filtering out all non-set-returning functions to only filteirng out windowing functions and aggregate functions.
Non-set-returning functions are legal in the FROM clause and can be quite useful.


darikg:
:+1: on the intent but it looks like you'll need to update a whole bunch of the tests in `test_smart_completion_public_schema_only.py` and `test_smart_completion_multiple_schemata.py`

darikg:
Thanks for fixing the tests!

PR #500, opened 2016-05-14
Make pgclirc_file argument regular (non-optional)

### LINKED ISSUE #499: PGCli initialization issue
On [PGCli](https://github.com/dbcli/pgcli/blob/master/pgcli/main.py#L109) initialization [`write_default_config()`](https://github.com/dbcli/pgcli/blob/master/pgcli/config.py#L33) is called with non-optional `destination` argument, which indicates that in PGCli constructor `pgclirc_file` argument should not be optional, as it is now set.
That argument is later again called by other functions in constructor.

klonuo:
Closes #499

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -95,8 +95,8 @@ class PGCli(object):
         # different than less or is already parameterized with their own arguments
         os.environ['LESS'] = '-SRXF'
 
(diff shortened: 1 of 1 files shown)


klonuo:
Closed by https://github.com/dbcli/pgcli/pull/506

PR #501, opened 2016-05-14
Fix bug where "USING(<cursor>" wasn't detected properly + add test case

koljonen:
prev_tok is something like '[table name] [table alias] USING' where we expect it to be 'USING', so I just added a quick fix where we check the last word in prev_token instead of the whole string.

### DIFF
--- a/pgcli/packages/sqlcompletion.py
+++ b/pgcli/packages/sqlcompletion.py
@@ -247,7 +247,9 @@ def suggest_based_on_last_token(token, text_before_cursor, full_text, identifier
 
         # Get the token before the parens
         prev_tok = p.token_prev(len(p.tokens) - 1)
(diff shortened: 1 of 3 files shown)


koljonen:
I've also added a PR at [https://github.com/andialbrecht/sqlparse/pull/236](https://github.com/andialbrecht/sqlparse/pull/236) to fix the underlying sqlparse bug.

darikg:
Good catch. We actually already have tests for the USING thing, it just didn't test the case where there wasn't any whitespace before the parentheses. Instead of adding an entirely new test, can you just tweak the ones we already have? [Here's how I would do it.](https://github.com/dbcli/pgcli/commit/8a29e7084cadfca7f5c45dfd229d2c47616dd1ae)

koljonen:
Good idea. Done.

PR #505, opened 2016-05-16
Stop suggesting * as a column

koljonen:
I don't know about you guys, but I don't think suggesting * is of use to anyone, and it's in the way for anyone who just wants a column.


darikg:
I'm cool with this. +6 -23 lines is especially nice.

amjith:
👍  Looks like this PR needs a rebase against master before merge.

koljonen:
Rebased.

PR #507, opened 2016-05-19
Add priority arguments to find_matches

koljonen:
@darikg suggested I make this a separate PR, so here it is. See #496 for more background.

**Original commit message:**
type_priority is an int that says how to sort these matches compared matches of other types.
type_collection is a an iterable that overrides the internal sorting of the matches.

The use case that motivated this change was to be able to
    1) sort join conditions ("t.someid = t2.someid") before table aliases ("t")
    2) sort FK-based join conditions before name-based join conditions and integer name-based join conditions before other name-based join conditions.


darikg:
Thanks! I've had some time to think about it and it's totally reasonable, and noone else objected.

PR #508, opened 2016-05-19
Fix problem where non-lower-case table names would not be recognized

koljonen:
This fixes the issue where pgcli interprets **SELECT * FROM Foo** as referring to the table **Foo**, whereas it actually refers to table **foo**.


darikg:
I only have trivial suggestions here: it'd be nice to have consistency in the camel_casing, so `schema_is_quoted` instead of `schemaisquoted` etc. And maybe something more descriptive like `standardize_name_case` instead of `getnames`. But all the changes/additions to the tests seem great.

koljonen:
I renamed the function parse_identifier, as it does do some more than the casing thing.

darikg:
Even better!

PR #510, opened 2016-05-25
Fix another problem where table names would not be recognized when nott in lower case

koljonen:
My previous fix didn't cover the case 'INSERT INTO Foo(\<cursor\>)', which is now covered.


darikg:
Right on

Issue #444, opened 2016-01-21
Support environment variables in .pgclirc


kristjan:
`.psqlrc` provides a few variables like `:HOST`, `:PORT` and `:DBNAME` so you can do useful things like keep database-specific history files. Similar functionality in `.pgclirc` would be wonderful.

amjith:
@kristjan Thanks for filing the issue. Can you give me a specific example? I haven't used this functionality in psqlrc so I'm not familiar with this feature.

kristjan:
Absolutely, here's what the history file configuration looks like:

    \set HISTFILE ~/.psql_history_ :DBNAME

Given that, DBs `foo` and `bar` get independent query histories written to `~/.psql_history_foo` and `~/.psql_history_bar`, which looks like this:

![db_history](https://cloud.githubusercontent.com/assets/34301/12475360/5330bafc-bfd7-11e5-9e8f-7a5218bb2618.gif)

fernandomora:
psql can set environment vars with \setenv command:
http://www.postgresql.org/docs/9.5/static/app-psql.html
> \setenv name [ value ]
> Sets the environment variable name to value, or if the value is not supplied, unsets the environment variable. Example:
> 
> test=# \setenv PAGER less
> test=# \setenv LESS -imx4F

Also you can include setting of environment vars in .psqlrc. e.g:
```
\setenv PAGER less
\setenv PAGER less
```

HISTFILE, DBNAME, PORT, etc... are not environment variables, they are internal psql variable and can be set with \set and shown with \echo :VARNAME. e.g:
```
test=# \echo :DBNAME
test
test=# \echo :ENCODING
test
```

> \set [ name [ value [ ... ] ] ]
> Sets the psql variable name to value, or if more than one value is given, to the concatenation of all of them. If only one argument is given, the variable is set with an empty value. To unset a variable, use the \unset command.

Also in .psqlrc. e.g:
```
\set VERBOSITY verbose
\set COMP_KEYWORD_CASE upper
```
I am not sure if it is easy to read/write this psql internal vars, but I'm sure it is at least for env vars. 

I purpose to allow same functionality: usage of \set (probably also \unset) and \setenv, or at least \setenv in pgcli both in command line and in .pgclirc.

Note: If this is implemented I think it would be good to make a breaking change to #452 before next release to configure default PAGER in .pgclirc using "\setenv" instead of "pager":
```
\setenv LESS '-iMSx4FXRe'
\setenv PAGER 'less'
```

[...]

PR #512, opened 2016-05-26
Custom casing of table/column/function/view names

koljonen:
This is for those of us need/want to use unescaped names in CamelCase or whatever. This means we can use pgcli to edit our functions without having to correct the casing all the time!

**Original commit message:**
This adds support for defining the casing to be used in completions (for unquoted names). The casing is determined by the casing file, which consists of a \n-separated list of names using the preferred casing. This file is initially populated based on usage in SQL and PLPGSQL database functions but can then be modified by the user.

### DIFF
--- a/pgcli/config.py
+++ b/pgcli/config.py
@@ -54,3 +54,10 @@ def get_config(pgclirc_file=None):
     write_default_config(default_config, pgclirc_file)
 
     return load_config(pgclirc_file, default_config)
+
+
+def get_casing_file(config):
+    casing_file = config['main']['casing_file']
+    if casing_file == 'default':
+        casing_file = config_location() + 'casing'
+    return casing_file
--- a/pgcli/pgcompleter.py
+++ b/pgcli/pgcompleter.py
@@ -43,8 +43,7 @@ class PGCompleter(Completer):
         self.prioritizer = PrevalenceCounter()
         settings = settings or {}
         self.casing_file = settings.get('casing_file')
-        self.generate_casing_file = (
-            settings.get('generate_casing_file') == 'True')
+        self.generate_casing_file = settings.get('generate_casing_file')
         self.asterisk_column_order = settings.get(
             'asterisk_column_order', 'table_order')
 
--- a/pgcli/pgcompleter.py
+++ b/pgcli/pgcompleter.py
@@ -41,8 +41,13 @@ class PGCompleter(Completer):
         self.smart_completion = smart_completion
         self.pgspecial = pgspecial
         self.prioritizer = PrevalenceCounter()
-        self.asterisk_column_order = (settings or {}).get(
+        settings = settings or {}
+        self.casing_file = settings.get('casing_file')
+        self.generate_casing_file = (
+            settings.get('generate_casing_file') == 'True')
+        self.asterisk_column_order = settings.get(
             'asterisk_column_order', 'table_order')
+
         self.reserved_words = set()
         for x in self.keywords:
             self.reserved_words.update(x.split())
@@ -52,6 +57,7 @@ class PGCompleter(Completer):
         self.dbmetadata = {'tables': {}, 'views': {}, 'functions': {},
                            'datatypes': {}}
         self.search_path = []
+        self.casing = {}
 
         self.all_completions = set(self.keywords + self.functions)
 
@@ -95,6 +101,14 @@ class PGCompleter(Completer):
                 metadata[schema] = {}
 
         self.all_completions.update(schemata)
+
+    def extend_casing(self, words):
+        """ extend casing data
+
+        :return:
+        """
+        # casing should be a dict {lowercasename:PreferredCasingName}
+        self.casing = dict((word.lower(), word) for word in words)
 
     def extend_relations(self, data, kind):
         """ extend metadata for tables or views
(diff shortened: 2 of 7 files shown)


pgcli/config.py
```diff
@@ -42,3 +42,21 @@ def write_default_config(source, destination, overwrite=False):
...
+    pgclirc_file = pgclirc_file or '%sconfig' % config_location()
+
+    default_config = os.path.join(package_root, 'pgclirc')
+    write_default_config(default_config, pgclirc_file)
+
+    return load_config(pgclirc_file, default_config)
+
+def get_casing_file(config):
+    casing_file = config['main']['casing_file']
+    if casing_file == 'default':
```
darikg: this logic is duplicated in the pgcli casing_file() @property

darikg:
Super cool.

If there's no case file in the default location, why bother writing one to disk that contains all possible names? That means the user's default case file depends on what database they connect to first. IMO it would make more sense to leave the default casefile empty/nonexistent, and the user would write into it only the few non-default capitalizations they wanted. (I.e., the casefile would contain only overrides.)

amjith:
I realize I'm out of my depth on this one.

I see that it creates a casing file in the config folder with a subset of table names in that DB. What's supposed to go in that file? How is that subset determined? I'm having a hard time deciphering the query for casing in the pgexecute file. 

You mentioned something about using unescaped CamelCase, but won't that result in postgres doing the conversion to lower case in the backend causing a failure? 

Seeing some examples might help with understanding.

koljonen:
@darikg The case I was trying to make it work out-of-the-box for was the one where the user uses one database and all code is written using CamelCase, with all names unescaped. The idea is the user starts pgcli (or pgclisublime) for the first time, and it automatically gets the casing of all names right, without any user intervention. Defaulting to an empty file wouldn't be very useful in that scenario, as the user would then need to somehow enter every name into the casing file.

@amjith The query takes all table/function/view/schema names in the database, joins that (case-insensitively) with all words used in the definitions of database functions, determines the most popular casing for each word/name, and then that gets put in the file.
E.g. if we have the function '...UPDATE Bar.Foo SET FooStatusID = _FooStatusID, ModificationDate = now() WHERE FooID = _FooID;...', the file will be populated thusly:
> Bar
> Foo
> FooID
> FooStatusID
> ModificationDate

Then, the next time we're writing a function, pgcli will suggest FooID instead of fooid.

darikg:
@koljonen  I was suggesting piping the output from `pgexecute.casing` directly to `completer.extend_casing`, and then taking values from the casing file (if it exists) and sending them to `completer.extend_casing`, overriding any default values. Does that make sense?

@amjith I merged a couple of PRs recently from @koljonen (#508, #510) that have made pgcli behave better with non-default casing. This PR is the flipside of that, getting more case-friendly suggestions.

There's also an open issue somewhere where somebody wanted lower case keywords, e.g. suggest `select` instead of `SELECT`, which this PR doesn't handle, but it would be a reasonable extension down the road.

koljonen:
@darikg Oh, now I think I understand what you're getting at. I should probably have put this in the commit message or something: The casing file is needed as a cache, because pgexecute.casing will take several seconds on a large database (I think it should scale linearly with the amount of function code). E.g. on my local 9.1 database of 2k functions with a total of 3.5M characters of SQL/PLPGSQL function code, it takes 10 seconds – which is a delay I can live with the first time I start pgcli, but definitely not every time completions get refreshed. This is actually the main reason there is a casing file at all.

darikg:
Well the completion refreshing happens in a separate thread, so it shouldn't be super visible to the user, although 10 seconds is pretty long.

I think wanting to get good default out-of-the-box behavior is great, but only if it's consistent. Currently this  will only work on the first database the user connects to, and even then it won't stay up to date if the database changes. I think it needs to either work by default in all cases, or not work by default at all, and require the user to explicitly fill out the casing file. 

Another alternative to a clever-but-slow database query to get default casing would be to pull it out of the history file. We're [already](https://github.com/dbcli/pgcli/blob/32029e27668dcdc6204bab8d8d64236962ecf973/pgcli/completion_refresher.py#L76) scanning the history to grab recently used identifiers so we can prioritize them as completions, that seems like a good place to also identify user capitalization preferences.

koljonen:
The problem with the 10 seconds is the first time the refreshing is run (i.e. just after starting pgcli), because you don't get any completions before it's done. 

I think the history file is too likely to be lacking in quality (because of ad-hoc queries where you don't care about capitalization) as well as quantity (because, in a large or new-to-you database, you've only used a small portion of all names) to be of much use here.

Anyway, I just added an option to enable/disable auto-generating the casing file, as I think that's the simplest solution right now.

Some ideas for the future: 
1. Split the file into one that's manual and one that's generated.
2. A config setting for scheduling regeneration of the file.
3. A backslash command for triggering regeneration of the file.
4. If we get around to doing #444, we could have separate files per database.

darikg:
I think you meant to push the 'Change % to os.path.join' to the other PR

koljonen:
Yeah, you're right, but we do need it here too.

darikg:
This needs to be rebased. I like the current default case, where the user has to explicitly opt in.

koljonen:
Squashed and rebased.

pgcli/pgcompleter.py
```diff
@@ -41,8 +41,13 @@ def __init__(self, smart_completion=True, pgspecial=None, settings=None):
         self.smart_completion = smart_completion
         self.pgspecial = pgspecial
         self.prioritizer = PrevalenceCounter()
-        self.asterisk_column_order = (settings or {}).get(
+        settings = settings or {}
+        self.casing_file = settings.get('casing_file')
+        self.generate_casing_file = (
+            settings.get('generate_casing_file') == 'True')
```
darikg: I think the conversion from the string 'True' to a boolean needs to happen as early as possible, when PGCli.__init__ pulls it out from the config option. The pgcompleter settings struct should expect a boolean value here

darikg:
Hooray!

PR #513, opened 2016-05-26
Order columns by in-table order when expanding *

koljonen:
Simply store them in that order when getting them from the DB, instead of sorted by name.

### DIFF
--- a/pgcli/config.py
+++ b/pgcli/config.py
@@ -42,3 +42,15 @@ def write_default_config(source, destination, overwrite=False):
 def upgrade_config(config, def_config):
     cfg = load_config(config, def_config)
     cfg.write()
+
+
+def get_config(pgclirc_file=None):
+    from pgcli import __file__ as package_root
+    package_root = os.path.dirname(package_root)
+
+    pgclirc_file = pgclirc_file or '%sconfig' % config_location()
+
+    default_config = os.path.join(package_root, 'pgclirc')
+    write_default_config(default_config, pgclirc_file)
+
+    return load_config(pgclirc_file, default_config)
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -103,7 +103,7 @@ class PGCli(object):
 
         from pgcli import __file__ as package_root
         package_root = os.path.dirname(package_root)
-        
+
         pgclirc_file = pgclirc_file or '%sconfig' % config_location()
 
         default_config = os.path.join(package_root, 'pgclirc')
@@ -136,7 +136,8 @@ class PGCli(object):
 
         # Initialize completer
         smart_completion = c['main'].as_bool('smart_completion')
-        completer = PGCompleter(smart_completion, pgspecial=self.pgspecial)
+        completer = PGCompleter(smart_completion, pgspecial=self.pgspecial,
+            config=self.config)
         self.completer = completer
         self._completer_lock = threading.Lock()
         self.register_special_commands()
@@ -576,8 +577,8 @@ class PGCli(object):
 
         callback = functools.partial(self._on_completions_refreshed,
                                      persist_priorities=persist_priorities)
-        self.completion_refresher.refresh(
-            self.pgexecute, self.pgspecial, callback, history=history)
+        self.completion_refresher.refresh(self.pgexecute, self.pgspecial,
+            callback, history=history, config=self.config)
         return [(None, None, None,
                 'Auto-completion refresh started in the background.')]
 
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -136,8 +136,10 @@ class PGCli(object):
 
         # Initialize completer
         smart_completion = c['main'].as_bool('smart_completion')
+        settings = {
+            'asterisk_column_order':c['main']['asterisk_column_order']}
         completer = PGCompleter(smart_completion, pgspecial=self.pgspecial,
-            config=self.config)
+            settings=settings)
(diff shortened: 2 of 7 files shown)


msztolcman:
As a user, i'm strongly against it. Sorted by name allows me for much more convenient and quickier finding what I need. Especially when most of tables are created by some ORM.

koljonen:
I think, quite simply, that what would be least surprising and most useful to most users is that 'SELECT * FROM Foo;' and  'SELECT \<expanded *\> FROM Foo;' produce the same result.

darikg:
I think I prefer table order but I can definitely see why some would prefer alphabetical. I guess we need a config option? We could also give two different completions and somehow get the prioritizer to figure out which the user prefers but that seems pretty gross

pgcli/pgclirc
```diff
@@ -31,6 +31,10 @@ history_file = default
 # and "DEBUG".
 log_level = INFO
 
+# Order of columns when expanding * to column list
+# Possible values: "table_order" and "alphabetic"
+column_order = table_order
```
darikg: Let's make this something a little bit more specific like `asterisk_column_order`

pgcli/pgexecute.py
```diff
@@ -354,9 +339,28 @@ def _columns(self, kinds=('r', 'v', 'm')):
...
         """
+        column_order = {'table_order': 'att.attnum', 'alphabetic': '3'}[
+            self.config['main']['column_order']]
+
+        columns_query = '''
+            SELECT  nsp.nspname schema_name,
+                    cls.relname table_name,
+                    att.attname column_name,
+                    typ.typname type_name
+            FROM    pg_catalog.pg_attribute att
```
darikg: what if we store attnum in the column metadata object, and then have the *-expansion code in `get_column_matches` sort by whatever's appropriate there? table-order vs alphabetical order might come up later in other completions so instead of fixing it at the level of `pgexecute.columns` we should fix it at the site of the suggestion

koljonen:
There have been so many changes now, so I squashed it all into one commit to make things simpler.
I've moved the sorting logic into **get_column_matches** (good idea, @darikg), but I think adding the column number to the metadata would needlessly complicate things, compared to just having the columns sorted in canonical order.
Note that this change means that functions too will now be affected by the setting. I guess that should be a good thing.

darikg:
> but I think adding the column number to the metadata would needlessly complicate things, compared to just having the columns sorted in canonical order.

Yeah the way you're doing it now is totally reasonable

Can we revert the changes to pgrefresher where it requires being passed the config object? It's not being used anymore and anyway it seems more tightly coupled than is necessary.

koljonen:
The config object is used when creating the new PGCompleter object in the refresher?

darikg:
Sorry, I didn't explain that very well. What I mean is to give pgcompleter its own `asterisk_column_order` property (and I guess pgrefresher would need one as well), instead of passing the config object as a whole. It seems maybe a little less easily extensible, but it means that pgcompleter is less tightly coupled to the pgcli config structure. For example, if you wanted to write a test that setting `asterisk_column_order = 'alphabetical'` behaves properly, you would have to create a mock config, which seems really weird to me.

koljonen:
I've changed it into a settings dict, but this does mean the setting won't be imported from the config file when running e.g. pgclisublime. Do you have any ideas on how to solve that?

pgcli/config.py
```diff
@@ -42,3 +42,15 @@ def write_default_config(source, destination, overwrite=False):
 def upgrade_config(config, def_config):
     cfg = load_config(config, def_config)
     cfg.write()
+
+
+def get_config(pgclirc_file=None):
+    from pgcli import __file__ as package_root
+    package_root = os.path.dirname(package_root)
+
+    pgclirc_file = pgclirc_file or '%sconfig' % config_location()
```
darikg: This should be `pgclirc_file or os.path.join(config_location(), 'config')` instead

pgcli/main.py
```diff
@@ -136,7 +136,10 @@ def __init__(self, force_passwd_prompt=False, never_passwd_prompt=False,
 
         # Initialize completer
         smart_completion = c['main'].as_bool('smart_completion')
-        completer = PGCompleter(smart_completion, pgspecial=self.pgspecial)
+        settings = {
+            'asterisk_column_order':c['main']['asterisk_column_order']}
```
darikg: space after the `:`

pgcli/pgcompleter.py
```diff
@@ -36,12 +36,13 @@ class PGCompleter(Completer):
...
 
-    def __init__(self, smart_completion=True, pgspecial=None):
+    def __init__(self, smart_completion=True, pgspecial=None, settings=None):
         super(PGCompleter, self).__init__()
         self.smart_completion = smart_completion
         self.pgspecial = pgspecial
         self.prioritizer = PrevalenceCounter()
-
+        self.asterisk_column_order = (settings or {}).get(
+            'asterisk_column_order', 'name_order')
```
darikg: 'table_order` not `name_order`, right?

darikg:
Ok, I'm content with the flat `settings` dict. I'll update pgclisublime to supply the settings on initialization. It makes sense that an external pgcompleter user would want to supply its own settings (e.g. through the sublime text json user settings framework) instead of relying on the pgclirc file. I only have a couple of minor points left in comments inline and then this is good to go.

PR #514, opened 2016-05-27
Fix two casing issues with schemas

koljonen:
1. In 'SELECT * FROM Foo.', Foo was not recognized as foo.
2. In 'SELECT * FROM "Foo"', objects in Foo were not suggested.

### DIFF
--- a/pgcli/packages/sqlcompletion.py
+++ b/pgcli/packages/sqlcompletion.py
@@ -291,7 +291,9 @@ def suggest_based_on_last_token(token, text_before_cursor, full_text,
             ('copy', 'from', 'update', 'into', 'describe', 'truncate')):
 
         schema = (identifier and identifier.get_parent_name()) or None
-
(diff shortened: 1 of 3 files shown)


pgcli/packages/sqlcompletion.py
```diff
@@ -291,7 +291,9 @@ def suggest_based_on_last_token(token, text_before_cursor, full_text,
             ('copy', 'from', 'update', 'into', 'describe', 'truncate')):
 
         schema = (identifier and identifier.get_parent_name()) or None
-
+        # If schema name is unquoted, lower-case it
+        if schema and identifier and identifier.value[0] != '"':
```
darikg: Can this just be `if schema and schema[0] != '"'` ?

koljonen: You're right that the identifier check was redundant, so I've removed that, but we do need to use identifier.value for the ", as it's been stripped from schema.

PR #516, opened 2016-05-31
Fix issue with 'SELECT Foo.*<cursor> FROM Foo'

koljonen:
In the expansion, we got '"Foo".\<col\>' when we should have gotten 'Foo.\<col\>'.


koljonen:
@darikg You want to merge this in?

PR #517, opened 2016-06-01
Use pg_proc.proargmodes &c. instead of parsing arg_list

koljonen:
@darikg, please check that this makes sense. It'd be great if we could get rid of all this rather complicated code!

**Original commit message:**
Getting the parameters from proargnames, proallargtypes and proargmodes instead of from parsing the arg_list string simplifies FunctionMetadata quite a bit.
I also made the ColumnMetadata for table/view columns use the same format for the type (i.e. regtype instead of typname). This means we now get join-condition suggestions for joining tables/views to functions, which didn't work before.

### DIFF
--- a/pgcli/packages/function_metadata.py
+++ b/pgcli/packages/function_metadata.py
@@ -49,8 +49,12 @@ class FunctionMetadata(object):
         if self.return_type.lower() == 'void':
             return []
         elif not self.arg_modes:
+            # For functions  without output parameters, the function name
+            # is used as the name of the output column.
+            # E.g. 'SELECT unnest FROM unnest(...);'
             return [ColumnMetadata(self.func_name, self.return_type, [])]
 
-        return [ColumnMetadata(n, t, [])
-            for n, t, m in zip(self.arg_names, self.arg_types, self.arg_modes)
-            if m in ('o', 'b', 't')]
+        return [ColumnMetadata(name, type, [])
+            for name, type, mode in zip(
+                self.arg_names, self.arg_types, self.arg_modes)
+            if mode in ('o', 'b', 't')] # OUT, INOUT, TABLE
--- a/pgcli/packages/function_metadata.py
+++ b/pgcli/packages/function_metadata.py
@@ -1,22 +1,22 @@
-import re
-import sqlparse
 from collections import namedtuple
-from sqlparse.tokens import Whitespace, Comment, Keyword, Name, Punctuation
 
-table_def_regex = re.compile(r'^TABLE\s*\((.+)\)$', re.IGNORECASE)
 ColumnMetadata = namedtuple('ColumnMetadata', ['name', 'datatype', 'foreignkeys'])
 ForeignKey = namedtuple('ForeignKey', ['parentschema', 'parenttable',
     'parentcolumn', 'childschema', 'childtable', 'childcolumn'])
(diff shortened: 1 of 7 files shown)


darikg:
Wow I love how many lines are being deleted here. I'll start going through it in the next couple days

pgcli/packages/function_metadata.py
```diff
@@ -30,122 +30,27 @@ def __ne__(self, other):
...
-        return
-
-    # sql is something like "x int, y text, ..."
-    tokens = sqlparse.parse(sql)[0].flatten()
-    for f in parse_typed_field_list(tokens):
-        if f.name and (not mode_filter or f.mode in mode_filter):
-            yield ColumnMetadata(f.name, None, [])
+        return [ColumnMetadata(n, t, [])
+            for n, t, m in zip(self.arg_names, self.arg_types, self.arg_modes)
+            if m in ('o', 'b', 't')]
```
darikg: Can you add a comment to explain argmodes 'o', 'b', and 't'?

darikg: Also `for name, type, mode` instead of `for n, t, m` would be nice

pgcli/packages/function_metadata.py
```diff
@@ -30,122 +30,27 @@ def __ne__(self, other):
...
+                   self.return_type, self.is_aggregate, self.is_window,
+                   self.is_set_returning))
 
     def fields(self):
         """Returns a list of output-field ColumnMetadata namedtuples"""
 
         if self.return_type.lower() == 'void':
             return []
+        elif not self.arg_modes:
+            return [ColumnMetadata(self.func_name, self.return_type, [])]
```
darikg: Why is the column name assigned as `self.func_name` here?

pgcli/packages/function_metadata.py
```diff
@@ -1,22 +1,22 @@
-import re
-import sqlparse
 from collections import namedtuple
-from sqlparse.tokens import Whitespace, Comment, Keyword, Name, Punctuation
 
-table_def_regex = re.compile(r'^TABLE\s*\((.+)\)$', re.IGNORECASE)
 ColumnMetadata = namedtuple('ColumnMetadata', ['name', 'datatype', 'foreignkeys'])
 ForeignKey = namedtuple('ForeignKey', ['parentschema', 'parenttable',
     'parentcolumn', 'childschema', 'childtable', 'childcolumn'])
+TypedFieldMetadata = namedtuple('TypedFieldMetadata', ['name', 'mode', 'type'])
```
darikg: `TypedFieldMetadata` isn't used anymore, right?

pgcli/packages/function_metadata.py
```diff
@@ -30,122 +30,31 @@ def __ne__(self, other):
...
-        if f.name and (not mode_filter or f.mode in mode_filter):
-            yield ColumnMetadata(f.name, None, [])
+        elif not self.arg_modes:
+            # For functions  without output parameters, the function name
+            # is used as the name of the output column.
+            # E.g. 'SELECT unnest FROM unnest(...);'
+            return [ColumnMetadata(self.func_name, self.return_type, [])]
+
+        return [ColumnMetadata(name, type, [])
+            for name, type, mode in zip(
```
darikg: argh I'm sorry I forgot that `type` is a builtin function in python. Can you use `typ` or `type_` instead? My fault

Issue #519, opened 2016-06-02
Make the "more than 1000 rows" warning limit configurable


mkataja:
Add a configuration option to change the number of rows that causes the following warning to be shown:

> The result set has more than 1000 rows.
> Do you want to continue? [y/N]:

Setting the option to 0 should disable the warning altogether.

drocco007:
@amjith Do you have an opinion about this being an option in the configuration file vs. a command line argument vs. both? The PR from @Smotko implements a command line argument.

amjith:
I don't have a strong opinion. If I had implemented I'd probably have picked the config file.

Smotko:
I've added it to the config file as well.

PR #521, opened 2016-06-02
Make too many rows warning limit configurable

Smotko:
This should resolve https://github.com/dbcli/pgcli/issues/519

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -85,7 +85,7 @@ class PGCli(object):
         if configured_pager:
             self.logger.info('Default pager found in config file: ' + '\'' + configured_pager + '\'')
             os.environ['PAGER'] = configured_pager
-        elif (os_environ_pager):
+        elif os_environ_pager:
             self.logger.info('Default pager found in PAGER environment variable: ' + '\'' + os_environ_pager + '\'')
             os.environ['PAGER'] = os_environ_pager
         else:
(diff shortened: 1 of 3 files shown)


pgcli/main.py
```diff
@@ -516,9 +523,8 @@ def _evaluate_command(self, text):
             logger.debug("headers: %r", headers)
             logger.debug("rows: %r", cur)
             logger.debug("status: %r", status)
-            threshold = 1000
-            if (is_select(status) and
-                    cur and cur.rowcount > threshold):
+            threshold = self.row_limit
+            if (self._should_show_limit_prompt(status, cur)):
```
drocco007: Let's remove the extra parentheses around the if test.

amjith:
@Smotko Thanks for the PR. It looks great. 

@drocco007 Can you review this and merge? I presume you were shepherding @Smotko.

drocco007:
Hey @Smotko! Thanks again for your help!

When you changed how the default is specified, we lost the implicit conversion to `int` that Click does if the default is an integer, so setting the row type via `-R` or the environment variable doesn't work. Fortunately the fix should be pretty easy: just add `type=click.INT` to your option definition. Also, please rebase this your branch against the updated master. Thanks!

cc: @amjith

Smotko:
@drocco007 ah, sorry about that. Should be fixed now.

amjith:
Works great! Thanks for the PR @Smotko. 

🍨

drocco007:
Thanks @Smotko! Nice working with you at PyCon! :)

PR #522, opened 2016-06-02
Filter trigger returning functions from suggestions

### LINKED ISSUE #520: Filter out PG trigger implementation functions
Using master checkouts for pgcli & pgspecial on PG 9.3.5, the Postgres internal trigger functions are being pulled in to the suggestions:

image

Since they're capitalized, they're also floating to the top of the suggestion list. I'm guessing this was inadvertent; it makes the completion list feel cluttered to me.

Perhaps we should exclude functions with the `RI_FKey` the way we currently do for `pg_` items?

@amjith

drocco007:
Closes #520

### DIFF
--- a/pgcli/pgexecute.py
+++ b/pgcli/pgexecute.py
@@ -432,6 +432,7 @@ class PGExecute(object):
                     FROM pg_catalog.pg_proc p
                             INNER JOIN pg_catalog.pg_namespace n
                                 ON n.oid = p.pronamespace
+                    WHERE p.prorettype::regtype != 'trigger'::regtype
                     ORDER BY 1, 2
                     '''
                 _logger.debug('Functions Query. sql: %r', query)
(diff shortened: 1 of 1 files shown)


amjith:
@drocco007 This needs a rebase with master. 

@koljonen Would you mind reviewing this?

drocco007:
@amjith fixed!

amjith:
This is what I see when I tried it locally. 

1__0_0_python2_7_-__pgcli___users_amjith_dropbox_code_python_pgcli____tmux__and_filter_trigger_returning_functions_from_suggestions_by_drocco007_ _pull_request__522_ _dbcli_pgcli

I would have expected tables to be listed at the top and then functions. But I'm seeing all the categories (functions, tables, views etc) combined together and listed alphabetically. I don't think it's an artifact of this PR, but along the way we have lost our grouped sorting and now everything is sorted alphabetically. 

Can we do something to address this?

koljonen:
@amjith I think that maybe if modify the functionality I added in #507 (priority arguments to **find_matches**) a bit, we can use it to get better sorting here. I have a feeling there were a few more modifications I wanted to make to how we sort the suggestions, so I'll have a look at that soonish.

drocco007:
Thanks everybody! :)

PR #524, opened 2016-06-04
Add a pgspecial logging handler

darikg:
Since we split pgspecial off into a separate package, its logging statements weren't being written to the pgcli log file. @j-bennet can you review and merge this?


amjith:
🚀

PR #525, opened 2016-06-07
`column` keyword suggests columns

darikg:
E.g. so `ALTER TABLE foo ALTER COLUMN ` suggests columns from `foo`. 

@koljonen can you review & merge?


amjith:
Thank you for the tests. 🎓

koljonen:
I think **ADD COLUMN** should be excluded.

amjith:
Actually no. We should include the suggestions for ADD COLUMN to make sure the user knows about the existence of a column similar to the one they are adding.

darikg:
Yeah, it's consistent with how e.g. Create table suggests table names. It's actually pretty useful

Issue #430, opened 2015-12-07
Display connection notices asynchronously 


stuartquin:
Currently notices emitted on the connection are output only after the query(ies) have completed: https://github.com/dbcli/pgcli/blob/master/pgcli/pgexecute.py#L291

It would be nice to have the notices appear as they are emitted as they currently do in psql.
This may require some fundamental changes to how queries are executed so will need some investigation.

Example of long running query with notices:

```
test> create function slow_warnings() returns void language plpgsql as $$
      begin
          for i in 1..5 loop
              perform pg_sleep(1);
              raise warning 'i = %', i;
          end loop;
      end
      $$; 
```

Run `select slow_warnings()`

riderinred:
Is there a chance this will be implemented? Just asking because the issue is from some years back.

j-bennet:
Most of the pgcli members, unfortunately, don't have much bandwidth, you understand we all work on pgcli in our free time. We still leave feature requests open, in the hopes that a community member will be invested enough to pick it up and submit a PR. Perhaps that will be you, @riderinred ? :) We welcome any and all contributions!

PR #526, opened 2016-06-07
Fix issue where notices/warnings/... were printed in reverse order

koljonen:
@amjith you want to merge this tiny fix?

Running the function in #430, I noticed the warnings were printed in reverse order.


amjith:
👑

PR #527, opened 2016-06-08
For 'JOIN <cursor>', suggest 'foo on foo.fooid = bar.fooid'

koljonen:
@darikg  you feel like taking a look at this?

**Original commit message:**
This is based on my previous work on suggesting join conditions, but here instead we suggest the whole join. What we do is simply check all the tables in the statement for FK relationships and then suggest joins based on those. I think this will not only save key presses, but also be rather useful when exploring an unfamiliar (part of a) database.

There's one non-obvious change in this commit (that I can think of): When calling **sqlcompletion.text_before_cursor**, the **text_before_cursor** argument now no longer includes **word_before_cursor**. This is because for 'SELECT * FROM foo JOIN bar<cursor>', we would otherwise consider the table **bar** already included in the statement and thus suggest e.g. 'baz on baz.barid = bar.barid'.

### DIFF
--- a/pgcli/packages/parseutils.py
+++ b/pgcli/packages/parseutils.py
@@ -197,7 +197,8 @@ def extract_tables(sql):
     # to have is_function=True
     identifiers = extract_table_identifiers(stream,
                                             allow_functions=not insert_stmt)
-    return tuple(identifiers)
+    # In the case 'sche.<cursor>', we get an empty TableReference; remove that
+    return tuple(i for i in identifiers if i.ref)
 
 
 def find_prev_keyword(sql):
--- a/pgcli/packages/parseutils.py
+++ b/pgcli/packages/parseutils.py
@@ -66,7 +66,9 @@ def last_word(text, include='alphanum_underscore'):
 
 TableReference = namedtuple('TableReference', ['schema', 'name', 'alias',
                                                'is_function'])
-TableReference.ref = property(lambda self: self.alias or self.name)
+TableReference.ref = property(lambda self: self.alias or (
+  self.name if self.name.islower() or self.name[0] == '"'
+  else '"' + self.name + '"'))
 
 
 # This code is borrowed from sqlparse example script.
@@ -140,7 +142,10 @@ def extract_table_identifiers(token_stream, allow_functions=True):
             schema_name = schema_name.lower()
         quote_count = item.value.count('"')
         name_quoted = quote_count > 2 or (quote_count and not schema_quoted)
-        if name and not name_quoted and name != name.lower():
+        alias_quoted = alias and item.value[-1] == '"'
+        if alias_quoted or name_quoted and not alias and name.islower():
+            alias = '"' + (alias or name) + '"'
+        if name and not name_quoted and not name.islower():
             if not alias:
                 alias = name
             name = name.lower()
@@ -198,7 +203,8 @@ def extract_tables(sql):
     identifiers = extract_table_identifiers(stream,
                                             allow_functions=not insert_stmt)
     # In the case 'sche.<cursor>', we get an empty TableReference; remove that
-    return tuple(i for i in identifiers if i.ref)
+    return tuple(i for i in identifiers if i.name)
+
 
 
 def find_prev_keyword(sql):
--- a/pgcli/packages/sqlcompletion.py
+++ b/pgcli/packages/sqlcompletion.py
@@ -22,6 +22,7 @@ Database = namedtuple('Database', [])
 Schema = namedtuple('Schema', [])
 Table = namedtuple('Table', ['schema'])
 JoinCondition = namedtuple('JoinCondition', ['tables', 'parent'])
+Join = namedtuple('Join', ['tables', 'schema'])
 
 Function = namedtuple('Function', ['schema', 'filter'])
 # For convenience, don't require the `filter` argument in Function constructor
@@ -68,6 +69,7 @@ def suggest_type(full_text, text_before_cursor):
 
     full_text = strip_named_query(full_text)
     text_before_cursor = strip_named_query(text_before_cursor)
+    text_before_cursor_including_last_word = text_before_cursor
 
     # If we've partially typed a word then word_before_cursor won't be an empty
     # string. In that case we want to remove the partially typed string before
@@ -78,8 +80,8 @@ def suggest_type(full_text, text_before_cursor):
         if word_before_cursor[-1] == '(' or word_before_cursor[0] == '\\':
             parsed = sqlparse.parse(text_before_cursor)
         else:
-            parsed = sqlparse.parse(
-                text_before_cursor[:-len(word_before_cursor)])
+            text_before_cursor = text_before_cursor[:-len(word_before_cursor)]
+            parsed = sqlparse.parse(text_before_cursor)
 
             identifier = parse_partial_identifier(word_before_cursor)
     else:
@@ -114,7 +116,7 @@ def suggest_type(full_text, text_before_cursor):
         # but the statement won't have a first token
         tok1 = statement.token_first()
         if tok1 and tok1.value == '\\':
-            return suggest_special(text_before_cursor)
+            return suggest_special(text_before_cursor_including_last_word)
 
     last_token = statement and statement.token_prev(len(statement.tokens)) or ''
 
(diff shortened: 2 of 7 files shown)


darikg:
Yup, I'm on it

pgcli/packages/sqlcompletion.py
```diff
@@ -22,6 +22,7 @@
 Schema = namedtuple('Schema', [])
 Table = namedtuple('Table', ['schema'])
 JoinCondition = namedtuple('JoinCondition', ['tables', 'parent'])
```
darikg: Can you add a quick comment here explaining the distinction between Join and JoinCondition?

pgcli/pgcompleter.py
```diff
@@ -384,6 +384,43 @@ def get_column_matches(self, suggestion, word_before_cursor):
...
+        # Set up some data structures for efficient access
+        qualified = dict((t.ref, t.schema) for t in suggestion.tables)
+        tbls = set((t.schema, t.name) for t in scoped_cols.keys())
+        tblprio = dict((t.ref, n) for n, t in enumerate(suggestion.tables))
+        joins, prios = [], []
+        # Iterate over FKs in existing tables to find potential joins
+        for fk, rtbl, rcol in ((fk, rtbl, rcol)
+          for rtbl, rcols in scoped_cols.items()
+          for rcol in rcols
+          for fk in rcol.foreignkeys):
```
darikg: This is really weird looking. I thought at first it was a mis-indented for-loop. Can you either switch to a traditional nested for-loop structure, or assign the generator expression to a named variable & loop over that variable?

pgcli/packages/parseutils.py
```diff
@@ -197,7 +197,8 @@ def extract_tables(sql):
     # to have is_function=True
     identifiers = extract_table_identifiers(stream,
                                             allow_functions=not insert_stmt)
-    return tuple(identifiers)
+    # In the case 'sche.<cursor>', we get an empty TableReference; remove that
```
darikg: Could you add a quick test for this behavior in test_parseutils to prevent a regression?

tests/test_sqlcompletion.py
```diff
@@ -99,14 +99,27 @@ def test_suggests_tables_views_and_schemas(expression):
...
+    assert set(suggestions) == set([
+        Table(schema=None),
+        View(schema=None),
+        Function(schema=None, filter='for_from_clause'),
+        Schema()
+    ])
+
+
+@pytest.mark.parametrize('expression', [
     'SELECT * FROM foo JOIN ',
```
darikg: We probably need another test for a second level join, e.g. `SELECT * FROM foo JOIN bar ON foo.id = bar.id JOIN `

darikg:
This is a nice user experience. We need tests in the two `test_smart_completion` files.

I haven't looked in detail at the text_before_cursor stuff yet, but I'm a little confused at the goal. Let's say we have three tables, `foo`, `bar`, and `qux` that all share a common key. Is the intention that `select * from foo join q` should suggest a join, or not?

Also, any thoughts on how this would interact will a similar feature that would suggest `JOIN ... USING ` completions instead of `JOIN ... ON`? Not saying that has to be done here, just wondering.

koljonen:
The text_before_cursor thing is so that in your example, if we have `select * from foo join qux<cursor>`, we don't get the suggestion `bar on bar.quxid = qux.quxid` (which would clearly be a bug). We do want to get `qux on qux.quxid = foo.quxid` for `select * from foo join q<cursor>`, though.

I'd rather not have both an `ON` suggestion and a `USING` suggestion for the same logical join, so I think a good approach might be to have a config setting that would make `get_join_matches` output a `USING` join whenever the two columns have the same name and an `ON` join otherwise.

darikg:
Sorry about the delay, I was away from my computer all weekend.

> if we have select * from foo join qux<cursor>, we don't get the suggestion bar on bar.quxid = qux.quxid  (which would clearly be a bug).

I'm not so sure about that. With fuzzy matching `select * from oo` suggests the table `foo` so having fuzzy matching also operate in the join conditions makes sense to me too. Actually, it seems it currently is so I'm still a little confused by the changes in sqlcompletion.

Here's a test I just added to `test_smart_completion_public_schema_only`:

```
@pytest.mark.parametrize('text', [
    'SELECT * FROM "Users" u JOIN u',
    'SELECT * FROM "Users" u JOIN uid',
    'SELECT * FROM "Users" u JOIN userid',
    'SELECT * FROM "Users" u JOIN id',
])
def test_suggested_joins_fuzzy(completer, complete_event, text):
    position = len(text)
    result = set(completer.get_completions(
        Document(text=text, cursor_position=position),
        complete_event))
    last_word = text.split()[-1]
    expected = Completion(text='users ON users.id = u.userid',
                          start_position=-len(last_word), display_meta='join')
    assert expected in result
```

All 4 of the conditions pass, which I think is correct behavior, but you're saying only the first should pass?

Also, I just noticed that join condition suggestions support either FK relationships, or name matches, but the join suggestions are only FK-based. Is that intentional?

koljonen:
Sorry, I think I was a bit unclear. The `text_before_cursor` thing is just there to prevent the generation of invalid code such as `select * from foo join bar on bar.quxid = qux.quxid`. That is, without it we'd consider `text_before_cursor` as one of the tables already present in the from clause.

I did leave out name-based joins intentionally, as I thought there might just be too many of them and that they'd probably not be all that useful. It is quite possible I'll change my mind on that after using this feature for a while.

tests/test_sqlcompletion.py
```diff
@@ -99,14 +99,42 @@ def test_suggests_tables_views_and_schemas(expression):
...
+        Table(schema=None),
+        View(schema=None),
+        Function(schema=None, filter='for_from_clause'),
+        Join(((None, 'foo', None, False), (None, 'bar', None, False)), None),
+        Schema(),
+    ])
+
+
+@pytest.mark.parametrize('expression', [
+    'SELECT * FROM foo JOIN '
```
darikg: Can you add one extra condition here like `'SELECT * FROM foo JOIN bar'`. That passes with your current PR, but fails if we revert the changes to the `text_before_cursor` stuff.

darikg:
Ahh I finally get it. Sorry for being slow. Can you add a test to capture that behavior? I just pointed out a spot where it's really easy to do that. Could you also add that test I suggested in the previous comment? Then this will be good to merge

koljonen:
While working on adding aliases for self joins, I noticed some problems related to casing and quoted identifiers, so I had to do some more changes.

pgcli/packages/parseutils.py
```diff
@@ -66,7 +66,9 @@ def last_word(text, include='alphanum_underscore'):
 
 TableReference = namedtuple('TableReference', ['schema', 'name', 'alias',
                                                'is_function'])
-TableReference.ref = property(lambda self: self.alias or self.name)
+TableReference.ref = property(lambda self: self.alias or (
```
darikg: Not necessary for this PR, but long-term if this gets any more complicated we should think about replacing the namedtuple based class with an actual standalone classdef with docstrings etc.

pgcli/pgcompleter.py
```diff
@@ -384,38 +385,50 @@ def get_column_matches(self, suggestion, word_before_cursor):
 
         return self.find_matches(word_before_cursor, flat_cols, meta='column')
 
+    def generate_alias(self, tbl, tbls):
```
darikg: I think this would be better off as a standalone function since it doesn't use any pgcompleter member variables. A docstring would help a lot.

darikg:
The auto-aliasing for self joins is pretty cool. It's too bad it's basically impossible to generalize to suggesting aliases in all cases. Like if it could some know that after `FROM foo f JOIN ` what I really want  is `JOIN bar b ON f.id = b.id` instead of `JOIN bar ON f.id = bar.id`. Just daydreaming out loud here.

koljonen:
I've actually experimented a bit with alias generation (FooBar -> FB, baz_qux -> bq). Those were just toggled with a config setting, though. I put that branch on hold after running into some complications, but I'll probably get back to it pretty soon.

darikg:
Wow, that's awesome. Let's just move `generate_alias` into a standalone function with a basic docstring and then we can merge this. Hopefully more sophisticated alias generation will be a future PR someday :pray:

pgcli/pgcompleter.py
koljonen: It does use self.case.

darikg:
Awesome

PR #528, opened 2016-06-16
Fix some join-condition issues

koljonen:
@darikg I think this one's for you.

**Commit message:**
When self-joining a table with an FK to or from some other table, we got a false FK-join suggestion for that column.
There was also a problem with quoted tables not being quoted in the join condition.
And there were a couple of problems when trying to join a non-existent table or using a non-existent qualifier (`SELECT * FROM Foo JOIN Bar ON Meow.`).

I also rewrote get_join_condition_matches a bit in the process, hopefully making it a bit simpler.


darikg:
Looks good!

PR #529, opened 2016-06-19
Remove duplication from smart completion tests

koljonen:
@darikg you mind taking a look at this?

This PR has two parts (plus some cosmetic changes):
1. The first commit just consolidates existing duplicated logic in the two test_smart_completion* files into a new file. 
2. Then there are a six commits that make it so that, for some type of completion (table, column, ...), instead of listing every individual completion we expect to get, we just say what kind of completions to get from the test `metadata` dict. E.g. instead of listing every table in the schema `foo` (say for `SELECT * FROM foo.<cursor>`), we simply say we expect `testdata.tables('foo')`. This means we get to write less test code and that fewer tests will break when we expand the test data.

### DIFF
--- a/tests/metadata.py
+++ b/tests/metadata.py
@@ -36,6 +36,14 @@ class MetaData(object):
     def keywords(self, pos=0):
         return [keyword(kw, pos) for kw in self.completer.keywords]
 
+    def columns(self, parent, schema='public', typ='tables', pos=0):
+        if typ == 'functions':
+            fun = [x for x in self.metadata[typ][schema] if x[0] == parent][0]
+            cols = fun[1]
+        else:
+            cols = self.metadata[typ][schema][parent]
+        return [column(escape(col), pos) for col in cols]
+
     def datatypes(self, schema='public', pos=0):
         return [datatype(escape(x), pos)
             for x in self.metadata.get('datatypes', {}).get(schema, [])]
--- a/tests/metadata.py
+++ b/tests/metadata.py
@@ -48,6 +48,10 @@ class MetaData(object):
         return [view(escape(x), pos)
             for x in self.metadata.get('views', {}).get(schema, [])]
 
+    def functions(self, schema='public', pos=0):
+        return [function(escape(x[0]), pos)
+            for x in self.metadata.get('functions', {}).get(schema, [])]
+
     def schemas(self, pos=0):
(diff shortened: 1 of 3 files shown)


tests/test_smart_completion_multiple_schemata.py
```diff
@@ -1,83 +1,53 @@
...
-    comp.extend_schemata(schemata)
-    comp.extend_relations(tables, kind='tables')
-    comp.extend_columns(columns, kind='tables')
-    comp.extend_functions(functions)
-    comp.extend_datatypes(datatypes)
-    comp.extend_foreignkeys(foreignkeys)
-    comp.set_search_path(['public'])
-
-    return comp
+    return testdata.completer
```
darikg: This makes every test share the same completer object. This would be bad if the completer's state changes during a test, because then the tests depend on what order they're run in. I think what we want here is `return testdata.get_completer()` and not have the MetaData object store a reference to the completer

darikg:
I think on one hand it's a good thing to have very explicit tests and too much abstraction can obscure what's actually being tested, but on the other hand these tests were definitely super verbose and that hampered readability and extensibility. On the whole I'm in favor of this. What do you think @amjith?

amjith:
I'm a huge fan of this change. I love how the tests are not only succinct but they're also make logical sense. 

I think the @darikg's suggestion to return the completer object but not storing it should get around the idea of the completer being mutated by a test. 

Thanks for making the change @koljonen. 👍

koljonen:
Yeah, that conflict between explicitness and DRYness is why I decided to autogenerate only the most trivial cases, i.e. those that can be done with a simple lookup in the `metadata` dict. I did try autogenerating the wildcard expansions as well, but I felt that added a bit too much complexity for too little benefit.

tests/metadata.py
```diff
@@ -0,0 +1,108 @@
...
+            for x in self.metadata.get('functions', {}).get(schema, [])]
+
+    def schemas(self, pos=0):
+        schemas = set(sch for schs in self.metadata.values() for sch in schs)
+        return [schema(escape(s), pos=pos) for s in schemas]
+
+    def get_completer(self):
+        metadata = self.metadata
+        import pgcli.pgcompleter as pgcompleter
+        self.completer = comp = pgcompleter.PGCompleter(smart_completion=True)
```
darikg: don't assign self.completer here

Issue #518, opened 2016-06-02
Add configuration option to disable startup and exit messages


mkataja:
Add a configuration option to disable some or all of the following startup and exit messages to reduce the amount of output:

> Version: 0.20.1
> Chat: https://gitter.im/dbcli/pgcli
> Mail: https://groups.google.com/forum/#!forum/pgcli
> Home: http://pgcli.com

> Goodbye!

The messages could still be enabled by default.

amjith:
This is now available in v1.0. Please upgrade. 

```
pip install -U pgcli
```

PR #530, opened 2016-06-19
Ported less_chatty option from mycli.

j-bennet:
This ports `less_chatty` config option and `NONE` log level from `mycli` to address issue #518.

Related PR:
https://github.com/dbcli/mycli/pull/268


amjith:
It looks so naked without the startup messages. 😢

j-bennet:
Yes, it does! But we got this request in both pgcli and mycli. I assume it's a wanted feature for some.

PR #531, opened 2016-06-20
Suggest table aliases + add tests for casing

koljonen:
@darikg, you showed some interest in this, so would you like to take a look?

**Original commit message:**
If config.main.generate_aliases is True, for `SELECT * FROM `, we suggest `FooBar FB` and `foo_bar fb` instead of `FooBar` and `foo_bar`, respectively.
To be able to add a test, I had to add support for testing with different settings, which meant I could also add tests for casing.

There are two non-obvious changes that I can think of:
1. The lexical sorting of matches is modified so as to sort spaces and underscores before letters and to sort case-insensitively. This is so that e.g `Foob F` comes before 'FooBar FB' when `foob` is input.
2. We now suggest `some_func()` instead of `some_func` (because suggesting `some_func sf` didn't make any sense).

### DIFF
--- a/pgcli/packages/sqlcompletion.py
+++ b/pgcli/packages/sqlcompletion.py
@@ -33,7 +33,7 @@ Join = namedtuple('Join', ['tables', 'schema'])
 
 Function = namedtuple('Function', ['schema', 'tables', 'filter'])
 # For convenience, don't require the `filter` argument in Function constructor
-Function.__new__.__defaults__ = (None, None, tuple())
+Function.__new__.__defaults__ = (None, tuple(), None)
 Table.__new__.__defaults__ = (None, tuple())
 View.__new__.__defaults__ = (None, tuple())
 FromClauseItem.__new__.__defaults__ = (None, tuple())
--- a/pgcli/packages/sqlcompletion.py
+++ b/pgcli/packages/sqlcompletion.py
@@ -20,20 +20,27 @@ else:
 Special = namedtuple('Special', [])
 Database = namedtuple('Database', [])
 Schema = namedtuple('Schema', [])
-Table = namedtuple('Table', ['schema'])
+# FromClauseItem is a table/view/function used in the FROM clause
+# `tables` contains the list of tables/... already in the statement,
+# used to ensure that the alias we suggest is unique
+FromClauseItem = namedtuple('FromClauseItem', 'schema tables')
+Table = namedtuple('Table', ['schema', 'tables'])
+View = namedtuple('View', ['schema', 'tables'])
 # JoinConditions are suggested after ON, e.g. 'foo.barid = bar.barid'
 JoinCondition = namedtuple('JoinCondition', ['tables', 'parent'])
 # Joins are suggested after JOIN, e.g. 'foo ON foo.barid = bar.barid'
 Join = namedtuple('Join', ['tables', 'schema'])
 
-Function = namedtuple('Function', ['schema', 'filter'])
+Function = namedtuple('Function', ['schema', 'tables', 'filter'])
 # For convenience, don't require the `filter` argument in Function constructor
-Function.__new__.__defaults__ = (None, None)
+Function.__new__.__defaults__ = (None, None, tuple())
+Table.__new__.__defaults__ = (None, tuple())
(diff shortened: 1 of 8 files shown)


darikg:
Yup I'm definitely interested, will start going through it in the next couple days

pgcli/pgcompleter.py
```diff
@@ -385,15 +392,27 @@ def get_column_matches(self, suggestion, word_before_cursor):
...
-    def generate_alias(self, tbl, tbls):
+    def alias(self, tbl, tbls):
         """ Generate a unique table alias
         tbl - name of the table to alias, quoted if it needs to be
-        tbls - set of table refs already in use, normalized with normalize_ref
+        tbls - TableReference iterable of tables already in query
         """
+        tbl = self.case(tbl)
+        tbls = set(normalize_ref(t.ref) for t in tbls)
+        if self.generate_aliases:
```
darikg: Let's break off the specific chunk of code that transforms a table name into an alias into a standalone function. That way 1) it's easy to test just the alias generation, and 2) savvy users will be able to monkey patch in their own custom alias generation by replacing just that function

pgcli/packages/sqlcompletion.py
```diff
@@ -20,20 +20,27 @@
 Special = namedtuple('Special', [])
 Database = namedtuple('Database', [])
 Schema = namedtuple('Schema', [])
+# AliasableTable is used in those cases we might want to generate an alias
+AliasableTable = namedtuple('AliasableTable', 'schema tables')
```
darikg: Can we add a `aliasable_tables` field (default None) to Table, View, and Function instead of separate suggestion types?

darikg: Could we also add a comment explaining that the tables field is accomplishing here? If I understand correctly, correctly suggesting a table & alias requires knowing what aliases have already been used

tests/test_sqlcompletion.py
```diff
@@ -25,7 +26,7 @@ def test_select_suggests_cols_with_qualified_table_scope():
 @pytest.mark.parametrize('expression', [
     'SELECT * FROM "tabl" WHERE ',
 ])
-def test_where_suggests_columns_functions_quoted_table(expression):
+def Aliasabletest_where_suggests_columns_functions_quoted_table(expression):
```
darikg: There's a bunch of test renames like this... regex replace mistake?

darikg:
It just occurred to me that another way to do this would be to make the alias a separate suggestion AFTER the table/view/function suggestion. You would just check if `last_token` is an identifier, and then check get_previous_keyword to see if we're in a state where an alias would be valid.

koljonen:
I've rebased and reorganized a bit.

I think the user experience of getting the alias automatically is better; particularly for join suggestions.

I tried a new approach with a FromClauseItem suggestion type combining the three aliasable types. The main benefit of this is that the logic for figuring out whether to add an alias becomes a lot simpler. (Code where `None`, `[]` and `not hasattr` aren't equivalent is bound to be misunderstood ...)

pgcli/packages/sqlcompletion.py
```diff
@@ -20,20 +20,27 @@
...
 # JoinConditions are suggested after ON, e.g. 'foo.barid = bar.barid'
 JoinCondition = namedtuple('JoinCondition', ['tables', 'parent'])
 # Joins are suggested after JOIN, e.g. 'foo ON foo.barid = bar.barid'
 Join = namedtuple('Join', ['tables', 'schema'])
 
-Function = namedtuple('Function', ['schema', 'filter'])
+Function = namedtuple('Function', ['schema', 'tables', 'filter'])
 # For convenience, don't require the `filter` argument in Function constructor
-Function.__new__.__defaults__ = (None, None)
+Function.__new__.__defaults__ = (None, None, tuple())
```
darikg: Should this be `(None, tuple(), None)`? Since `tables` is the second arg?

darikg:
I'll go over this more carefully tomorrow. The `FromClauseItem` approach makes a lot of sense!

darikg:
Sorry, this slipped off my radar. This is great!

PR #532, opened 2016-06-21
Some sqlcompletion refactoring

darikg:
The function signature for `suggest_based_on_last_token` was starting to get pretty unwieldy so I wrapped most of those variables into a standalone class unimaginably named `SqlDocument`. This cleaned up  `suggest_based_on_last_token` a little bit and also formalized a couple of repeated patterns there, specifically schema name extraction and the reduction of `text_before_cursor` to a more simplified version by finding the preceding token.

### DIFF
--- a/pgcli/packages/sqlcompletion.py
+++ b/pgcli/packages/sqlcompletion.py
@@ -115,15 +115,18 @@ def suggest_type(full_text, text_before_cursor):
     return suggest_based_on_last_token(doc.last_token, doc)
 
 
+named_query_regex = re.compile(r'^\s*\\ns\s+[A-z0-9\-_]+\s+')
+
+
 def _strip_named_query(txt):
     """
     This will strip "save named query" command in the beginning of the line:
     '\ns zzz SELECT * FROM abc'   -> 'SELECT * FROM abc'
     '  \ns zzz SELECT * FROM abc' -> 'SELECT * FROM abc'
     """
-    pattern = re.compile(r'^\s*\\ns\s+[A-z0-9\-_]+\s+')
-    if pattern.match(txt):
-        txt = pattern.sub('', txt)
+
+    if named_query_regex.match(txt):
+        txt = named_query_regex.sub('', txt)
     return txt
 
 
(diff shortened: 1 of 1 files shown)


pgcli/packages/sqlcompletion.py
```diff
@@ -42,6 +42,54 @@
 Path = namedtuple('Path', [])
 
 
+class SqlDocument(object):
```
amjith: Should this be called SqlStatement() instead of SqlDocument() ?

pgcli/packages/sqlcompletion.py
```diff
@@ -53,46 +101,40 @@ def suggest_type(full_text, text_before_cursor):
...
-    # sending it to the sqlparser. Otherwise the last token will always be the
-    # partially typed string which renders the smart completion useless because
-    # it will always return the list of keywords as completion.
-    if word_before_cursor:
-        if word_before_cursor[-1] == '(' or word_before_cursor[0] == '\\':
-            parsed = sqlparse.parse(text_before_cursor)
-        else:
-            text_before_cursor = text_before_cursor[:-len(word_before_cursor)]
-            parsed = sqlparse.parse(text_before_cursor)
+    doc = SqlDocument(full_text, text_before_cursor)
```
amjith: If we decide to go with SqlStatement() as the classname this should also be renamed to `stmt` instead of `doc`.

amjith:
I like this change. Keeps the code organized. Once we decide on whether to call it `SqlDocument` or `SqlStatement` we can merge this.

pgcli/packages/sqlcompletion.py
darikg: I started with Document because `full_text`/`text_before_cursor` could contain multiple `;` separated statements. But then the constructor parses the document and throws alway all but the current statement so Statement does seem more accurate.

darikg:
Ok, SqlStatement was better. Updated.

amjith:
Looks great. ♦️ ♦️

PR #533, opened 2016-06-21
Fix link to Click

mortonfox:
Click 3 is no more. I'd suggest linking to the root of the Click website and letting it redirect to the current stable version since they seem to be removing pages for the older versions.


amjith:
You're right. I was just being lazy and copied whatever was in my url bar. 

Thanks for fixing the link. 🚀

Issue #536, opened 2016-06-24
Crash when entering standard psql connect string


svenstaro:
I'm getting this:

```
pgcli 'postgresql:secretpassword@secrethostname:5432/mydatabase'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/lib/python3.5/site-packages/click/core.py", line 716, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python3.5/site-packages/click/core.py", line 696, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python3.5/site-packages/click/core.py", line 889, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python3.5/site-packages/click/core.py", line 534, in invoke
    return callback(*args, **kwargs)
  File "/usr/lib/python3.5/site-packages/pgcli/main.py", line 728, in cli
    pgcli.connect_uri(database)
  File "/usr/lib/python3.5/site-packages/pgcli/main.py", line 265, in connect_uri
    self.connect(*list(map(lambda p: unquote(p) if p else p, arguments)))
  File "/usr/lib/python3.5/site-packages/pgcli/main.py", line 265, in <lambda>
    self.connect(*list(map(lambda p: unquote(p) if p else p, arguments)))
  File "/usr/lib/python3.5/urllib/parse.py", line 537, in unquote
    if '%' not in string:
TypeError: argument of type 'int' is not iterable
```

This happens with the latest release as well as the latest git commit.

amjith:
What version of pgcli are you using? Can you try upgrading to the latest? 

```
pip install -U pgcli
```

svenstaro:
At the bottom of my issue I said it was using HEAD at the time and the newest release at the time (which was v0.20.1).

amjith:
@svenstaro Fix available in master. Can you install from master? 

```
pip install -U https://github.com/dbcli/pgcli/archive/master.zip
```

svenstaro:
Fix works for me, thanks!

PR #538, opened 2016-06-29
Repair exception thrown on urls with ports

eswald:
When the database URL contains a port, uri.port is (at least in Python 2.7.6) an integer, not a string, so urlparse.unquote chokes on it.
Fixes issue #536, but is probably worth verifying on Python 3.


amjith:
Works great. Thanks for the quick fix. 🍨

PR #543, opened 2016-07-06
Better scoping for tables in insert statements

koljonen:
@amjith mind taking a look?

**Original commit message:**
This commit makes it so that given `INSERT INTO foo(<cursor1>) SELECT <cursor2> FROM bar;`, we suggest `bar` columns for `<cursor2>` and `foo` columns for `<cursor1>`. Previous behaviour is sugggesting columns from both tables in both cases.


amjith:
Looks good to me. 🎄

PR #546, opened 2016-07-09
Replace timestamptz alias since AWS Redshift does not support it

majorgreys:
When I tried connecting to Redshift with pgcli v 1.0 I got a `alias timestamptz not supported` error. I made this change to avoid this incompatibility.


koljonen:
Thanks! I'll just merge this in.

PR #548, opened 2016-07-15
Show host and user information in prompt

limdauto:
I can also make this configurable if you are keen

screen shot 2016-07-15 at 12 14 08 pm

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -463,7 +463,8 @@ class PGCli(object):
             set_vi_mode_enabled=set_vi_mode)
 
         def prompt_tokens(_):
-            return [(Token.Prompt, '%s> ' % self.pgexecute.dbname)]
+            return [(Token.Prompt, '{user}@{host}:{dbname}> '.format(
+                user=self.pgexecute.user, host=self.pgexecute.host, dbname=self.pgexecute.dbname))]
 
         def get_continuation_tokens(cli, width):
             return [(Token.Continuation, '.' * (width - 1) + ' ')]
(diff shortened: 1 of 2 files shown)


j-bennet:
We already made it configurable in mycli:

https://github.com/dbcli/mycli/blob/master/mycli/myclirc#L53-L60

I'd rather port the same functionality to pgcli.

limdauto:
@j-bennet done!

j-bennet:
@limdauto You also need a default there, for users that don't have it in config.

limdauto:
@j-bennet Cool I will add it this weekend. It was just that when I read existing code, it doesn't seem to concern with defaults too much. mycli seems much more robust.

limdauto:
@j-bennet updated!

amjith:
> It was just that when I read existing code here, it doesn't seem too concerned with defaults. mycli seems much more robust in that regards.

I want to address why we didn't deal with defaults in pgcli. It's because when you try to read a value from the config file it is always superimposed on top of the default config file (pgclirc) that is shipped with every version of pgcli. 

So a missing value for prompt in the user's config file won't result in an exception because it will be supplied by the default `pgclirc` file. So even though it is a good practice to supply the default values (as requested by j-bennet) it is not strictly required in order to work.

I'd like to think that we try to maintain the same level of robustness for both mycli as well as pgcli. So if you find that one or the other is lacking in quality, I'll be happy to address them.

amjith:
The functionality in this PR looks great!! Thank you for taking the time to contribute. 🍰 

I'll let @j-bennet do the final review and merge this in.

j-bennet:
Looks good to me. I'd say go ahead and also port overriding prompt at run-time by providing a command line parameter, like mycli does:

https://github.com/dbcli/mycli/blob/master/mycli/main.py#L724-L726

but that's probably an overkill until someone asks for that feature.

Merging.

👍

PR #549, opened 2016-07-20
Upgrade to sqlparse-0.2.0, required for Python-3.5 support

dev-zero:
The tests pass with python-2.7 and python-3.5, but other than that I didn't do much testing.

### DIFF
--- a/pgcli/packages/parseutils.py
+++ b/pgcli/packages/parseutils.py
@@ -296,7 +296,7 @@ def parse_partial_identifier(word):
     n_tok = len(p.tokens)
     if n_tok == 1 and isinstance(p.tokens[0], Identifier):
         return p.tokens[0]
-    elif p.token_next_match(0, Error, '"'):
+    elif p.token_next_by(m=(Error, '"'))[1]:
         # An unmatched double quote, e.g. '"foo', 'foo."', or 'foo."bar'
         # Close the double quote, then reparse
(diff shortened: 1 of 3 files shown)


darikg:
Nice! Those test failures look unrelated to the PR, let me see if I can figure out what's wrong

darikg:
@amjith Any idea what's going on here? All the test failures are tests with unicode in `test_pgexecute`. I can reproduce it locally, but upgrading sqlparse doesn't change any other dependencies and the PR doesn't touch `pgexecute` :confused: I can fix the failing tests by adding `from __future__ import unicode_literals` but that breaks another test.

SebCorbin:
👍

amjith:
I wonder if sqlparse is enforcing some encoding while it parses the query. I'll test this locally to see what it does with tables with unicode. 

Sorry about the delayed reply. I have family visiting from India and it's hard to find time. :(

amjith:
Yup! sqlparse is being "helpful" by automatically casting strings as unicode while [splitting](https://github.com/dbcli/pgcli/blob/master/pgcli/pgexecute.py#L220) them, so the tests are getting messed up. 

We should be explicitly casting the strings to unicode. I'm going to merge this PR and then update the tests with a subsequent PR. 

Thanks for the PR @dev-zero. 🎌

PR #553, opened 2016-07-27
CTE-aware suggestions

darikg:
The general idea here is to take a query like
```
WITH my_cte AS (
   SELECT foo, baz FROM bar WHERE foo > 1
)
SELECT * FROM my_cte WHERE
```
and simplify it into either 
```SELECT foo, baz FROM bar WHERE foo > 1```
or 
```SELECT * FROM my_cte WHERE```
 depending on whether the cursor is inside or outside the CTE body. If outside, parse the fieldnames and cte names and represent them in a manner analogous to real database tables. That way, they can be injected into the completion machinery, so CTE names get suggested alongside table names, and CTE column names get suggested when appropriate. Currently, CTEs are not suggested for joins or join conditions, but I'll fix that in a follow up PR.


amjith:
This is quite the PR. I'll probably take some time to review this, but I'll get started tonight.

amjith:
Sorry about taking so long. Last week has not been quite busy with the kids. I'll try and find some time to review this soon.

amjith:
👏  This has got to be the best PR I've seen in my life. 

Nicely broken up commits that clearly show a progression of the feature. I'm very very happy that you took the time to make sensible commits so the reviewer doesn't get lost.

Very well done. I'm especially thankful that you thought about preserving the blame history by duplicating the file. 

I just have one question, did you break them up into nice bite sized commits after you did the entire feature or did this progression occur naturally? Because this is a clear showcase of well thought out design and implementation.

☕ 🍵 🍨 🍻

darikg:
Thanks! It's a pretty heavily rebased and reordered feature branch. There were a lot of false starts and dead ends. Once sqlparse 0.20 was released I rewrote a bunch of the CTE processing code and at that point everything was such a mess I reordered everything with an eye towards a readable history. Glad it paid off.

PR #555, opened 2016-07-28
configurable null format

avdd:
Allows a configurable string for the null result. 

I like to use `⊥` to visually distinguish from normal text, so the config value needs to be decoded to unicode.  Such decoding should probably be applied to other formatting options, e.g. the prompt.

Ideally, the output formatting would be themeable to make nulls, and other types, more distinguishable!


avdd:
Apologies for duplicate: I ballsed up the commit metadata

PR #556, opened 2016-07-28
configurable null format

avdd:
Allows a configurable string for the null result.

I like to use ⊥ to visually distinguish from normal text, so the config value needs to be decoded to unicode. Such decoding should probably be applied to other formatting options, e.g. the prompt.

Ideally, the output formatting would be themeable to make nulls, and other types, more distinguishable!

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -85,6 +85,7 @@ class NullHandler(logging.Handler):
 class PGCli(object):
 
     default_prompt = '\\u@\\h:\\d> '
+    default_null_string = '<null>'
 
     def set_default_pager(self, config):
         configured_pager = config['main'].get('pager')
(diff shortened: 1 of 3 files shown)


amjith:
@avdd Would you mind taking a look at the failing tests?

avdd:
I gave up on unicode -- I don't know how the project wants to deal with python 2/3 compatibility.
The suggestion by @koljonen to use ANSI colours works well enough for me, although that approach seems brittle: `^[[35;1m` is a valid ANSI escape that isn't handled properly.

amjith:
@avdd Thank you for reporting the unicode issue. I dug into this a little bit and the problem comes from the fact that `ConfigObj` that is used to read the config file was assuming an ascii encoding. Once I changed that it works great with the unicode characters. 

If you'd like to make the change yourself and add a few tests to check if unicode chars are allowed, I'll be most happy. But if you're short for time, I'll be happy to merge this PR and made the necessary changes myself. 

This is the line you have to change. https://github.com/dbcli/pgcli/blob/master/pgcli/config.py#L17

That should be changed to the following: 

```
    cfg.merge(ConfigObj(expanduser(usr_cfg), interpolation=False, encoding='utf-8'))
```

Let me know how you'd like to proceed. 

Thanks again for taking the time to add this feature.

avdd:
@amjith I tried, but I can't immediately think of a clean way to test: configobj seems to only work with filenames (!) and pytest's tmpdir appears not to do encodings (!).

This PR is small and does what I need: to be able to visually distinguish nulls.  Over to you!

Cheers! :-)

amjith:
That's fair. 😃

PR #557, opened 2016-07-29
Add command-line option --single-connection

koljonen:
This option makes it so that the executor and the completer use the same database connection. This is needed for the completer to function when using one-time passwords for connecting to the database.

@darikg wanna have a look?


darikg:
Yeah I'll have a look in the next couple days hopefully

darikg:
Sorry for the delay -- this is totally reasonable

Issue #544, opened 2016-07-06
\watch command does not work


gabfl:
(this is a minor issue as `\watch` is not a well known command)

PostgreSQL has a `\watch` command to repeat queries every X seconds (as documented here): https://www.postgresql.org/docs/current/static/app-psql.html

It works perfectly on the psql CLI:

``` sql
gab@db > SELECT 1; \watch 1;
 ?column? 
----------
        1
(1 row)

Time: 0.187 ms
Watch every 1s  Wed Jul  6 12:01:54 2016

 ?column? 
----------
        1
(1 row)

Time: 0.047 ms
Watch every 1s  Wed Jul  6 12:01:55 2016

 ?column? 
----------
        1
(1 row)

Time: 0.154 ms
Watch every 1s  Wed Jul  6 12:01:56 2016

 ?column? 
----------
        1
(1 row)

Time: 0.164 ms
```

It does not work with pgcli:

``` sql
db> SELECT 1; \watch 1;
+------------+
|   ?column? |
|------------|
|          1 |
+------------+
SELECT 1
syntax error at or near "\"
LINE 1: \watch 1
        ^

Time: 0.003s
```

amjith:
This has not been implemented yet. But this is something we want to support.

gabfl:
great, thanks Amjith!

amjith:
Please upgrade to v1.2.0. 

```
pip install -U pgcli
```

This issue is addressed in the new version.

gabfl:
Just tested and it works great, thank you so much

levin-kelevra:
Version: 1.5.1
Doesn't work.
ОШИБКА:  ошибка синтаксиса (примерное положение: "/")
LINE 1: /watch 1

ERROR: Syntax error

amjith:
@j-bennet Can you take a look at this? I don't have a working postgres on my new computer to debug this.

j-bennet:
@levin-kelevra Were you entering `/watch` or `\watch`? The error seems to suggest that it was a forward slash. It should be a backslash.

j-bennet:
I'm going to leave the issue closed since there's been no update. @levin-kelevra, feel free to reopen if you're still having problems.

ivyleavedtoadflax:
This also does not work for me in 1.5.1 on OSX:

```
select * from users where id >12;\watch 1;
...
SELECT 2
syntax error at or near "\"
LINE 1: \watch 1
        ^
```

[...]

PR #561, opened 2016-08-03
Issue #544 Implementation of watch command

stuartquin:
**Note** I don't think the tests will pass until the related https://github.com/dbcli/pgspecial/pull/22 is merged. @amjith what would you normally do in this case?

An implementation of `\watch` from psql.

```
SELECT * FROM titles; \watch 1;
```

If the command can be parsed as a 'watch' command it is executed, then we sleep the specified number of seconds then execute again. Ctrl+C cancels

Any feedback and suggestions on testing welcome

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -11,7 +11,7 @@ import threading
 import shutil
 import functools
 import humanize
-from time import time
+from time import time, sleep
 from codecs import open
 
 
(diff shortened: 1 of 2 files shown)


amjith:
@stuartquin Sorry I haven't had a chance to respond. I'll take a look at this soon (this weekend probably).

pgcli/main.py
```diff
@@ -386,65 +450,17 @@ def run_cli(self):
...
-                except Exception as e:
-                    logger.error("sql: %r, error: %r", document.text, e)
-                    logger.error("traceback: %r", traceback.format_exc())
-                    click.secho(str(e), err=True, fg='red')
+                watch_command, timing = special.get_watch_command(document.text)
+                if watch_command:
+                    while watch_command:
+                        try:
+                            query = self.execute_command(watch_command)
+                            watch_command, timing = special.get_watch_command(document.text)
```
amjith: Do we need this inside the `while` loop? Looks like we're already parsing the watch_command outside the loop and I don't anticipate that it will change.

amjith:
Tested this branch and it works as expected. Let's wait to release pgspecial and fix the repeated parsing of the watch command. 

I'm also wondering if it would be nice to add a info message that says, "Waiting for 'x' seconds before repeating command."

amjith:
pgspecial 1.6.0 has been released. If you update the setup.py file this should pass.

amjith:
Works great. ⏰

PR #566, opened 2016-08-23
Fix crash bug with named queries

koljonen:
And add behave tests.

@amjith you want to merge this?


amjith:
Looks good. Thanks! 

🏁

Issue #563, opened 2016-08-11
Exception on `SELECT column AS` 


LiraNuna:
When inputting the query `SELECT column AS`, the following exception raises:

```
analytics> SELECT x ASException in thread Thread-11:
Traceback (most recent call last):                                  
  File "/usr/lib/python3.4/threading.py", line 920, in _bootstrap_inner
    self.run()                                                      
  File "/usr/lib/python3.4/threading.py", line 868, in runs-mode                                                                                                                                                                                                                     
    self._target(*self._args, **self._kwargs)
  File ".pex/install/prompt_toolkit-1.0.5-py3-none-any.whl.bd98d0b047ddd5ae887af8c17dfa67aa1127eb00/prompt_toolkit-1.0.5-py3-none-any.whl/prompt_toolkit/interface.py", line 836, in run
    completions = list(buffer.completer.get_completions(document, complete_event))
  File ".pex/install/pgcli-1.1.1b1-py3-none-any.whl.79f0fc40898c1538f1570abfc24c9a7f0cf0efbf/pgcli-1.1.1b1-py3-none-any.whl/pgcli/pgcompleter.py", line 351, in get_completions
    suggestions = suggest_type(document.text, document.text_before_cursor)
  File ".pex/install/pgcli-1.1.1b1-py3-none-any.whl.79f0fc40898c1538f1570abfc24c9a7f0cf0efbf/pgcli-1.1.1b1-py3-none-any.whl/pgcli/packages/sqlcompletion.py", line 129, in suggest_type
    stmt = SqlStatement(full_text, text_before_cursor)
  File ".pex/install/pgcli-1.1.1b1-py3-none-any.whl.79f0fc40898c1538f1570abfc24c9a7f0cf0efbf/pgcli-1.1.1b1-py3-none-any.whl/pgcli/packages/sqlcompletion.py", line 73, in __init__
    self.identifier = parse_partial_identifier(word_before_cursor)
  File ".pex/install/pgcli-1.1.1b1-py3-none-any.whl.79f0fc40898c1538f1570abfc24c9a7f0cf0efbf/pgcli-1.1.1b1-py3-none-any.whl/pgcli/packages/parseutils.py", line 295, in parse_partial_identifier
    p = sqlparse.parse(word)[0]
[...]

PR #568, opened 2016-08-23
Temporary hack for sqlparse crashing after AS

koljonen:
And a regression test.

@amjith, mind merging this in? 
 #563


amjith:
@koljonen Would you mind adding a comment saying this is a hack and should be removed after it is fixed in sqlparse?

Otherwise it looks good to be merged.

amjith:
Nice! :peach:

PR #571, opened 2016-08-25
WIP: Use tweaked version of psycopg2.extras.wait_select to supply a timeout

darikg:
See psycopg/psycopg2#468 for a minimal example of the problem

### DIFF
--- a/pgcli/pgexecute.py
+++ b/pgcli/pgexecute.py
@@ -1,14 +1,14 @@
 import traceback
 import logging
-import itertools
 import psycopg2
 import psycopg2.extras
 import psycopg2.extensions as ext
 import sqlparse
 import pgspecial as special
+import select
+from psycopg2.extensions import POLL_OK, POLL_READ, POLL_WRITE
 from .packages.parseutils.meta import FunctionMetadata, ForeignKey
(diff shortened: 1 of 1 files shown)


amjith:
Thanks for tracking that down. I think we should add this in as a temporary patch while it is fixed in psycopg2. 

Looks like Daniele Varrazzo (psycopg2 maintainer) is open to having that timeout added to psycopg2 itself. I wouldn't hurt to send him a PR with the fix. That could move things along faster. 

Is this ready for merge or do you want to wait a bit longer?

darikg:
You're right, I should submit a PR to psycopg2.

I haven't had much of a chance to stress test this PR -- not sure how it interacts with long running queries. I also wasn't sure if we should hardcode the timeout or make that another pgclirc option or whatever.

amjith:
Is there value in having this value configurable? Will users ever have to modify this value to get it working on their system? 

Regarding stress tests. I ran `select pg_sleep(30)` and it succeeded without disconnecting. But I don't know if that's the kind of stress test you had in mind.

darikg:
Yeah, that's all I meant really. Feel free to merge!

amjith:
Awesome! ✈️

PR #572, opened 2016-08-25
More prompt

rjuju:
Hello,

first patch is a bugfix (I'm using python2):

```
$ python -m pgcli.main -d "dbname=rjuju"
[...]
  File "/.../pgcli/main.py", line 682, in get_prompt
    string = string.replace('\\u', self.pgexecute.user or '(none)')
TypeError: coercing to Unicode: need string or buffer, tuple found
```

Second one is to add more placeholder to the new prompt variable:
  * \p for server port
  * \i for backend PID
  * \# for superuser status (same as the %# placeholder in psql)

That's mostly the placeholder I need for my personal use, so feel free to merge only one or no one of these patch.

PS: I didn't use the same placeholder as psql since existing one were already different (except for \#, didn't find better idea). If the one I choose are badly chosen, feel free to change them.

### DIFF
--- a/pgcli/pgexecute.py
+++ b/pgcli/pgexecute.py
@@ -136,6 +136,9 @@ class PGExecute(object):
         host = (host or self.host)
         port = (port or self.port)
         dsn = (dsn or self.dsn)
+        pid = -1
+        superuser = False
+        print("dsn: %s" % dsn)
         if dsn:
             if password:
                 dsn = "{0} password={1}".format(dsn, password)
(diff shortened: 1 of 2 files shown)


pgcli/pgexecute.py
```diff
@@ -136,24 +136,32 @@ def connect(self, database=None, user=None, password=None, host=None,
         host = (host or self.host)
         port = (port or self.port)
         dsn = (dsn or self.dsn)
+        pid = -1
+        superuser = False
+        print("dsn: %s" % dsn)
```
amjith: Is this print necessary?

rjuju:
Oups, sorry I forgot to remove it before committing, last commit fixes this.

amjith:
I like the additions. I'm glad you decided to use some sane specifiers such as `\p` instead of using psql's `%>` for port. Good choice. 

Thanks for the quick turn around on addressing the comments too. 🍨

rjuju:
Thanks, and thanks a lot for this nice tool!

PR #573, opened 2016-08-29
do not install setproctitle on cygwin

ysangkok:


amjith:
Can you explain why this is needed?

I'm under the impression that cygwin supplies with a `ps` command and if you run pgcli under cygwin and run `ps` you can see the password listed in the process list right?

ysangkok:
Setproctitle doesnt work on cygwin due to at compilation error. I didn't test if the password is visible, but I consider it a hack anyway, the password should never have been in argv. Since it isn't a hard dependency, isn't it better to at least have pgcli working?

amjith:
Thanks!

🚀

PR #577, opened 2016-09-04
Make semicolon not trigger execution in multiline mode

koljonen:
Basically, having to go to the end of the input to delete the semicolon, then back to wherever you want to insert your newline, and then back to the end to re-add the semicolon before executing your edited query is way too complicated, and I keep getting it wrong, resulting in half-written queries being executed and me getting frustrated. 
My suggested solution is to (1) remove the power of the semicolon to decide what [Enter] does, and (2) make the UI (toolbar) tell the user about [Esc] [Enter].

The reason it's [Esc] [Enter] and not [Alt]-[Enter] in the toolbar is that [Alt]-[Enter] doesn't work by default on Mac OS (it requires the user to change the settings of the terminal app).

So ... does anyone else use multiline mode and have an opinion on this?

### DIFF
--- a/pgcli/pgbuffer.py
+++ b/pgcli/pgbuffer.py
@@ -4,13 +4,20 @@ from .packages.parseutils.utils import is_open_quote
 
 
 class PGBuffer(Buffer):
-    def __init__(self, always_multiline, *args, **kwargs):
+    def __init__(self, always_multiline, multiline_mode, *args, **kwargs):
         self.always_multiline = always_multiline
+        self.multiline_mode = multiline_mode
 
         @Condition
         def is_multiline():
             doc = self.document
-            return self.always_multiline and not _multiline_exception(doc.text)
+            if not self.always_multiline:
+                return False
+            if self.multiline_mode == 'safe':
+                return True
+            else:
+                return not _multiline_exception(doc.text)
+
 
         super(self.__class__, self).__init__(*args, is_multiline=is_multiline,
                                              tempfile_suffix='.sql', **kwargs)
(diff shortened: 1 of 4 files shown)


amjith:
I think a lot of people rely on multi-line mode to emulate the behavior of psql where it doesn't execute the query unless a semi-colon is present at the end of the line. This will break compatibility for those folks.

I'm open to the idea of adding alt-enter or esc-enter as an additional mode and call it safe-edit mode or something else.

koljonen:
Ah, so that's why it works the way it does! I've replaced my previous commit with one where it's controlled by an config setting.

pgcli/pgbuffer.py
```diff
@@ -4,13 +4,20 @@
...
-    def __init__(self, always_multiline, *args, **kwargs):
+    def __init__(self, always_multiline, multiline_mode, *args, **kwargs):
         self.always_multiline = always_multiline
+        self.multiline_mode = multiline_mode
 
         @Condition
         def is_multiline():
             doc = self.document
-            return self.always_multiline and not _multiline_exception(doc.text)
+            if not always_multiline:
```
amjith: This has to be `if not self.always_multiline`. That makes sure we're checking the instance variable instead of the local variable. We modify the instance variable when the user presses F3 to toggle the multi-line.

amjith:
Except for that tiny bug it looks good. 👍 

Ping me once you get a chance to push an update and I'll review and merge.

koljonen:
Good catch, @amjith!

amjith:
🍫

PR #582, opened 2016-09-13
XDG_CONFIG_HOME support for config file location

amigrave:


amjith:
@amigrave Thank you for the PR. Can you tell me what you use for your XDG_CONFIG_HOME? 

One of the reasons I'm reluctant to add this change is because it make support issues harder to resolve. There is no reliable way for me to say change this in your config file which is located at ~/.config/pgcli/config or look at the log file under ~/.config/pgcli/log etc. 

Also this change will break existing installations if the user has a different XDG_CONFIG_HOME than the `~/.config`, because previous versions of pgcli would have written it to ~/.config/pgcli/config and the user might have edited it to their desire. Now pgcli will try to read the config file from $XDG_CONFIG_HOME/pgcli/config and it will be empty thus reverting to defaults.

I'm not saying I won't merge this, but I'd like to hear more thoughts before I make a decision.

amigrave:
@amjith Hi Amjith and thanks for that fine piece of software !

I can reassure you that this PR won't break other people's config because the default value of XDG_CONFIG_HOME is actually "$HOME/.config" as defined in the specification:
https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html

So it won't change for anyone who does not configure it's `XDG_CONFIG_HOME` variable (eg: 99% of people I guess).
The people who actually change the value of `XDG_CONFIG_HOME` does that for a particular purpose and they are aware of the consequence (they just want their configuration files to be stored elsewhere than in `~/.config`) whether they are running Linux or OSX.

So if you still haven't got this PR before, it's just because none of your users customise the value of `XDG_CONFIG_HOME` (yet)... or they where too lazy to do a PR :-)

amjith:
That's a fair reasoning. Thanks for taking the time to explain as well as send a PR to improve pgcli. 🍧

PR #583, opened 2016-09-15
master: numbers division for int and float

foxyterkel:
I apologize in advance for possible mistakes, first time trying to take part in open source project.  All decimal columns defined as text field, I change this behavior. Now decimal defined as float. Changed default formatting for float and int. And separate int and text cases.

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -137,7 +137,8 @@ class PGCli(object):
         self.null_string = c['main'].get('null_string', '<null>')
         self.prompt_format = c['main'].get('prompt', self.default_prompt)
         self.on_error = c['main']['on_error'].upper()
-
+        self.decimal_format = c['data_formats']['decimal']
+        self.float_format = c['data_formats']['float']
         self.completion_refresher = CompletionRefresher()
 
         self.query_history = []
@@ -586,8 +587,8 @@ class PGCli(object):
                 max_width = None
 
             formatted = format_output(
-                title, cur, headers, status, self.table_format, self.null_string,
-                self.pgspecial.expanded_output, max_width)
+                title, cur, headers, status, self.table_format, self.decimal_format,
+                self.float_format, self.null_string,  self.pgspecial.expanded_output, max_width)
 
             output.extend(formatted)
             total = time() - start
@@ -791,7 +792,9 @@ def obfuscate_process_password():
 
     setproctitle.setproctitle(process_title)
 
-def format_output(title, cur, headers, status, table_format, missingval='<null>', expanded=False, max_width=None):
+
+def format_output(title, cur, headers, status, table_format, dcmlfmt, floatfmt,
+                  missingval='<null>', expanded=False, max_width=None):
     output = []
     if title:  # Only print the title if it's not None.
         output.append(title)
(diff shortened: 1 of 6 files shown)


amjith:
@foxyterkel I'll take a look at this PR over this weekend. I'm slammed at the moment. 

@dbcli/pgcli-core If you folks have time, please take this for a spin and provide your feedback.

j-bennet:
Unit tests fail on python 2.6 that don't have format specifiers like `,d` or `,g`. Do we still care about python 2.6, or can we remove it from our test suite since it's deprecated?

foxyterkel:
I'm using the same logic in the task so if we need to support python 2.6 my updates need to be refactored in format method

amjith:
The changes look good to me. It works as expected. 

I have two requests. 

1. Can we make this behavior optional via a config value? (Have this enabled by default).
2. How hard is it to make it python 2.6 compatible? 


Comment:
If you have to spend more than 30 mins to make it 2.6 compatible, don't bother. We will drop support for 2.6.

foxyterkel:
1. I think we can make some sort of config file. But i don't see conf files in project, so it will be only for this formatting, or we need to create conf file with some structure?
2. I'll spend some time on weekend to look for some old fashion formatting style which well be compatible with 2.6.

j-bennet:
@foxyterkel pgcli reads settings from pgclirc file. Defaults are here:

https://github.com/dbcli/pgcli/blob/fa9e01cd858da6c2087ddd8e0c3230de704109f8/pgcli/pgclirc

Here is the code that reads config file and sets pgcli properties accordingly:

https://github.com/dbcli/pgcli/blob/fa9e01cd858da6c2087ddd8e0c3230de704109f8/pgcli/main.py#L113-L139

I would suggest adding a new config section. Something like:

```
[data_formats]
int = "d"
float = ""
```

(just an example).

Also, in 2.6 ",g" does not work, but "g" does. See here:

In 2.6:

```
(pgcli26) irina@irina-macbook ~> bpython
/Users/irina/.venvs/pgcli26/lib/python2.6/site-packages/cryptography/__init__.py:26: DeprecationWarning: Python 2.6 is no longer supported by the Python core team, please upgrade your Python. A future version of cryptography will drop support for Python 2.6
  DeprecationWarning
bpython version 0.15 on top of Python 2.6.9 /Users/irina/.venvs/pgcli26/bin/python2.6
>>> format(float(100000000), 'g')
'1e+08'
```

In 2.7:

```
(pgcli) irina@irina-macbook ~> bpython
bpython version 0.15 on top of Python 2.7.11 /Users/irina/anaconda/bin/python
>>> format(float(100000000), 'g')
'1e+08'
```

foxyterkel:
As for thousand separator on 2.6. I don't see easy way to make it compatible. In format specification ",g", exactly "," makes local independent thousand separator. It was added from 2.7.

amjith:
@foxyterkel Are you still working on this? 

If you're too busy, I'm happy to merge this in and implement the config file change myself. Let me know.

foxyterkel:
@amjith I'm working on it in me free time. Last days i have a lot of work but still i'm going to finish it, when i'll have free minutes. But if you wanna finish it faster, you can fell free.

amjith:
There is absolutely no hurry. I wasn't trying to rush you. I just wanted to know if you needed some help to pick it up if you were feeling uninterested.

foxyterkel:
Decimal and float format now taking from config file. But problem with python 2.6 still exists.

amjith:
Changes look good. I'm going to merge this in and tackle Python 2.6 in a separate PR.

Thank you for the contribution. 🌺

PR #589, opened 2016-09-21
fixes #588 invalid setting for redshift

### LINKED ISSUE #588: pgcli v1.2.0 not working for redshift
Introduced by https://github.com/dbcli/pgcli/commit/3f310f5070d45852afda29ed50ba656ff1cb7465#diff-8859be39408e4fbeea82b923784d9e65R163

```
> select current_setting('is_superuser')::bool
cannot cast type text to boolean
```

Leaving away the cast

```
> select current_setting('is_superuser')
must be superuser to examine "is_superuser"
```

Which I'm guessing means I'm not a superuser :-)

mbarkhau:


mbarkhau:
I wonder if amazon would be willing

mbarkhau:
Closed and reopend to trigger flaky travis build again.

amjith:
I will try and review this later today. Sorry things are a little bit tight
on the homefront.

pgcli/pgexecute.py
```diff
@@ -186,7 +185,9 @@ def connect(self, database=None, user=None, password=None, host=None,
 
             cursor = conn.cursor()
 
-        superuser = self._select_one(cursor, "select current_setting('is_superuser')::bool")[0]
+        cursor.execute("SHOW ALL")
```
amjith: This seems such a broad stroke to check whether it is a super user or not. Are you sure there isn't a simpler way to do this?

Does `select current_setting('is_superuser')` not work in redshift? I thought the only issue in redshift is the typecasting of that value to bool.

mbarkhau: My thinking was, if we're going to get the database settings, we might as well just get them all and get it over with, in the future maybe somebody wants to examine some other setting and it'll already be there.

Yes it does actually throw the error `must be superuser to examine "is_superuser"` on redshift, which unfortunately means that the current version on pypi is broken for redshift users.

mbarkhau: There might be a simpler way. For example, if we're on redshift, just assume is_superuser is False.

amjith: Not a big fan assuming that super user is false for redshift. 

I think it makes sense to get all of the config vars if we decide to use it in the future. I'm satisfied. 👍

amjith:
I've tested this on a non-redshift db and it works fine. 

Thanks for the PR. 🚀

Issue #590, opened 2016-10-02
Formatting error with string-typed numbers


darikg:
This seems to be an issue with the new comma formatting cc @foxyterkel 

```
dg@z.db:zdb_110z> select '123'
Unknown format code 'd' for object of type 'unicode'
```

fpietka:
@amjith isn't it fixed now? seems the fix got merged.

j-bennet:
You're right - this is not a problem anymore. Closing.

PR #591, opened 2016-10-02
Minimal fix for #590

darikg:
issue #590


darikg:
This probably isn't the correct thing to do -- numeric formatting shouldn't apply to string types, even if they look like numbers. But it looks like tabulate is happy to treat strings as numbers if they look like numbers

foxyterkel:
I thought about definition of value type. Is it right to define '1234' like int? Could be better to make some fixes in _column_type method? But @darikg  decision will work to.

darikg:
@foxyterkel agreed. I updated the PR

amjith:
Thanks @darikg!

🔨

Issue #595, opened 2016-10-14
Problem with Python 3.6.0b2


lelit:
Hi all,

this is just an alert, because I think the culprit is Python, not pgcli. Nevertheless, digging into the problem I found what seems an _innocent_ glitch, read below.

I quote here a [message](https://mail.python.org/pipermail/python-list/2016-October/715462.html) I just sent to the Python user list:

> Hi all,
> 
> trying out pgcli with Python 3.6.0b2 I got an error related to what seem a
> different behaviour, or even a bug, of re.sub().
> 
> The original intent is to replace spaces within a string with the regular expression
> \s+ (see https://github.com/dbcli/pgcli/blob/master/pgcli/packages/prioritization.py#L11,
> ignore the fact that the re.sub() call seem underoptimal).
> 
> With Python 3.5.2 is straightforward:
> 
>   $ python3.5
>   Python 3.5.2+ (default, Sep 22 2016, 12:18:14) 
>   [GCC 6.2.0 20160927] on linux
>   Type "help", "copyright", "credits" or "license" for more information.
>   >>> import re
>   >>> re.sub(r'\s+', r'\s+', 'foo bar')
>   'foo\\s+bar'
> 
> While Python 3.6.0b2 gives:
> 
>   $ python3.6
>   Python 3.6.0b2+ (default, Oct 11 2016, 08:30:05) 
>   [GCC 6.2.0 20160927] on linux
>   Type "help", "copyright", "credits" or "license" for more information.
>   >>> import re
>   >>> re.sub(r'\s+', r'\s+', 'foo bar')
>   Traceback (most recent call last):
>     File "/usr/local/python3.6/lib/python3.6/sre_parse.py", line 945, in parse_template
>       this = chr(ESCAPES[this][1])
>   KeyError: '\\s'
> 
>   During handling of the above exception, another exception occurred:
> 
>   Traceback (most recent call last):
>     File "<stdin>", line 1, in <module>
>     File "/usr/local/python3.6/lib/python3.6/re.py", line 191, in sub
>       return _compile(pattern, flags).sub(repl, string, count)
>     File "/usr/local/python3.6/lib/python3.6/re.py", line 326, in _subx
>       template = _compile_repl(template, pattern)
>     File "/usr/local/python3.6/lib/python3.6/re.py", line 317, in _compile_repl
>       return sre_parse.parse_template(repl, pattern)
>     File "/usr/local/python3.6/lib/python3.6/sre_parse.py", line 948, in parse_template
[...]

PR #596, opened 2016-10-15
Fix for issue #595

lelit:
All behaviors and tests pass, with Python 3.6b2. I did not exercise them on previous versions, but I can surely do it, if needed.


lelit:
The failures with Python 2.6 do not seem related with my change. Please let me know if I should investigate further on them.

amjith:
Yes the Python 2.6 failure is expected. I need to either fix that or ignore that test or drop 2.6 support. 

I just haven't found the time this past few weeks to do one of those. 

Thank you for taking the time to dig in and fix the issue. I appreciate it. It makes sense that the raw string qualifiers are enforced in Python 3.6 since `f-strings` are being introduced.

Issue #605, opened 2016-10-26
Completion raises an error with sqlparse 0.2.2


lelit:
Recently released version 0.2.2 of sqlparse [changed](https://github.com/andialbrecht/sqlparse/blob/0.2.2/CHANGELOG#L19) two methods of the `Token` class into properties, in particular `is_group()`, used [here](https://github.com/dbcli/pgcli/blob/master/pgcli/packages/parseutils/tables.py#L17), so the following happens:

```
user@testdb> select * from Exception in thread Thread-14:
Traceback (most recent call last):
  File ".../python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File ".../python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "/tmp/e/lib/python3.6/site-packages/prompt_toolkit/interface.py", line 841, in run
    completions = list(buffer.completer.get_completions(document, complete_event))
  File ".../pgcli/pgcli/pgcompleter.py", line 358, in get_completions
    suggestions = suggest_type(document.text, document.text_before_cursor)
  File ".../pgcli/pgcli/packages/sqlcompletion.py", line 152, in suggest_type
    return suggest_based_on_last_token(stmt.last_token, stmt)
  File ".../pgcli/pgcli/packages/sqlcompletion.py", line 366, in suggest_based_on_last_token
    tables = extract_tables(stmt.text_before_cursor)
  File ".../pgcli/pgcli/packages/parseutils/tables.py", line 147, in extract_tables
    return tuple(i for i in identifiers if i.name)
  File ".../pgcli/pgcli/packages/parseutils/tables.py", line 147, in <genexpr>
    return tuple(i for i in identifiers if i.name)
  File ".../pgcli/pgcli/packages/parseutils/tables.py", line 96, in extract_table_identifiers
    for item in token_stream:
  File ".../pgcli/pgcli/packages/parseutils/tables.py", line 34, in extract_from_part
    if is_subselect(item):
  File ".../pgcli/pgcli/packages/parseutils/tables.py", line 17, in is_subselect
    if not parsed.is_group():
TypeError: 'bool' object is not callable
```

darikg:
Thanks for the heads up. I can fix that but it looks like a couple more things broke... Looking into it now

j-bennet:
Fixed in 1.3.1.

PR #606, opened 2016-10-26
Bump sqlparse to 0.2.2

darikg:
See #605


amjith:
Looks like I released a broken version to PyPI 😢. 

Thanks for the quick fix for the broken build @darikg. It works great. 🚠

Issue #603, opened 2016-10-22
\i raises a TypeError


lelit:
With Python 3.6b2, I got the following immediately after entering `\i` followed by a space:

```
postgres@mydb> \i Exception in thread Thread-4:                            
Traceback (most recent call last):                                              
  File "/usr/local/lib/python3.6/threading.py", line 916, in _bootstrap_inner   
    self.run()                                                                  
  File "/usr/local/lib/python3.6/threading.py", line 864, in run                
    self._target(*self._args, **self._kwargs)                                   
  File "/usr/local/lib/python3.6/site-packages/prompt_toolkit/interface.py", line 841, in run
    completions = list(buffer.completer.get_completions(document, complete_event))                                                                                                                                 
  File "/usr/local/lib/python3.6/site-packages/pgcli/pgcompleter.py", line 371, in get_completions
    reverse=True)                                                               
TypeError: '<' not supported between instances of 'NoneType' and 'NoneType'     
```

I cannot investigate further right now, but I will try to narrow the case.

j-bennet:
Hmm, I can't even get this far. For me, running in 3.6 breaks when importing sqlparse.

```
Traceback (most recent call last):
  File "/Users/irina/.venvs/pgcli36/bin/pgcli", line 11, in <module>
    load_entry_point('pgcli', 'console_scripts', 'pgcli')()
  File "/Users/irina/.venvs/pgcli36/lib/python3.6/site-packages/pkg_resources/__init__.py", line 564, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/Users/irina/.venvs/pgcli36/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2608, in load_entry_point
    return ep.load()
  File "/Users/irina/.venvs/pgcli36/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2268, in load
    return self.resolve()
  File "/Users/irina/.venvs/pgcli36/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2274, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/Users/irina/src/pgcli/pgcli/main.py", line 38, in <module>
    from pgspecial.main import (PGSpecial, NO_QUERY, content_exceeds_width)
  File "/Users/irina/.venvs/pgcli36/lib/python3.6/site-packages/pgspecial/__init__.py", line 12, in <module>
    from . import iocommands
[...]

PR #608, opened 2016-10-31
Fix issue #603

lelit:
Python 3 forbids comparisons between different types: use a tuple
containing a single 0 (zero) as the priority for path matches so that it
can be compared with those generated by the workhorse method find_matches().


j-bennet:
Looks good. Merging. And 👍  for the unit test.

PR #614, opened 2016-11-20
Fix scoping for columns from CTEs

koljonen:
We were suggesting all CTE columns regardless of whether the CTE in
question was included in the FROM clause.

This also makes populate_scoped_columns output `TableRefence` namedtuples for CTEs, like it does for tables/views/functions. Previously it would emit strings for CTEs.

### DIFF
--- a/pgcli/pgcompleter.py
+++ b/pgcli/pgcompleter.py
@@ -689,7 +689,7 @@ class PGCompleter(Completer):
         :param local_tbls: tuple(TableMetadata)
         :return: {TableReference:{colname:ColumnMetaData}}
         """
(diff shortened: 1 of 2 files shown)


j-bennet:
@koljonen what's the benefit of returning tuples vs strings for CTEs?

koljonen:
Mostly it's that I think it's confusing for a function's return type to vary like that. I actually found the scoping problem while investigating a crash caused by getting a string where I expected a `TableReference` (while developing something unrelated to CTEs).

j-bennet:
👍 
Merging.

PR #619, opened 2016-12-05
Table-qualifying column suggestions

koljonen:
@darikg Do you have time to look at yet another of my tweaks to the completion engine?

> Support for table-qualifying column suggestions
> 
> ... i.e. suggesting foo.fooid instead of just fooid
> Controlled using a config-file setting:
> **qualify_columns**: always/never/**if_more_than_one_table**.
> 
> To enable the proper sorting of qualified column suggestions, we
> introduce the concept of synonyms for suggestions
> (in `pgcompleter.find_matches`). They way synonyms work is that a
> number of synonyms may be provided for a suggestion sent to
> `find_matches`. If synonyms are provided, sorting is based on how
> well the best synonym matches the input, instead of only comparing
> the input to the suggestion text.
> In this case, the unqualified name acts as a synonym.
> I have a couple of other ideas of use cases where we can use synonyms
> to get better completions with less typing for the user, which I intend
> to explore later.

### DIFF
--- a/pgcli/pgcompleter.py
+++ b/pgcli/pgcompleter.py
@@ -1,7 +1,7 @@
 from __future__ import print_function, unicode_literals
 import logging
 import re
-from itertools import count, repeat, chain
+from itertools import count, repeat
 import operator
 from collections import namedtuple, defaultdict
 from pgspecial.namedqueries import NamedQueries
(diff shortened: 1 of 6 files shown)


darikg:
Yes, will do!

pgcli/pgcompleter.py
```diff
@@ -300,21 +307,18 @@ def _match(item):
...
         matches = []
+        for cand in collection:
+            if isinstance(cand, _Candidate):
+                item, prio, display_meta, synonyms = cand
+            else:
+                item, display_meta, prio, synonyms = cand, meta, 0, [cand]
 
-        for item, meta, prio in collection:
-            sort_key = _match(item)
+            sort_key = max(_match(x) for x in synonyms)
```
darikg: I think the `sort_key` assigment could be moved into the `if isinstance` block to avoid a lot of unnecessary list allocations:

```
if isinstance(cand, _Candidate):
    item, prio, display_meta, synonyms = cand
    sort_key = max(_match(x) for x in synonyms)
else:
    item, display_meta, prio = cand, meta, 0
    sort_key = _match(cand)
````

darikg:
This is great! Very ergonomic to use and the cleanup of the find_matches signature is a relief. I just left one minor suggestion inline

PR #621, opened 2016-12-07
Add transaction status to toolbar

koljonen:
This adds a red "Failed transaction" text to the toolbar if the the current transaction is doomed, and a green "Transaction" text if we're in a non-doomed transaction. 

I'd especially welcome any improvement on the choice of colours/styling, as I feel it's not quite what I wanted (I'd want the text to "pop" more), and I suspect we'll be stuck with whatever initial choice we make.

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -497,7 +497,9 @@ class PGCli(object):
             return [(Token.Continuation, '.' * (width - 1) + ' ')]
 
         get_toolbar_tokens = create_toolbar_tokens_func(
-            lambda: self.vi_mode, self.completion_refresher.is_refreshing)
+            lambda: self.vi_mode, self.completion_refresher.is_refreshing,
+            self.pgexecute.failed_transaction,
+            self.pgexecute.valid_transaction)
 
         layout = create_prompt_layout(
             lexer=PygmentsLexer(PostgresLexer),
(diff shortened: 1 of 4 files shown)


j-bennet:
@koljonen How do you reproduce those status messages?

amjith:
I like it. 😄 

Here's how I reproduced it: 

```
amjith@(none):expenses> begin
BEGIN
Time: 0.000s
amjith@(none):expenses> drop table joint_checking;
cannot drop table joint_checking because other objects depend on it
DETAIL:  view all_accounts depends on table joint_checking
HINT:  Use DROP ... CASCADE to drop the dependent objects too.

Time: 0.118s

```

1__0_0_python2_7_-__pgcli___users_amjith_dropbox_code_python_pgcli____tmux_

This is what I see in the status bar.

I like this feature quite a bit. 

Regarding the colors if you'd like to get the colors to pop, try changing the background color to something other than black. You can even make it blinking (which I'm sure people will hate).

j-bennet:
Goodness me, please no blinking. :)

j-bennet:
Suggestion:

```
Token.Toolbar.Transaction.Valid = "bg:#222222 #00ff5f bold"
Token.Toolbar.Transaction.Failed = "bg:#222222 #ff005f bold"
```

Failed:

pgcli-transaction-failed

Good:

pgcli-transaction

koljonen:
Thanks, @j-bennet! I knew someone would be better than me at finding good colours.

j-bennet:
No problem, very nice PR! Verified that it's working, merging.

PR #630, opened 2017-01-06
Set default data_formatting to nothing.

### LINKED ISSUE #629: The data-formats should default to none
Right now the data-formats in the config file default to `,g` and `,d` for float and int. This makes it hard to copy and paste.

We should default this to none.

amjith:
Fixes #629 

Reviewer: @stuartquin


stuartquin:
:+1:

PR #631, opened 2017-01-11
Changelog and version update for release v1.4.0.

j-bennet:
@dbcli/pgcli-core Please review.


amjith:
The changelog looks good. 

I noticed that you've bumped the version number in `__init__.py`. The `release.py` scripts won't work correctly if the version is already bumped. The way it is written it will need the __init__.py file updated but uncommited. So the script will commit that file with the appropriate commit message.

j-bennet:
Thanks for the tip @amjith, I'll keep that in mind.

Issue #298, opened 2015-07-26
Make Exanded output a config option


amjith:
Provide a config option to always enable expanded output.

moll:
Just copying my Gitter reply for the record:
> The the good thing about .psqlrc that it can contain pretty much anything you can do in the REPL.

amjith:
@danieljwest This is a good issue to get your feet wet. 

In `pgcli` there is a way to toggle expanded mode from the prompt which is `\x`. When this is enabled the output is printed in a horizontal fashion as follows: 

```
amjith> \x
Expanded display is on.
Command Time: 0.000s
Format Time: 0.000s
amjith> SELECT * from information_schema.sql_features
-[ RECORD 0 ]-------------------------
feature_id       | B011
feature_name     | Embedded Ada
sub_feature_id   |
sub_feature_name |
is_supported     | NO
is_verified_by   | None
comments         |
-[ RECORD 1 ]-------------------------
feature_id       | B012
feature_name     | Embedded C
sub_feature_id   |
sub_feature_name |
is_supported     | YES
is_verified_by   | None
comments         |
-[ RECORD 2 ]-------------------------
feature_id       | B013
feature_name     | Embedded COBOL
sub_feature_id   |
sub_feature_name |
is_supported     | NO
is_verified_by   | None
comments         |
```

This feature request is to set this option to be on/off by default from the config file `~/.pgclirc`. 

I'd recommend adding this to https://github.com/dbcli/pgcli/blob/master/pgcli/pgclirc and then checking for this config in main.py and calling `pgspecial.toggle_expanded_mode()`. But this might require  changing the `pgspecial.toggle_expanded_mode()` to `set_expanded_mode(arg)` that takes an argument so you can choose to set it based on the value from config file.

Dangeranger:
Yes. Please do this :+1: 

A `.pgcli` api that can match `.psqlrc` feature wise would be a huge improvement in PgCLI as far as I am concerned.

j-bennet:
Available in 1.5.0.

PR #633, opened 2017-01-21
Per #298, make expand modes configurable.

AGhost-7:
This makes it possible to enable through the configuration the always
expand and automatic expansion modes in pgcli.


j-bennet:
Looks good. Thanks for the PR.

Issue #587, opened 2016-09-21
Switch the shorted prompt when prompt length is too long


amjith:
Most of the complaints we receive about the pgcli prompt is that it is too big when people are using it with AWS hosts. 

It would be nice to automatically switch to database name only when the prompt size is bigger than a certain threshold. Say 30 chars (this will need a bit of experimentation to get right).

j-bennet:
Fixed in 1.5.0.

PR #639, opened 2017-02-08
Switch the shorted prompt when prompt length is too long. 

jnth:
Connect #587 : switch the shorted prompt when prompt length is too long (if using the default prompt).


amjith:
Nice work! I love it. 💯

PR #641, opened 2017-02-25
Suggest keywords after ALTER

darikg:
This is super simple I just wanted `ALTER TABLE foo ALTER ` to suggest `COLUMN`


j-bennet:
Looks good. Merging.

PR #642, opened 2017-02-25
User config to reserve more menu space

darikg:
By default we only reserve 4 lines for the suggestion menu, which feels cramped to me. This just makes the number of lines user-configurable


j-bennet:
👍

darikg:
Thanks @j-bennet  !

PR #647, opened 2017-03-03
Adjust the connection order to work around psycopg2 restriction.

### LINKED ISSUE #646: set_session cannot be used inside a transaction
Just freshly installed `pgcli` and upon trying to connect to my local postgres database I get:

   orf@mean-lean-dev-machine:~|⇒  pgcli -d doge doge -U doge
   set_session cannot be used inside a transaction

Does this mean anything?

amjith:
Reviewer: @darikg 

This fixes #646 

Looks like psycopg2 released a new version yesterday that broke us. 😞 

The change was you can't change the `autocommit` status while a cursor is still active. So I moved the sql execution after we set the `autocommit` to `True`.

### DIFF
--- a/pgcli/pgexecute.py
+++ b/pgcli/pgexecute.py
@@ -185,10 +185,6 @@ class PGExecute(object):
 
             cursor = conn.cursor()
 
-        cursor.execute("SHOW ALL")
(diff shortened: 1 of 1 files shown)


j-bennet:
👍

Issue #650, opened 2017-03-06
Completion for materialized views.


RazerM:
pgcli does not suggest/complete the following:

```
CREATE MATERIALIZED VIEW ...
DROP MATERIALIZED VIEW ...
REFRESH MATERIALIZED VIEW ...
```

j-bennet:
Implemented in #653.

PR #653, opened 2017-03-06
Add MATERIALIZED VIEW keywords

koljonen:
For issue #650.


amjith:
Since we're so close to releasing, can we merge this after the release? I don't feel like this is an urgent fix.

j-bennet:
That was one quick release @amjith .

PR #654, opened 2017-03-07
Casing of column headers (using casing file)

koljonen:
I think it's a lot nicer looking.
```
koljonen@localhost:koljonen> SELECT transferid, parenttransferid, amount from transfers
╒══════════════╤════════════════════╤══════════╕
│ TransferID   │ ParentTransferID   │ Amount   │
╞══════════════╪════════════════════╪══════════╡
...
...
...
```


koljonen:
Popping the dict and checking that it's empty is the cleanest way I could think of.
I've rebased as there was a conflict with master.

amjith:
I think this request from @j-bennet requires a little bit of context. She was looking for a way to do static checking where the IDE (such as PyCharm) could alert you when you mistype the name of a certain variable. When you pass in a dictionary there is no way to do static checking because the keys could be anything.

What you have implemented is a nice way to do runtime checking. This guarantees that the function will crash if called incorrectly. But to me it feels like a python anti-pattern. We shouldn't be mutating a dictionary passed into a function unless absolutely necessary.

So here's a compromise, why not use `namedtuple`? Define a namedtuple called settings or something equally descriptive and pass that in as an argument. This will achieve the desired outcome of doing static checking without having a zillion function arguments.

j-bennet:
Thanks @amjith, you explained it better.

koljonen:
Oh, I see. That makes perfect sense. I've squashed it all into two new commits.

koljonen:
Oops, sorry, I seem to have misplaced some code when rebasing. 😳 
Fixed now.

PR #655, opened 2017-03-07
Parse function bodies

koljonen:
Consider this script
```
CREATE FUNCTION foo() returns text LANGUAGE SQL AS $func$
SELECT 1 FROM Bar;
SELECT <cursor> FROM Baz;
$func$;
```
The change here is that `SELECT <cursor> FROM Baz;` will be seen as the
current statement, instead of the whole function definition.
This means we'll no longer get column suggestions from `Bar`.

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -1,3 +1,10 @@
+Upcoming
+=====
+
+Features:
+---------
+* Better suggestions when editing functions (Thanks: `Joakim Koljonen`_)
+
 1.5.0
 =====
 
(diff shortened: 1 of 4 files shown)


amjith:
I'll review this later tonight.

amjith:
Looks good to me. Thank you! 🚗 

Would you mind updating the changelog with a one line description of this PR? 

I'm trying to see if we can update the changelog as we go instead of waiting to update it right before release.

koljonen:
I've rebased it and amended the commit with an update to the change.

amjith:
Awesome!! Thank you!

PR #659, opened 2017-03-10
Add a command line option for --less-chatty. Fix #626.

### LINKED ISSUE #626: Add a command line option for --less-chatty
This should suppress the startup message when supplied.

kanet77:
I saw this issue labeled Easy. I took a stab at it and I'm open to feedback.

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -137,11 +137,7 @@ class PGCli(object):
         self.syntax_style = c['main']['syntax_style']
         self.cli_style = c['colors']
         self.wider_completion_menu = c['main'].as_bool('wider_completion_menu')
-        if less_chatty is not None:
-            self.less_chatty = less_chatty
-        else:
-            self.less_chatty = c['main'].as_bool('less_chatty') or False
-
+        self.less_chatty = bool(less_chatty) or c['main'].as_bool('less_chatty')
         self.null_string = c['main'].get('null_string', '<null>')
         self.prompt_format = prompt if prompt is not None else c['main'].get('prompt', self.default_prompt)
         self.on_error = c['main']['on_error'].upper()
@@ -735,7 +731,7 @@ class PGCli(object):
 @click.option('-R', '--row-limit', default=None, envvar='PGROWLIMIT', type=click.INT,
         help='Set threshold for row limit prompt. Use 0 to disable prompt.')
 @click.option('--less-chatty', 'less_chatty', is_flag=True,
-        default=None,
+        default=False,
         help='Skip intro on startup and goodbye on exit.')
 @click.option('--prompt', help='Prompt format (Default: "\\u@\\h:\\d> ").')
 @click.argument('database', default=lambda: None, envvar='PGDATABASE', nargs=1)
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -105,7 +105,7 @@ class PGCli(object):
 
     def __init__(self, force_passwd_prompt=False, never_passwd_prompt=False,
                  pgexecute=None, pgclirc_file=None, row_limit=None,
-                 single_connection=False, prompt=None):
+                 single_connection=False, less_chatty=None, prompt=None):
 
         self.force_passwd_prompt = force_passwd_prompt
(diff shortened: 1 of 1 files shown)


pgcli/main.py
```diff
@@ -137,7 +137,11 @@ def __init__(self, force_passwd_prompt=False, never_passwd_prompt=False,
         self.syntax_style = c['main']['syntax_style']
         self.cli_style = c['colors']
         self.wider_completion_menu = c['main'].as_bool('wider_completion_menu')
-        self.less_chatty = c['main'].as_bool('less_chatty')
+        if less_chatty is not None:
+            self.less_chatty = less_chatty
+        else:
+            self.less_chatty = c['main'].as_bool('less_chatty') or False
```
kanet77: This line might not be intuitive enough because it relies on Python behavior of boolean logic with `None`. (See ["False or None vs. None or False"](http://stackoverflow.com/questions/3914667/false-or-none-vs-none-or-false).)

The best alternative I see is a nested `if/else` but that's verbose compared with how other settings are handled here. I was disappointed `as_bool` doesn't take a default.

I went to all this trouble in order to handle when `less_chatty` is `True`, `False`, or absent from the config file.

It's only possible to turn chattiness _off_ with the command line option. On reflection, perhaps it makes sense for `--less-chatty` to take a true/false value.

pgcli/main.py
```diff
@@ -729,11 +734,15 @@ def get_prompt(self, string):
         help='Use DSN configured into the [alias_dsn] section of pgclirc file.')
 @click.option('-R', '--row-limit', default=None, envvar='PGROWLIMIT', type=click.INT,
         help='Set threshold for row limit prompt. Use 0 to disable prompt.')
+@click.option('--less-chatty', 'less_chatty', is_flag=True,
+        default=None,
+        help='Skip intro on startup and goodbye on exit.')
```
kanet77: Wording taken verbatim from the default pgclirc file.

One alternative is "Suppress the startup message", from issue #626, but I went with the more verbose option.

kanet77:
# Manual testing

## --less-chatty listed in --help

```
$ pgcli --help
Usage: pgcli [OPTIONS] [DATABASE] [USERNAME]

Options:
  -h, --host TEXT          Host address of the postgres database.
  -p, --port INTEGER       Port number at which the postgres instance is
                           listening.
  -U, --user TEXT          User name to connect to the postgres database.
  -W, --password           Force password prompt.
  -w, --no-password        Never prompt for password.
  --single-connection      Do not use a separate connection for completions.
  -v, --version            Version of pgcli.
  -d, --dbname TEXT        database name to connect to.
  --pgclirc TEXT           Location of pgclirc file.
  -D, --dsn TEXT           Use DSN configured into the [alias_dsn] section of
                           pgclirc file.
  -R, --row-limit INTEGER  Set threshold for row limit prompt. Use 0 to
                           disable prompt.
  --less-chatty            Skip intro on startup and goodbye on exit.
  --prompt TEXT            Prompt format (Default: "\u@\h:\d> ").
  --help                   Show this message and exit.
```

## less_chatty is not in config

```
$ pgcli
Version: 1.5.0
Home: http://pgcli.com
tk@(none):tk>
Goodbye!

$ pgcli --less-chatty
tk@(none):tk>
```

## less_chatty = False in config

```
$ pgcli
Version: 1.5.0
Home: http://pgcli.com
tk@(none):tk>
Goodbye!

$ pgcli --less-chatty
tk@(none):tk>
```

## less_chatty = True in config

```
$ pgcli
tk@(none):tk>

$ pgcli --less-chatty
tk@(none):tk>
```

pgcli/main.py
j-bennet: What about `self.less_chatty = bool(less_chatty) or c['main'].as_bool('less_chatty')`

pgcli/main.py
j-bennet: Why set `default` to `None` and not `False`?

pgcli/main.py
kanet77: Much better.

j-bennet:
Well done. Thanks!
🌷

Issue #664, opened 2017-03-13
set_session cannot be used inside a transaction in v1.5.0


jwhitmarsh:
I've seen this raised and closed in #646, #652 and #658  but i'm experiencing this in v1.5.0 when using a service (which was working in 1.3.1):
`pgcli service=foo_db`

I've tried `pip install psycopg2==2.6.2` and this is returned: 
`Requirement already satisfied: psycopg2==2.6.2 in /usr/local/lib/python2.7/site-packages`

j-bennet:
What else does your `pip freeze` show?

jwhitmarsh:
```
altgraph==0.12
awscli==1.10.13
botocore==1.4.4
click==6.6
colorama==0.3.3
configobj==5.0.6
Django==1.8.2
docutils==0.12
futures==3.0.5
git-sweep==0.1.1
gitdb2==2.0.0
GitPython==2.1.0
Glances==2.7.1
humanize==0.5.1
jmespath==0.9.0
macholib==1.7
modulegraph==0.12.1
pgcli==1.2.0
pgspecial==1.6.0
prompt-toolkit==1.0.7
psutil==4.4.0
psycopg2==2.6.2
py2app==0.9
pyasn1==0.1.9
pyfiglet==0.7.5
Pygments==2.1.3
pyobjc==3.0.4
pyobjc-core==3.0.4
pyobjc-framework-Accounts==3.0.4
pyobjc-framework-AddressBook==3.0.4
pyobjc-framework-AppleScriptKit==3.0.4
pyobjc-framework-AppleScriptObjC==3.0.4
pyobjc-framework-Automator==3.0.4
pyobjc-framework-CalendarStore==3.0.4
pyobjc-framework-CFNetwork==3.0.4
pyobjc-framework-Cocoa==3.0.4
pyobjc-framework-Collaboration==3.0.4
pyobjc-framework-CoreData==3.0.4
pyobjc-framework-CoreLocation==3.0.4
pyobjc-framework-CoreText==3.0.4
pyobjc-framework-CoreWLAN==3.0.4
pyobjc-framework-DictionaryServices==3.0.4
pyobjc-framework-DiskArbitration==3.0.4
pyobjc-framework-EventKit==3.0.4
pyobjc-framework-ExceptionHandling==3.0.4
pyobjc-framework-FSEvents==3.0.4
pyobjc-framework-InputMethodKit==3.0.4
pyobjc-framework-InstallerPlugins==3.0.4
[...]

Issue #657, opened 2017-03-08
port shouldn't be required for DSN's in alias_dsn section


msztolcman:
Currently dsn in form:

    somename = postgresql://postgres:password@host/eden

is invalid, after running pgcli `pgcli -D somename` it prints:

    invalid port number: "None"

It should be required (wasn't in previous versions). I didn't verified other fields.

j-bennet:
Looks like this is also caused by latest psycopg2. See if this fixes it:

`pip install psycopg2==2.6.2`

j-bennet:
Aha. Found it. https://github.com/psycopg/psycopg2/issues/524

msztolcman:
Yep, after downgrading psycopg it's better now, thanks :)

amjith:
Sigh. We should fix this. ☹️

j-bennet:
Fixed in 1.5.1.

PR #665, opened 2017-03-13
Fix set_session cannot be used inside a transaction when using dsn.

j-bennet:
Addresses #664

### DIFF
--- a/pgcli/pgexecute.py
+++ b/pgcli/pgexecute.py
@@ -189,10 +189,9 @@ class PGExecute(object):
             # When we connect using a DSN, we don't really know what db,
             # user, etc. we connected to. Let's read it.
             # Note: moved this after setting autocommit because of #664.
-            db = self._select_one(cursor, 'select current_database()')[0]
-            user = self._select_one(cursor, 'select current_user')[0]
-            host = self._select_one(cursor, 'select inet_server_addr()')[0]
-            port = self._select_one(cursor, 'select inet_server_port()')[0]
+            db, user, host, port = self._select_one(
+                cursor,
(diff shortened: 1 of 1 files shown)


amjith:
Good find. 👍 

Do you think this will also address https://github.com/dbcli/pgcli/issues/657?

j-bennet:
I don't think so, different exception. They already fixed this in psycopg2 though (https://github.com/psycopg/psycopg2/issues/524).

pgcli/pgexecute.py
```diff
@@ -190,6 +184,16 @@ def connect(self, database=None, user=None, password=None, host=None,
...
+
+        if dsn:
+            # When we connect using a DSN, we don't really know what db,
+            # user, etc. we connected to. Let's read it.
+            # Note: moved this after setting autocommit because of #664.
+            db = self._select_one(cursor, 'select current_database()')[0]
+            user = self._select_one(cursor, 'select current_user')[0]
+            host = self._select_one(cursor, 'select inet_server_addr()')[0]
+            port = self._select_one(cursor, 'select inet_server_port()')[0]
+
```
amjith: Can we collapse this into a single sql call instead of 4 calls? 

```
select current_database(), currect_user, inet_server_addr, inet_server_port
```

and assign the results to the appropriate variables? This should cut down on the start up time if the database instance is remote and there is a noticeable latency in the connection.

j-bennet: Ah good point. Give me a moment.

j-bennet:
@amjith Merge if looks ok?

amjith:
Nicely done! 

🎸

PR #667, opened 2017-03-15
Parametrize completer in tests + deduplicate test code

koljonen:
Instead of having a bunch of hard-coded completer fixtures (with different configs) in tests/test_smart_completion_multiple_schemata.py and tests/test_smart_completion_public_schema_only.py, let's have a completer factory in tests/metadata.py. Also, instead of running a test case for one completer config, let's run it for a bunch of different completer configs. That's the basic idea.

Also, I noticed there was some room for deduplication of the smart_completion code, so I moved a bunch of duplicated code from the test functions into metadata.py. There's no change in function for any commit except the first ("Parametrize ...").

### DIFF
--- a/tests/metadata.py
+++ b/tests/metadata.py
@@ -2,13 +2,24 @@ from pgcli.packages.parseutils.meta import FunctionMetadata, ForeignKey
 from prompt_toolkit.completion import Completion
 from functools import partial
 from itertools import product
+from prompt_toolkit.document import Document
+from mock import Mock
 
 escape = lambda name: ('"' + name + '"' if not name.islower() or name in (
     'select', 'insert') else name)
 
 def completion(display_meta, text, pos=0):
-    return Completion(text, start_position=pos,
-        display_meta=display_meta)
+    return Completion(text, start_position=pos, display_meta=display_meta)
+
+def get_result(completer, text, position=None):
+    position = len(text) if position is None else position
+    return completer.get_completions(
+        Document(text=text, cursor_position=position), Mock()
+    )
+
+def result_set(completer, text, position=None):
+    return set(get_result(completer, text, position))
+
 
 # The code below is quivalent to
 # def schema(text, pos=0):
--- a/tests/metadata.py
+++ b/tests/metadata.py
@@ -80,6 +80,31 @@ class MetaData(object):
         schemas = set(sch for schs in self.metadata.values() for sch in schs)
         return [schema(escape(s), pos=pos) for s in schemas]
 
+    def functions_and_keywords(self, schema='public', pos=0):
+        return (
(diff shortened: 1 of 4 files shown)


amjith:
Sorry I haven't had time to look at this. I'll get to it soon-ish, if no one else picks it up.

j-bennet:
I looked at this briefly, but I want to take a better look as well. It does make the test code less complex and more readable. I'm a little concerned that it may make debugging a broken test less clear - you'll have a longer stack trace, more calls to subroutines to get through - but it may be a worthwhile tradeoff.

tests/metadata.py
```diff
@@ -1,13 +1,30 @@
...
 from functools import partial
+from itertools import product
+from prompt_toolkit.document import Document
+from mock import Mock
+import pytest
 
 escape = lambda name: ('"' + name + '"' if not name.islower() or name in (
     'select', 'insert') else name)
 
+parametrize = pytest.mark.parametrize
```
darikg: This could just be `from pytest.mark import parametrize`

darikg:
I only skimmed it, but +1 from me

tests/metadata.py
koljonen: I'd actually struggled a bit with trying to write it as an import, but all I ended up with was this:
```
    from pytest.mark import parametrize
E   ImportError: No module named mark
```

j-bennet: Indeed, this is what happens, because `pytest.mark` is not a module, but a class with static methods. I also tried the same thing. :)

j-bennet:
I think this PR looks great. I only have nitpicks about pep8-compliant formatting (like two line breaks between function defs; no space around `=` in `method_call(kwarg_name = kwarg_value)`; docstrings should be `"""Docstring"""` and not `'''Docstring'''` etc.)

amjith:
@j-bennet I haven't been following strict pep-8 or even not so strict pep-8 in my code. 

Should we? If so, we need an automatic way of enforcing it instead of a human check.

amjith:
I don't think it is fair to hold up this PR for PEP-8. Especially since we never expressed that as a requirement in the past.

j-bennet:
I think it's more or less assumed for any Python code. We do not have an automatic checker, because there's no reason to enforce it strictly, this is true, but "not so strict" and still close to the guidelines would be good.

It's your project @amjith, so I can agree to disagree.

amjith:
> It's your project @amjith, so I can agree to disagree.

Nope. 

It is our project. I'm not even the lead anymore. I'm just expressing opinions not enforcing them (and never will).

darikg:
> I think it's more or less assumed for any Python code. We do not have an automatic checker, because there's no reason to enforce it strictly, this is true, but "not so strict" and still close to the guidelines would be good.

:+1:

Issue #668, opened 2017-03-15
\e is not working on pgcli 1.5.0 installed via brew


amjith:
Conversation from gitter: 

> When I write `select 1 \e`, I correctly see `select 1` in my editor. When I edit and quit, I don't see `select 2`. I don't see anything at all. If I type `;<RET>`, I get the can't execute an empty query error message.
> 
> nvim 0.1.7, macOS Sierra (10.12.1). nvim exits normally when I do the same thing from bash. And it works with psql. I believe it also worked fine in pgcli prior to my last brew upgrade, but I'm not sure what version I was on previously.
> Judging by `brew list --versions pgcli`, I was probably on 1.3.1_1.
> 

I'm unable to reproduce it since my brew setup is screwed up. 

Steps to reproduce:

1. brew update && brew install pgcli
2. Launch pgcli
3. Type `select 1 \e`
4. Change the query to `select 2` in the editor. 
5. Save and quit the editor. 
6. Check to see if the prompt shows `select 2`.

jdpopkin:
I'm ignorant of these libraries and Python in general, so I crammed a bunch of printlns into `handle_editor_command` to try to figure out what's going on here.

    def handle_editor_command(self, cli, document):
        """
        Editor command is any query that is prefixed or suffixed
        by a '\e'. The reason for a while loop is because a user
        might edit a query multiple times.
        For eg:
        "select * from \e"<enter> to edit it in vim, then come
        back to the prompt with the edited query "select * from
        blah where q = 'abc'\e" to edit it again.
        :param cli: CommandLineInterface
        :param document: Document
        :return: Document
        """
        while special.editor_command(document.text):
            self.logger.error("XYZZY initial document text: " + document.text)
            filename = special.get_filename(document.text)
            sql, message = special.open_external_editor(filename,
                                                          sql=document.text)
            self.logger.error("XYZZY " + sql)
            if message:
                # Something went wrong. Raise an exception and bail.
                self.logger.error("XYZZY message: " + message)
                raise RuntimeError(message)
            cli.current_buffer.document = Document(sql, cursor_position=len(sql))
            self.logger.error("XYZZY no message")
            self.logger.error("XYZZY middle-of-loop current-buffer-document text: " + cli.current_buffer.document.text)
            document = cli.run(False)
            self.logger.error("XYZZY tail-of-loop document text: " + document.text)
            self.logger.error("XYZZY tail-of-loop current-buffer-document text: " + cli.current_buffer.document.text)
[...]

PR #669, opened 2017-03-15
Fix \e

jdpopkin:


jdpopkin:
There's probably a better way to do this, or at least a better place to put this code. Do let me know. (Or just close it and fix it in a cleaner way; I just want my pgcli to work again 😿)

This fixes #668 by working around changes introduced by jonathanslenders/python-prompt-toolkit@9030182.

jdpopkin:
Actually, there's a place where pgcli expects a reset to happen: https://github.com/dbcli/pgcli/pull/669/files#diff-38f3aeda57742756afa677ca09dcea43R446

So maybe we should overwrite cli.accept_action both before and after that call to cli.run. But that sounds awkward to me.

j-bennet:
Perhaps @jonathanslenders could advise.

jonathanslenders:
@j-bennet: What is the actual problem this is supposed to solve? I don't completely understand it.

jonathanslenders:
Ok, I see what's going on.
One possibility would be to use indeed a custom ``AcceptAction``. But rather then calling ``set_return_value``, check whether the input equals ``\e``. If so, call ``buffer.open_in_editor``. If not, return as usual. (like ``AcceptAction.RETURN_DOCUMENT`` does.)

jdpopkin:
Closing this in favor of #670 - thanks, everyone! 🎈

PR #670, opened 2017-03-16
Attempt to fix #668 (not pretty).

### LINKED ISSUE #668: \e is not working on pgcli 1.5.0 installed via brew

j-bennet:
This is an attempt to fix #668 .

There's another PR to do the same: https://github.com/dbcli/pgcli/pull/669, but it does not fix the problem completely, because we need to reset the cli buffer in most cases. Editor command seems to be the special case where we don't want the reset.

This PR is work in progress, because I dislike using internal `prompt_toolkit` API like this. I would like very much for @jonathanslenders to weigh in on this.

cc @amjith @jdpopkin

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -351,6 +351,7 @@ class PGCli(object):
         :param document: Document
         :return: Document
         """
+        saved_callables = cli.application.pre_run_callables
         while special.editor_command(document.text):
             filename = special.get_filename(document.text)
             sql, message = special.open_external_editor(filename,
@@ -359,8 +360,10 @@ class PGCli(object):
                 # Something went wrong. Raise an exception and bail.
                 raise RuntimeError(message)
             cli.current_buffer.document = Document(sql, cursor_position=len(sql))
-            document = cli.run(False)
+            cli.application.pre_run_callables = []
+            document = cli.run()
             continue
+        cli.application.pre_run_callables = saved_callables
         return document
 
     def execute_command(self, text, query):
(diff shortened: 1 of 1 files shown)


amjith:
I don't know enough about this change to say whether it is the right call or not. But it looks like Jonathan agrees that there is no ideal solution. So if this fixes the issue let's merge this. 

Thoughts?

j-bennet:
I'm going to merge it with an additional FIXME comment. Jonathan was talking about some major refactorings in his next release of `prompt_toolkit`, so I want to mark the potentially breakable spot.

jonathanslenders:
I'll let you know in advance when 2.0 is going to be released. There's still some work to do, but probably I'll synchronize it with an IPython release and maybe other prompt_toolkit applications.
Right now, the 2.0 branch feel stable, but I'm going to rewrite at least ptpython, pymux and pyvim against 2.0 first in order to see whether the API is good enough.

PR #673, opened 2017-03-19
Functional tests housekeeping

j-bennet:
## Description

* Removes an unnecessary step from functional tests.
* Splits test steps into different files.
* Adds support for coverage.

Work in progress.

## Checklist


j-bennet:
@meeuw I did the rename, but I don't see any syntax errors in that line. Can you take another look? Thanks!

meeuw:
Hi @j-bennet, thanks for the rename. This test is wrong because it's matching the nano buffer and not the output in pgcli. It should read something like `wrappers.expect_exact(context, 'select\x1b[0m \x1b[0;38;5;241m*\x1b[0m \x1b[0;38;5;28;1mfrom\x1b[0m \x1b[0mabc', timeout=2)` to match the correct output.

meeuw:
I've made some fixes for the functional tests please have a look: 1ed8333e5d4da1d7a17e3df3e61c675bc4874648

j-bennet:
@meeuw Ah, now I see what you're saying. Yes, I was trying to avoid checking for unreadable strings like that. I wish that there was an option to run the cli with color output disabled. I asked @jonathanslenders about it, but he said there's no such thing. I'll see if I can use a regex there.

j-bennet:
@meeuw So I added a regexp instead. I don't want to bother checking color codes for correctness, so I'm allowing anything between actual query tokens.

j-bennet:
@meeuw That was a very nice catch. :) Apparently this test never even worked in travis, because nano was not installed. Checking for colored output revealed the problem.

meeuw:
@j-bennet I don't have nano installed on my system so I've switched to ex as editor for mycli:
https://github.com/dbcli/mycli/blob/meeuw/pgcli_behave_framework/tests/features/steps/iocommands.py

meeuw:
@j-bennet please have a look at the other changes I've made:
https://github.com/meeuw/pgcli/commit/5e2bbd62e3e69bcdcfff898b1e58d8dd3af9fdbb

PGPASS should be PGPASSWORD and pgcli should be run in tests/.. to get the coverage output files in the right directory. I've also updated .travis.yml to combine the coverage files.

tests/features/steps/iocommands.py
```diff
@@ -0,0 +1,44 @@
...
+
+@when('we exit the editor')
+def step_edit_quit(context):
+    context.cli.sendcontrol('x')
+
+
+@then('we see the sql in prompt')
+def step_edit_done_sql(context):
+    colored_expr = re.compile('select(.+?)\*(.+?)from(.+?)abc')
+    wrappers.expect(context, colored_expr, timeout=2)
```
meeuw: I think you could also do:
```
wrappers.expect_exact('select')
wrappers.expect_exact('*')
wrappers.expect_exact('from')
wrappers.expect_exact('abc')
```

j-bennet: I know, but that doesn't look any any better to me.

j-bennet:
@meeuw I never worked with ex editor. Do you think that it's better than nano for functional tests? Why?

Thanks for the other change, https://github.com/meeuw/pgcli/commit/5e2bbd62e3e69bcdcfff898b1e58d8dd3af9fdbb. I'll port that over to my branch.

tests/features/steps/iocommands.py
meeuw: Maybe this is better?
```
for match in 'select * from abc'.split(' '):
    wrappers.expect_exact(context, match, timeout=1)
```

It saves you from adding `wrappers.expect` and I guess is might be a tiny bit faster because you're not using regular expressions.

meeuw:
I think ex is better because
1. I think ex is POSIX (and installed on every unix system)
2. ex is built for scripting (an editor like nano might have surprises for different versions)
3. We are testing pgcli and not the editor, I think we should choose what can be scripted best.

But I'm not very known to functional tests and in the end the nano code is less than what I've written for ex so it's open for discussion what's better.

tests/features/steps/iocommands.py
j-bennet: I like this one. I'll use that.

j-bennet:
@meeuw Like you said, we're not testing the editor. I chose `nano` because command sequences are short (only send one sequence to write out the file, one to exit). But I don't mind switching to `ex`. If it's available on most systems (including travis) without an additional install, that's a good argument. Would you be able to commit the `nano` -> `ex` change into this branch?

meeuw:
sure, here it is: f1e0da998ebdf6753e3917ec8b6895f5a24db035 ([meeuw/pgcli.git j-bennet/behave-tests-housekeeping](/meeuw/pgcli/commits/j-bennet/behave-tests-housekeeping))

j-bennet:
Thank you @meeuw. I think it's time to add you to the `pgcli-core` team as well.

amjith:
> Thank you @meeuw. I think it's time to add you to the pgcli-core team as well.

👍

j-bennet:
@meeuw You're now a member of pgcli-core team. Congratulations. :) 🎈 

I think the housekeeping should be good to go. What do you think?

meeuw:
😀 thanks. I think this is a big improvement. After merging this I'll refresh mycli's PR and then that is also ready to merge. I'm working on a follow up PR with a custom pager (which add a starting and trailing boundary like mime mail) so the results can be matched better.

j-bennet:
Merged!

PR #678, opened 2017-03-30
Allow configurable multiline continuation char

owst:
## Description
Makes the character used for continuing multi-line queries configurable. To me, having a space rather than '.' is cleaner, (particularly when pasting the query/output for others to read). 

Now, setting `multiline_continuation_char = ' '` changes:
```
postgres@localhost:testdb1> SELECT *
........................... FROM foo
........................... WHERE bar = 1;
```
to
```
postgres@localhost:testdb1> SELECT *
                            FROM foo
                            WHERE bar = 1;
```

## Checklist

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -56,6 +56,7 @@ Contributors:
     * Manuel Barkhau
     * Sergii V
     * Emanuele Gaifas
+    * Owen Stephens
 
 Creator:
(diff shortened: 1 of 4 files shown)


amjith:
Very nice! 

Thank you for the contribution. 🍨

PR #679, opened 2017-04-05
Standardize command line option names

russelldavies:
This standardizes on the same names as psql and removes some
conflicts.

## Description
I was using a database tool that was launching pgcli (thinking it was psql) with the `--username` option. This failed as pgcli expected the username option as `--user`. Since pgcli has extra and different functionality than psql it makes sense to have additional command line options but I think it's important for compatibility purposes to mimic psql's command line interface.

## Checklist

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -57,6 +57,7 @@ Contributors:
     * Sergii V
     * Emanuele Gaifas
     * Owen Stephens
+    * Russell Davies
 
 Creator:
 --------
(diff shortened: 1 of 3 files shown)


j-bennet:
I agree with `--username`, but I think `-V` in psql is very annoying. I would not want to copy that.

russelldavies:
The `-v` option conflicts with psql's `-v assignment` but I suppose if you don't see it conflicting in the future then I can revert it.

j-bennet:
Yes, how about we leave `-v` in pgcli until someone actually implements `-v VAR_NAME=var_value`.

russelldavies:
Okay, I've updated the commit with that change.

j-bennet:
Thank you!
🎈

PR #680, opened 2017-04-08
Ported pep8radius addition from mycli.

j-bennet:
## Description

Ported from mycli. Discussion: https://github.com/dbcli/mycli/issues/392.

## Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,12 +8,15 @@ python:
 
 install: 
   - pip install . pytest mock codecov==1.5.1 behave pexpect==3.3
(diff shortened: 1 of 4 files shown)


amjith:
@dbcli/pgcli-core This PR adds [pep8radius](https://pypi.python.org/pypi/pep8radius) to the build step. This will check the lines that were modified for pep8 compliance. The only exception from pep8 is the line length (you can have up to 99 chars in a line). 

So if the code you're modifying or adding doesn't comply with pep8 you might see PRs failing.

PR #684, opened 2017-04-19
Change author name and email.

amjith:
## Description
Update author and email.


j-bennet:
@amjith pep8radius seems to be unhappy with the change.

PR #686, opened 2017-04-23
Suggest columns for `ORDER BY` and `DISTINCT` (fixes #685)

### LINKED ISSUE #685: Autocompletion includes alias name twice for `ORDER BY` clause in `SELECT`
## Description
When I write a simple `SELECT` query `JOIN`ing an aliased table, the autocompletion of the `ORDER BY` clause incorrectly includes the alias name twice.

In particular, given tables:
```
> \d foo
+----------+---------+-------------+
| Column   | Type    | Modifiers   |
|----------+---------+-------------|
| id       | integer |             |
+----------+---------+-------------+
Indexes:
    "iduniq" UNIQUE CONSTRAINT, btree (id)
Referenced by:
    TABLE "stuff" CONSTRAINT "foofk" FOREIGN KEY (foo_id) REFERENCES foo(id) MATCH FULL
```
and
```
> \d stuff
+-----------+--------------------------+-------------+
| Column    | Type                     | Modifiers   |
|-----------+--------------------------+-------------|
| foo_id    | integer                  |             |
| some_time | timestamp with time zone |             |
+-----------+--------------------------+-------------+
Foreign-key constraints:
    "foofk" FOREIGN KEY (foo_id) REFERENCES foo(id) MATCH FULL
```

if I write:
```
SELECT * FROM foo JOIN stuff s ON s.foo_id = foo.id ORDER BY s.
```
and hit tab and then select `s.some_time` the query is completed as
```
SELECT * FROM foo JOIN stuff s ON s.foo_id = foo.id ORDER BY s.s.time
```
rather than 
```
SELECT * FROM foo JOIN stuff s ON s.foo_id = foo.id ORDER BY s.time
```
it looks like the auto complete "hasn't recognised" the `s.` context as the options it offers me are:
```
foo.id
s.foo_id
s.some_time
```
rather than what I would expect:
```
foo_id
some_time
```
(i.e. columns on `s` without the alias prefix)

Interestingly, the autocompletion works fine for the `WHERE` clause.
## Your environment
OSX 10.10.5
HEAD (`66cd634`)
```
>pip freeze
appdirs==1.4.3
click==6.7
configobj==5.0.6
humanize==0.5.1
packaging==16.8
-e git+git@github.com:dbcli/pgcli.git@66cd634de462c685cd11d08b56e39289dbdd109e#egg=pgcli
pgspecial==1.7.0
prompt-toolkit==1.0.14
psycopg2==2.7.1
Pygments==2.2.0
pyparsing==2.2.0
setproctitle==1.1.10
six==1.10.0
sqlparse==0.2.3
wcwidth==0.1.7
```

owst:
## Description
Having typed an alias name in an `ORDER BY` or (`SELECT`) `DISTINCT` clause,
the alias was not taken account of, and the completion simply listed all
columns. This change fixes that, and makes the autocompletion behave the
same as in `SELECT` and `WHERE` clauses.

## Checklist

### DIFF
--- a/tests/test_sqlcompletion.py
+++ b/tests/test_sqlcompletion.py
@@ -4,8 +4,9 @@ from pgcli.packages.sqlcompletion import (
 from pgcli.packages.parseutils.tables import TableReference
 import pytest
 
-# Returns the expected select-clause suggestions for a single-table select
 def cols_etc(table, schema=None, alias=None, is_function=False, parent=None):
+    """Returns the expected select-clause suggestions for a single-table
+    select."""
     return set([
         Column(table_refs=(TableReference(schema, table, alias, is_function),),
                qualifiable=True),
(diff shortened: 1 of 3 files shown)


amjith:
Thanks for taking the time to make pgcli better. Sorry I haven't had time to look into this. I won't have time till end of this week, but if none of the other core devs have reviewed it by then, I'll take a look. Sorry about the delay.

tests/test_sqlcompletion.py
```diff
@@ -4,10 +4,17 @@
 from pgcli.packages.parseutils.tables import TableReference
 import pytest
 
-# Returns the expected select-clause suggestions for a single-table select
+
+def table_reference(table, schema=None, alias=None, is_function=False):
```
j-bennet: `TableReference` instantiation is already a one-liner. What's the point of wrapping it into a function?

tests/test_sqlcompletion.py
```diff
@@ -213,9 +220,62 @@ def test_truncate_suggests_qualified_tables():
 ])
 def test_distinct_suggests_cols(text):
     suggestions = suggest_type(text, text)
-    assert suggestions ==(Column(table_refs=(), qualifiable=True),)
+    assert set(suggestions) == set([
```
j-bennet: Well done adding tests!

tests/test_sqlcompletion.py
owst: On reflection, probably none, I think I added it before I did a bunch of line-wrapping. Of course, one can avoid passing a few parameters since `table_reference` has default parameters, but I don't think the method pulls its weight, I'll remove it.

tests/test_sqlcompletion.py
owst: Wouldn't be a good PR without them...!

owst:
No worries about taking your time @amjith! Thanks for the review @j-bennet - I've addressed the comment and squashed it into one commit since it was a small tweak (I hope that's ok for this project!)

j-bennet:
@owst Also, the line in changelog:

 "Fixed completions after ORDER BY and DISTINCT when using table aliases".

owst:
Yeah, fair enough, I'm certainly not wedded to the current wording. The only thing is "fixed" to me suggests it was "broken" before, when I think it was just a missing feature. Perhaps 
> Completions after ORDER BY and DISTINCT now take account of table aliases.

?

j-bennet:
@owst Yes, that is better!

j-bennet:
Looks good. Merging.

Thank you!

🌷

PR #688, opened 2017-04-25
Configure suggestions for \sf

lelit:
This requires https://github.com/dbcli/pgspecial/pull/29, but shouldn't hurt even before it gets integrated.


j-bennet:
👍

PR #689, opened 2017-04-25
WIP - Complete ALTER SYSTEM

bersace:
Hi,

This PR add completion for keywords defined in https://www.postgresql.org/docs/9.4/static/sql-altersystem.html . As a side effect, `ALTER` is now easier to use for `ALTER DATABASE`, `ALTER USER`, etc. other than `ALTER TABLE`.

Is it possible to reduce keywords suggested depending on previous keywords ?

I want to complete settings too. Do you want to make this in another PR ?


j-bennet:
@bersace I don't think we don't have keyword filtering based on previous keywords yet. If you wanted to add that functionality, you'd be looking at these two files, where the completion logic resides:

https://github.com/dbcli/pgcli/blob/master/pgcli/pgcompleter.py
https://github.com/dbcli/pgcli/blob/master/pgcli/packages/sqlcompletion.py

[These lines](https://github.com/dbcli/pgcli/blob/5da1a4154cdba3e5b4462627a049b8d361c27e66/pgcli/packages/sqlcompletion.py#L497-L498) currently tell the completer that when it sees an `ALTER` token, it should return keyword matches. You'd probably want to make changes to `Keyword` named tuple and in [`get_keyword_matches` here](https://github.com/dbcli/pgcli/blob/0df6e37dcafa0fde7d9bcadfb6c85f7a504eb43d/pgcli/pgcompleter.py#L652-L666).

j-bennet:
@bersace Oh, and to address your other question, what do you mean by completing settings?

bersace:
@j-bennet i would like to complete settings from `pg_catalog.pg_settings` when typing `SHOW ` or `ALTER SYSTEM SET `.

darikg:
More sophisticated suggestions based more than just the last token are a good idea. I wonder if it might be sufficient to check the _first_ token around [here] somewhere(https://github.com/dbcli/pgcli/blob/master/pgcli/packages/sqlcompletion.py#L154) and delegate to a more specific function. I'm imagining something like

```
first_tok = get_first_token_somehow()
if == 'ALTER':
    return suggest_based_on_last_token_in_alter_context(stmt.last_token, stmt)
elif first_tok == 'CREATE':
    return suggest_based_on_last_token_in_create_context(stmt.last_token, stmt)
else:
    return suggest_based_on_last_token(stmt.last_token, stmt)
```

This might maintain the simplicity of only looking at the previous token but increase suggestion power

darikg:
fyi you can get the first token with

```first_tok = stmt.parsed.token_next(-1, skip_ws=True, skip_cm=True)[1].value.upper()```

(There's probably a couple of degenerate cases like empty statements you'll have to look for first)

**Comment by bersace** [comment]
@j-bennet i will do another PR to complete pg_settings.

@darikg I did update pgliterals to describe *following* keywords. What do you think of this ? This is quite naïve but improve the situation.

**Comment by darikg** [comment]
I think there's a small mistake here that is causing a lot of unnecessary changes to the tests.

The problem is around [here](https://github.com/dbcli/pgcli/pull/689/files#diff-36fcb873925169e72e8c5ca4e8a8a581R653)
```
keywords = self.keywords_tree.keys()
if last_token:
    sub_keywords = self.keywords_tree.get(last, [])
    if sub_keywords:
        keywords = sub_keywords
```
If `last_token` is empty, or `sub_keywords` is empty, `keywords` remains equal to `keywords_tree.keys()`.

However, a lot of the tests compare to [`MetaData.keywords()`](https://github.com/dbcli/pgcli/blob/a56ff81b19fb112f5559a6fa56f7d9293450ed8a/tests/metadata.py#L68) which returns `[keyword(kw, pos) for kw in self.completer.keywords]` rather than `[keyword(kw, pos) for kw in self.completer.keyword_tree.keys()]`

You might want to change one or the other depending on you think the suggestions should behave

bersace:
@darikg yes, I prefer to narrow default keywords suggestions ton *root* keywords. e.g. we don't want ton suggest `TABLE` other than in `ALTER`, `CREATE`, etc.

And yes, all tests are based on allways-all-keywords.

amjith:
Is this PR still valid, considering that we've just merged this PR https://github.com/dbcli/pgcli/pull/693?

darikg:
no we can close it

PR #692, opened 2017-04-27
Update DEVELOP.rst with external pgspecial project

bersace:
`DEVELOP.rst` is wrong since pgspecial has been extracted in a dedicated project. I suggest to simply point to pgspecial project.


amjith:
👍

PR #693, opened 2017-04-27
Complete well known following keywords

bersace:
I extracted the *well known following keywords* mechanism from #689 in a dedicated PR.

The basic idea is to store keywords as a dict of *keyword* -> list of *well known following keywords*.  This allow to provide a smarter set of candidates e.g. after `CREATE `.


bersace:
Green :green_heart: :)

tests/test_sqlcompletion.py
```diff
@@ -709,12 +716,13 @@ def test_statements_in_function_body(text):
 @pytest.mark.parametrize('text', functions)
 def test_statements_with_cursor_after_function_body(text):
     suggestions = suggest_type(text, text[:text.find('; ') + 1])
-    assert set(suggestions) == set([Keyword()])
+    assert set(suggestions) == set([Keyword(), Special()])
```
bersace: I think this closes https://github.com/dbcli/pgcli/pull/695

PR #695, opened 2017-04-27
Bug in test_statements_with_cursor_after_function_body

bersace:
Hi,

While working on #693 , i hit a bug in `test_statements_with_cursor_after_function_body`. Here is just the diff making CI fails. Can you help me debugging this ?

Regards,
Étienne


j-bennet:
@bersace Looks like a valid test failure to me, because `len(set(suggestions))` should be `== 2` [here](https://github.com/dbcli/pgcli/blob/293b6b3a412255cc2c12f0fe94d41760d4f47d8e/tests/test_sqlcompletion.py#L716) and not `== 1`. Is this still a problem?

bersace:
@j-bennet yep, but I find the `set` comparison bellow misleading.

PR #698, opened 2017-04-28
behave quit pgcli nicely

meeuw:
## Description
behave quit pgcli nicely so that coverage can be collected, before this patch pgcli was killed by pexpect at the end of a feature test

## Checklist


j-bennet:
👍

PR #701, opened 2017-04-28
Fix handling of failure to obtain lock

### LINKED ISSUE #700: Client disconnects upon failing to acquire a lock
## Description

When I fail to acquire a lock (via `SELECT [...] FOR UPDATE NOWAIT;`), pgcli prompts me to reconnect:

    db> select * from user u where id = 1 for update nowait;
    Connection reset. Reconnect (Y/n)

More-verbose repro instructions: run two instances of `pgcli` (in, say, two Terminal.app windows). Run the following in each of them:

    db> begin;
    db> select * from user u where id = 1 for update nowait;

The second `pgcli` instance (the one that failed to grab the lock) will prompt you to reconnect.

## Your environment

macOS 10.12.4
1.5.1

```    backports.ssl-match-hostname==3.5.0.1
    botocore==1.5.44
    click==6.7
    colorama==0.3.7
    configobj==5.0.6
    docutils==0.13.1
    feedparser==5.2.1
    futures==3.1.1
    greenlet==0.4.9
    humanize==0.5.1
    jedi==0.10.0
    jmespath==0.9.2
    mock==1.0.1
    msgpack-python==0.4.7
    neovim==0.1.7
    pathlib==1.0.1
    pgcli==1.5.1
    pgspecial==1.7.0
    pika==0.10.0
    Pillow==3.1.1
    prompt-toolkit==1.0.9
    psycopg2==2.7.1
    py==1.4.26
    pyasn1==0.2.3
    Pygments==2.2.0
    pypandoc==1.3.3
    pytest==2.7.0
    python-dateutil==2.6.0
    pytz==2016.7
    PyYAML==3.12
    Pyzotero==1.1.19
    rauth==0.7.2
    requests==2.12.1
    rsa==3.4.2
    s3transfer==0.1.10
    setproctitle==1.1.10
    six==1.10.0
    sqlparse==0.2.3
    trollius==2.1
    virtualenv==14.0.5
    wcwidth==0.1.7
    websocket-client==0.40.0
    -e git+git@github.com:jbaiter/zotero-cli.git@5ad46e8388a109ab6afc8d31d28297e742e48528#egg=zotero_cli```

psycopg treats this particular error (`55P03`, `lock_not_available`) as an `OperationalError`, and pgcli prompts to reconnect when it hits one of those. I have a fix; I'll make a PR soon.

### PR DESCRIPTION
**Pull request by jdpopkin**
## Description

Fixes #700.

## Checklist

### DIFF
--- a/pgcli/pgexecute.py
+++ b/pgcli/pgexecute.py
@@ -2,6 +2,7 @@ import traceback
 import logging
 import psycopg2
 import psycopg2.extras
+import psycopg2.errorcodes
 import psycopg2.extensions as ext
 import sqlparse
 import pgspecial as special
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: pgcli/pgexecute.py [hunk 1]
```diff
@@ -307,6 +308,12 @@ def run(self, statement, pgspecial=None, exception_formatter=None,
                 if not on_error_resume:
                     break
 
+    def _cannot_catch(self, e):
```
j-bennet: I think that `_cannot_catch` is not a very descriptive name. We actually can catch the exception here, we just prefer not to. I would rather use something like `_must_raise` or `_should_raise`.

pgcli/pgexecute.py
```diff
@@ -307,6 +308,12 @@ def run(self, statement, pgspecial=None, exception_formatter=None,
...
+        # Raise operational errors regardless of on_error
+        # specification.
```
j-bennet: Also, the comment says "Raise operational errors regardless of on_error specification", but it does not fully apply anymore, since we now raise selectively, and not on any OperationalError. Now that this logic moved into a method, there should also be a properly formatted docstring.

jdpopkin:
Hi @j-bennet - let me know if this is any better?

A link to https://github.com/psycopg/psycopg2/blob/master/psycopg/pqpath.c#L80 would probably do a better job than my docstring of explaining exactly what psycopg2 classifies as an `OperationalError`, but it'd also go stale...

j-bennet:
Better! I would also add a comment that we don't want to raise an OperationalError in case of `LOCK_NOT_AVAILABLE `, since we handle those errors by reconnecting, and there's no point to reconnect in this case.

j-bennet:
👍

PR #702, opened 2017-04-29
Fix typo in parseutils.meta.FunctionMetadata

hatarist:
## Description
Just noticed that it was "type" instead of "typ" in the list comprehension.
I'm not sure what cases this issue was covering, so sorry, I have no idea what to write in changelog.
It's also too minor to add myself to the `AUTHORS` :)

## Checklist


j-bennet:
Nice catch! Thank you for the PR.

Issue #580, opened 2016-09-12
How to display time in prompt?


d33tah:
I'd like my prompt to be prepended with `[`date "+%d/%m/%y %H:%M:%S"`]` so I could see how long queries take and when they finish. How could I do that most easily?

amjith:
@d33tah We currently don't have that available in the prompt specifiers. But it is trivial to add. Here's the relevant code snippet. https://github.com/dbcli/pgcli/blob/master/pgcli/main.py#L687-L694

I wouldn't be opposed to the idea if you want to send a PR. 😄 

ps: We do time the queries and print the results at the end of each query. In fact we fixed a bug recently to fix a bug in the calculation #578.

gustavo-castro:
I would like to be assigned this issue if possible

j-bennet:
@gustavo-castro Absolutely! Thank you for your help!

j-bennet:
@gustavo-castro So looks like we can't assign an issue to a non-collaborator (https://github.com/isaacs/github/issues/100), but go ahead and send a PR anyway. PRs are always welcome.

gustavo-castro:
@j-bennet Great! I'll send it by the week.

j-bennet:
Fixed via https://github.com/dbcli/pgcli/pull/715.

PR #715, opened 2017-05-09
Adding time option for the Prompt Format

gustavo-castro:
## Description
I added a date option for the prompt (as asked for in #580 ). To use it, you just have to add "\\t" to the Prompt Format option.

## Checklist

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -11,7 +11,7 @@ import threading
 import shutil
 import functools
 import humanize
-import datetime
+import datetime as dt
 from time import time, sleep
 from codecs import open
 
@@ -153,7 +153,7 @@ class PGCli(object):
         self.decimal_format = c['data_formats']['decimal']
         self.float_format = c['data_formats']['float']
 
-        self.now = datetime.datetime.today()
+        self.now = dt.datetime.today()
 
         self.completion_refresher = CompletionRefresher()
 
@@ -494,7 +494,7 @@ class PGCli(object):
                 else:
                     query = self.execute_command(document.text, query)
 
-                self.now = datetime.datetime.today()
+                self.now = dt.datetime.today()
 
                 # Allow PGCompleter to learn user's preferred keywords, etc.
                 with self._completer_lock:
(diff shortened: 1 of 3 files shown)


pgcli/main.py
```diff
@@ -491,6 +494,8 @@ def run_cli(self):
                 else:
                     query = self.execute_command(document.text, query)
 
+                self.now = datetime.datetime.today()
```
gustavo-castro: I tried it at first, but the problem is the program calls get_prompt for every single action that is done (typing one letter, erasing one letter, etc), so it keeps updating the timer every single time get_prompt is called.

pgcli/main.py
```diff
@@ -11,6 +11,7 @@
 import shutil
 import functools
 import humanize
+import datetime
```
gustavo-castro: Going to change it.

pgcli/main.py
j-bennet: Ah, good point! In this case it's a sensible choice.

gustavo-castro:
Sorry about the question, I'm quite new to open-source contributing, but I guess I should take a look at the fact that one check has failed? I actually created it in the european format, but it seems this can be easily fixed, there are parameters that actually do this, I will change it too.

j-bennet:
@gustavo-castro Oh good! I only saw the initial `'%d/%m/%y %H:%M:%S'`. Yes, the `'%x %X'` is better.

Once you add your contribution to AUTHORS and changelog, this PR should be good to merge.

Thank you!

j-bennet:
@gustavo-castro Oh, and the tests fail because you added your name to changelog, but it refers to a missing link, which should go in the end of changelog file, like this:

```
.. _`Gustavo Castro`: https://github.com/gustavo-castro
```

j-bennet:
@gustavo-castro, you are so close. :) Just add that missing link to the changelog.

gustavo-castro:
@j-bennet Thanks for the help and the incentive! :) I didn't have access to my home computer during the day (european time), I was intending to finish it earlier.

j-bennet:
Perfect. :) Merging.

Thank you for the PR, and congratulations on your first contribution to `pgcli`!

🍬

Issue #717, opened 2017-05-15
alias_dns not loading from custom PGCLIRC


OmeGak:
## Description

I'm loading the config specifying `PGCLIRC` in the environment. The DNS aliases defined in the custom location pgclirc are ignored and loaded from the standard location pgclirc instead (`~/.config/pgcli/config`). Other config options like `multi_line` are successfully loaded from custom location pgclirc. The same thing happens when specifying pgclirc from CLI options.

## Your environment

```console
$ uname -v
Darwin Kernel Version 16.5.0: Fri Mar  3 16:52:33 PST 2017; root:xnu-3789.51.2~3/RELEASE_X86_64
```

```console
$ pgcli --version
Version: 1.5.1
```

```
$ pip freeze
pipg freeze
alabaster==0.7.6
appnope==0.1.0
argcomplete==0.8.9
Babel==1.3
backports.shutil-get-terminal-size==1.0.0
backports.ssl-match-hostname==3.4.0.2
blindspin==2.0.0
certifi==2015.4.28
click==6.7
click-completion==0.2.1
colorama==0.3.7
configparser==3.5.0
crayons==0.1.2
cycler==0.10.0
decorator==4.0.10
delegator.py==0.0.8
docutils==0.12
enum34==1.1.6
filemagic==1.6
flake==0.1.9
flake8==3.2.1
flake8-docstrings==1.0.2
functools32==3.2.3.post1
gnureadline==6.3.3
httpie==0.9.4
icalendar==3.0
ipykernel==4.2.0
ipython==5.1.0
ipython-genutils==0.1.0
ipywidgets==4.1.1
jedi==0.9.0
Jinja2==2.7.3
jsonschema==2.5.1
jupyter==1.0.0
jupyter-client==4.1.1
jupyter-console==4.0.3
jupyter-core==4.0.6
kaptan==0.5.8
libtmux==0.6.3
log-colorizer==1.6
Mako==1.0.6
MarkupSafe==0.23
matplotlib==1.5.1
mccabe==0.5.2
mistune==0.7.1
nbconvert==4.1.0
nbformat==4.0.1
nodeenv==0.13.3
nose==1.3.7
notebook==4.0.6
numpy==1.11.0
numpydoc==0.5
parse==1.6.6
path.py==8.1.2
pathlib==1.0.1
pathlib2==2.1.0
pbr==3.0.0
pep8==1.5.7
pew==0.1.26
pexpect==4.2.1
pickleshare==0.7.4
[...]

PR #722, opened 2017-05-28
Fix alias dsn from custom rcfile

j-bennet:
## Description

Fix for #717.

## Checklist

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -823,7 +823,7 @@ def cli(database, username_opt, host, port, prompt_passwd, never_prompt,
 
     if dsn is not '':
         try:
-            cfg = load_config(config_full_path)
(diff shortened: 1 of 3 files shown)


pgcli/main.py
```diff
@@ -823,7 +823,7 @@ def cli(database, username_opt, host, port, prompt_passwd, never_prompt,
 
     if dsn is not '':
         try:
-            cfg = load_config(config_full_path)
+            cfg = load_config(pgclirc, config_full_path)
```
amjith: I think you have the order of the arguments reversed. The second arg is the def_cfg which should point to pgclirc.

j-bennet: I think first one is supposed to be custom config, and second one default config, and this is what I'm passing in.

amjith:
Looks good. 🌮

Issue #718, opened 2017-05-16
PostgreSQL update during pgcli installation via Homebrew


pboschmann:
## Description
Hi there,

I discovered pgcli yesterday and gave it a shot (like it so far).
I installed it via Homberew and unfortunately it seems that Homebrew also updated Postgres during the installation process. I can not start the Postgres server with my existing database files anymore. PostgreSQL server log says:
```
FATAL:  database files are incompatible with server
DETAIL:  The data directory was initialized by PostgreSQL version 9.4, which is not compatible with this version 9.6.3.
```
I've checked my ``/usr/local/bin`` directory - the timestamps of the pgcli and postgres symlinks are equal, whereas Python related symlinks and all of my other Homebrew installed programs are of an older date.
Could it be, that there is a bug in the Homebrew formula and Postgres will be always installed in bundle with pgcli? I couldn't find the formula for pgcli in this repository and don't have time ATM to investigate further.

## Your environment
macOS 10.12.4
pgcli 1.5.1
PostgreSQL 9.4 yesterday, and 9.6.3 today 🙈 
pip freeze says:
```
backports-abc==0.4
certifi==2016.8.31
click==6.6
Jinja2==2.8
livereload==2.4.1
Markdown==2.6.6
MarkupSafe==0.23
mkdocs==0.15.3
mkdocs-bootstrap==0.1.1
mkdocs-bootswatch==0.4.0
PyYAML==3.12
singledispatch==3.4.0.3
six==1.10.0
tornado==4.4.1
vboxapi==1.0
wheel==0.26.0
```

j-bennet:
`pgcli` brew formula [specifies postgresql as a dependency](https://github.com/Homebrew/homebrew-core/blob/master/Formula/pgcli.rb#L16), but without attaching to any particular version. I'm not sure if that is the source of the problem; have to look into it.

j-bennet:
So for `pgcli` in the brew formula, we specify `python`, `openssl` and `postgresql` as dependencies. For `mycli`, we only specify `python` and `openssl`. Looks like we could safely leave out `postgresql` from `pgcli` formula.

j-bennet:
Some context from @amjith:

> So the reason for having postgresql as a dependency is because otherwise the formula won’t pass in their pull request tests. You’re welcome to remove it and give it a shot. Things have changed with homebrew since I originally created that formula.
[...]

PR #727, opened 2017-06-02
Update Homebrew install instructions

tsroten:
## Description

Updates the macOS installation instructions to use dbcli's Homebrew tap.

This address #718.

## Checklist

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -23,7 +23,7 @@ If you already know how to install python packages, then you can simply do:
 
     or
 
-    $ brew install pgcli  # Only on OS X
+    $ brew tap-pin dbcli/tap && brew install pgcli  # Only on macOS
 
 If you don't know how to install python packages, please check the 
 `detailed instructions`__.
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -92,22 +92,19 @@ My email: amjith.r@gmail.com, Twitter: `@amjithr <http://twitter.com/amjithr>`_
 Detailed Installation Instructions:
 -----------------------------------
 
-OS X:
-=====
+macOS:
```
j-bennet: Apple seems to be going back and forth between macOS and OS X. :) Should we maybe say `macOS/OS X`?

tsroten: Haha, I think at 10.12 they officially changed the name to macOS so it matches their other products. OS X doesn't refer to their current operating system anymore.

But, if you have 10.11 or below, it's still OS X -- confusing!

I'm up for whichever. Personally, I think a single name is better (less reading, more clear, etc.).

j-bennet: Yes, let's stick with the latest name for now!

Issue #37, opened 2015-01-06
Read socket configuration information


scw:
When connecting to a local database configured with a local socket, `pgcli` fails:

```
pgcli -U postgres data_entry

    could not connect to server: No such file or directory
    Is the server running locally and accepting
    connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5432"?
```

The equivalent call to `psql` succeeds:

```
psql -U postgres data_entry
psql (8.4.22)
Type "help" for help.

data_entry#
```

Running `strace` against `psql` shows that it reads the socket configuration directory from the `unix_socket_directory` of the postgresql.conf file. I can get `pgcli` to run by specifying `-h /tmp` which is the socket location on this distribution (Ubuntu 12.04.5), perhaps because this Postgres release is backported. The psql code executes [this search](https://github.com/jbq/postgresql-common/blob/master/PgCommon.pm#L354-381) to find the appropriate path.

Some defaults for sockets on various distributions and OSes:
- "/var/run/postgresql" # Debian default
- "/var/pgsql_socket" # MacOS default
- "/usr/local/var/postgres" # FreeBSD default; custom installation
- "/tmp" # custom installation

May be worth just iterating over these for an appropriate socket directory before failing with a blank 'host' parameter.

amjith:
Awesome bug report! :)

When I fall back to empty string for host, I'm relying to `psycopg2` to do the right thing. But I'll see if I can iterate over the common locations to find the socket before doing that.

SpotlightKid:
On a related note, if you specify a port with the `-p` command-line option, but no host with the `-h` option, `pgcli` still tries to use a unix domain socket. Maybe it should assume an inet socket and `localhost` as the default hostname in that case, since specifying a port doesn't make sense with unix domain sockets.

amjith:
Good point. Added #229

demidovakatya:
Hey there, I have the same error (os x, `pgcli` v.1.5.1, installed via brew, `postgresql` v.9.6.2)
```
pgcli
	could not connect to server: No such file or directory
	Is the server running locally and accepting
	connections on Unix domain socket "/tmp/.s.PGSQL.5432"?
```

demidovakatya:
upd.: [this](http://dba.stackexchange.com/questions/21587/postgresql-is-running-locally-but-i-cannot-connect-why) helped

```
sudo mkdir /var/pgsql_socket
sudo chown -R whoami /var/pgsql_socket
[...]

PR #733, opened 2017-06-14
Search socket dir before falling back to psycopg2 guess

bersace:
Closes: #37 

As suggested in the description of the issue, I just use the first existing socketdir found from a list of well known candidates. This should be enough.

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -292,9 +292,6 @@ class PGCli(object):
         pgspecial_logger.addHandler(handler)
         pgspecial_logger.setLevel(log_level)
 
-    def connect_dsn(self, dsn):
-        self.connect(dsn=dsn)
-
     def connect_uri(self, uri):
         uri = urlparse(uri)
         database = uri.path[1:]  # ignore the leading fwd slash
@@ -854,9 +851,9 @@ def cli(database, username_opt, host, port, prompt_passwd, never_prompt,
     elif '://' in database:
         pgcli.connect_uri(database)
     elif "=" in database:
-        pgcli.connect_dsn(database)
+        pgcli.connect(dsn=database)
     elif os.environ.get('PGSERVICE', None):
-        pgcli.connect_dsn('service={0}'.format(os.environ['PGSERVICE']))
+        pgcli.connect(dsn='service={0}'.format(os.environ['PGSERVICE']))
     else:
         pgcli.connect(database, host, user, port)
 
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -316,9 +316,23 @@ class PGCli(object):
         # unquote str(each URI part (they may be percent encoded)
         self.connect(**arguments)
 
+    def guess_socketdir(self):
+        candidates = [
+            '/var/run/postgresql',
+            '/var/pgsql_socket',
+            '/usr/local/var/postgres',
+            '/tmp',
+        ]
+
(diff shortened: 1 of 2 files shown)


pgcli/main.py
```diff
@@ -316,9 +316,23 @@ def fixup_possible_percent_encoding(s):
...
+
+        for candidate in candidates:
+            if os.path.isdir(candidate):
+                return candidate
+
     def connect(self, database='', host='', user='', port='', passwd='',
                 dsn='', **kwargs):
         # Connect to the database.
+        if not host:
+            host = self.guess_socketdir() or ''
```
bersace: Actually, I do not like mixing param guess inside `connect` function. This is misleading. I suggest writing a `guess_dsn` function that do all the guess I/O, env var fallback, etc. and let `connect` just trigger the connection or fail.

j-bennet: @bersace Feel free to extract the "guesswork" into a separate method. You're right that the `connect` method is becoming difficult to read.

bersace: I prefer working on this in another PR :-)

bersace:
@j-bennet Ready for merge.

bersace:
🆙 @j-bennet

amjith:
@bersace This change made my connection experience worse. I connect to my local instance via localhost:5432. But this change give preference to socket over port. 

Previously I used to type `pgcli` and nothing else it will correctly connect to postgres. Now if I type `pgcli` I get the following error: 

```
could not connect to server: No such file or directory
        Is the server running locally and accepting
        connections on Unix domain socket "/usr/local/var/postgres/.s.PGSQL.5432"?
```

bersace:
@amjith That's not the goal ^^

What is the behaviour of psql for you ? Do you want `pgcli` to glob for `.PGSQL*` socket ?

For the sharing, I run simple docker compose file and use `direnv` to setup `PGHOST` when I cd to `pgcli` directory.

``` yaml
# cat /my-docker-compose.yml
version: '2'

services:
  pg96:
    image: postgres:9.6
```

``` bash
# cat .envrc
export PGHOST=postgres.docker
export PGUSER=postgres
. ${WORKON_HOME}/pgcli/bin/activate
export COMPOSE_FILE=my-docker-compose.yml
```

amjith:
Right now the behavior pgcli mimics the behavior of psql. I don't think there is a standard name for postgres socket files.

I'm inclined to say that we stick with psycopg2's order of detection to connect to a database and not do anything fancy ourselves.

j-bennet:
I think this PR was caused by https://github.com/dbcli/pgcli/issues/37, where `pgcli` failed to connect to a socket, while `psql` did the right thing. It would be good to fix that. I think it should be possible to check for being able to connect on those sockets first, and if that fails, delegate to psycopg2 to do the rest.

j-bennet:
So I verified the current `pgcli` vs `psql` behavior, and it seems to me that pgcli actually does the right thing. It can connect to the UNIX socket if it is correctly specified in `postgresql.conf` as `unix_socket_directory` or `unix_socket_directories`. If it's not correctly specified there, then `psql` would not be able to connect either (see https://dba.stackexchange.com/questions/21587/postgresql-is-running-locally-but-i-cannot-connect-why).

I think this change is not needed, sorry @bersace.

amjith:
Closing this in favor of existing behavior.

Issue #617, opened 2016-11-26
Numeric output is not showed correctly


MrOutis:
I'm using `pgcli` version: 1.3.1

This is the output using `pgcli`:
![1480184973](https://cloud.githubusercontent.com/assets/7363250/20642431/2cf60846-b3d4-11e6-9f0e-cf58c174da7b.png)

And the output for the same query, using `psql`:
![1480184890](https://cloud.githubusercontent.com/assets/7363250/20642430/289aee7e-b3d4-11e6-8b1d-ffa81e8990fd.png)

------------------------

This is a better test that anyone can replicate:

`pgcli`:
![1480185134](https://cloud.githubusercontent.com/assets/7363250/20642444/9a85ebe2-b3d4-11e6-8c2a-67ff6d04905b.png)

`psql`:
![1480185172](https://cloud.githubusercontent.com/assets/7363250/20642449/a245f20a-b3d4-11e6-8642-bb71107c0ec6.png)

------------------------

The behavior is new; at least 4 months ago it was working as intended.

MrOutis:
I didn't found any relevant commit that could change that behaivour, just this one: https://github.com/dbcli/pgcli/pull/591

j-bennet:
Formatting in `pgcli` is done by `tabulate`. `pgcli` keeps its own slightly tweaked version of tabulate [here](https://github.com/dbcli/pgcli/blob/master/pgcli/packages/tabulate.py). Tabulate tries to be smart about formatting numbers.

```
>>> print tabulate.tabulate([['x'], [1.00000000]], headers='firstrow')[0]
  x
---
  1
```

MrOutis:
Thank you, @j-bennet, I'll give it a check.

j-bennet:
I'm not sure what the good solution is here, have to think about it. In a lot of cases, "simplifying" numbers like tabulate does is a good thing, but not when you explicitly format it to certain precision.

MrOutis:
Maybe putting an option in the configuration file to turn off _numeric_simplification_.

MrOutis:
Also, it would be nice to have a bullet about this simplification in the [features section](https://github.com/dbcli/pgcli#features).

evanslify:
Just spend a couple of hours trying to figure out why values become rounded. This option (or at least explicitly written on README.md) could be helpful!

MrOutis:
👍 thanks!

shaib:
[...]

PR #735, opened 2017-06-19
Use CLI Helpers for output formatting

tsroten:
## Description

This pull request moves all output formatting to CLI Helpers. It fixes https://github.com/dbcli/pgcli/issues/617.

It removes:
- the packaged tabulate library.
- the expanded output module.

One other change is that the expanded output starts the record numbers at 1 instead of 0, which actually makes it behave like psql. So, instead of:
`-[ RECORD 0 ]-------------------------`
you see:
`-[ RECORD 1 ]-------------------------`

~~**Note: The tests will fail until CLI Helpers v2.0 is released.**~~

## Checklist


pgcli/main.py
```diff
@@ -156,6 +157,7 @@ def __init__(self, force_passwd_prompt=False, never_passwd_prompt=False,
         self.now = dt.datetime.today()
 
         self.completion_refresher = CompletionRefresher()
+        self.formatter = TabularOutputFormatter()
```
amjith: I'm mildly curious. Is there a reason for this to be an instance variable? 

Why not instantiate the TabularOutputFormatter() inside the format_output() and use it?

tsroten: Not really, I'll change it 😄

tsroten: Ok, done. 👍

pgcli/main.py
```diff
@@ -757,6 +759,48 @@ def get_last_query(self):
         """Get the last query executed or None."""
         return self.query_history[-1][0] if self.query_history else None
 
+    def format_output(self, title, cur, headers, status, settings):
```
amjith: With the new change it looks like format_output doesn't need to be part of the PGCli class. The self variable is not used anywhere. 

We can either make it a @staticmethod if we think that's appropriate or we can move it outside of this class. What do you think?

tsroten: Good point. I'll move it out of the class and update the tests.

tsroten: Done 😄

amjith:
This looks great!! Fantastic work. 👏  🍰

PR #742, opened 2017-07-01
Fix link

lwm:


j-bennet:
@lwm Thanks for making pgcli better!

PR #744, opened 2017-07-07
Better * expansion for inserts

koljonen:
## Description
When expanding * for `INSERT INTO tbl(*`, skip columns that have a default value of either some sequence (`nextval(...`) or of `now()`.

## Checklist

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -7,6 +7,7 @@ Upcoming
 * Only set `LESS` environment variable if it's unset. (Thanks: `Irina Truong`_)
 * Quote schema in `SET SCHEMA` statement (issue #469) (Thanks: `Irina Truong`_)
 * Use CLI Helpers for pretty printing query results (Thanks: `Thomas Roten`_).
+* Skip serial columns when expanding * for `INSERT INTO foo(*` (Thanks: `Joakim Koljonen`_).
 
 1.6.0
 =====
--- a/pgcli/packages/parseutils/meta.py
+++ b/pgcli/packages/parseutils/meta.py
@@ -1,6 +1,19 @@
 from collections import namedtuple
 
-ColumnMetadata = namedtuple('ColumnMetadata', ['name', 'datatype', 'foreignkeys'])
+_ColumnMetadata = namedtuple(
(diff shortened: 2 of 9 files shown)


amjith:
Should this skip columns of type SERIAL as well? Right now I see `id` being suggested by `id` is of type `SERIAL` which should be auto-incremented.

amjith:
How do I trigger this feature? Here's what I'm seeing: 

1__dbcli_0_python2_7_-__pgcli___users_amjith_code_python_pgcli____tmux__and_postgresql_-_add_timestamp_column_with_default_now___for_new_rows_only_-_stack_overflow


I have a column with default set to now() and a column that has a nextval() but both of them are suggested by completion.

koljonen:
@amjith I think `serial` columns are just syntactic sugar for `integer` with a `nextval(...` default, so they should be covered? I get these results on 9.1.24 and 9.6.2:
```
vagrant@vagrant:vagrant> CREATE TABLE bar(id serial);
........................ SELECT att.atttypid::regtype, def.adsrc
........................ FROM pg_attribute att
........................ LEFT OUTER JOIN pg_attrdef def
........................     ON def.adrelid = att.attrelid
........................     AND def.adnum = att.attnum
........................ WHERE att.attrelid = 'bar'::regclass AND att.attname = 'id';
NOTICE:  CREATE TABLE will create implicit sequence "bar_id_seq" for serial column "bar.id"

CREATE TABLE
╒════════════╤═════════════════════════════════╕
│ atttypid   │ adsrc                           │
╞════════════╪═════════════════════════════════╡
│ integer    │ nextval('bar_id_seq'::regclass) │
╘════════════╧═════════════════════════════════╛
SELECT 1
```

koljonen:
@amjith You enter `INSERT INTO test (*` and trigger completion.

amjith:
Awesome! That works as expected. 👍 

Reading through the code now.

amjith:
Is there an advantage in even suggesting the column that have `nextval()` and `now()` in the completion menu?

amjith:
Code looks good. Merging. 👍

PR #756, opened 2017-07-21
Drop ipython integration

bersace:
Hi,

This code has undefined variables. It's guaranteed to fail. There is no test nor documentation. I suggest to drop it and let's us use history if ever the use case arise.


j-bennet:
Hi @bersace, this was added by @darikg and it seems he described it quite well in his [original PR](https://github.com/dbcli/pgcli/pull/91) and [this one](https://github.com/dbcli/pgcli/pull/104). I'm not sure if @darikg has some bandwidth to document it and finish up the TODO items he had. If the feature works though, I think it's useful, and we should document it and test-case it, rather than dropping it.

bersace:
@j-bennet that's ok. I guess that `get_ipython` is implicitly imported when loaded by `ipython`.

PR #758, opened 2017-07-24
Amjith/remove continuation char

amjith:
## Description
Change the continuation char from `....` to empty string. This allows folks to copy and paste sql statements easily from the REPL.

## Checklist
~- [ ] I've added my name to the `AUTHORS` file (or it's already there).~

### DIFF
--- a/pgcli/pgclirc
+++ b/pgcli/pgclirc
@@ -121,7 +121,7 @@ prompt = '\u@\h:\d> '
 min_num_menu_lines = 4
 
 # Character used to left pad multi-line queries to match the prompt size.
-multiline_continuation_char = '.'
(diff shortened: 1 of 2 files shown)


pgcli/pgclirc
```diff
@@ -121,7 +121,7 @@ prompt = '\u@\h:\d> '
 min_num_menu_lines = 4
 
 # Character used to left pad multi-line queries to match the prompt size.
-multiline_continuation_char = '.'
+multiline_continuation_char = ''
```
bersace: Good idea to be copy-pastable. I use `Ctrl+select` to select region in vte terminals.

That's not obvious that empty char will produce space padded prompt. I thought that this would drop padding.

PR #761, opened 2017-07-25
Behave tests should also accept pgport.

j-bennet:
## Description

Fixes the problem with behave tests disregarding PGHOST and PGPORT. This is needed if using a dockerized postgres instance for testing.

Example command that did not work before:

`PGHOST=localhost PGUSER=postgres PGPORT=8432 behave`

(note the non-standard port because it's forwarded from the docker container).

## Checklist

- ~[ ] I've added this contribution to the `changelog.md`~.

### DIFF
--- a/tests/features/db_utils.py
+++ b/tests/features/db_utils.py
@@ -6,17 +6,17 @@ from psycopg2 import connect
 from psycopg2.extensions import AsIs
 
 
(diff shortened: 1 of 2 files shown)


fpietka:
@j-bennet fixes the problem for me. You just have a slight lint issue on a doc block.

PR #764, opened 2017-07-27
Use raw strings where needed to avoid backslash ambiguity

fpietka:
## Description
This is rather not really an issue as it is, but this changes allow to avoid ambiguity in slashes interpretation in strings (W1401).

## Checklist
- [ ] ~~I've added my name to the `AUTHORS` file (or it's already there).~~


amjith:
⚽

PR #771, opened 2017-08-06
Don't quote identifiers that are non-reserved keywords

koljonen:
## Description
Keep track of which keywords are reserved, and quote identifiers that coïncide with those. 
The list of reserved words is from https://www.postgresql.org/docs/10/static/sql-keywords-appendix.html.

## Checklist


amjith:
🍅

PR #772, opened 2017-08-06
Better formatting for array output

koljonen:
## Description
Before:
```
vagrant@vagrant:vagrant> SELECT
                                 array[1, 2, 3]::bigint[] as bigint_array,
                                 '{{1,2},{3,4}}'::numeric[] as nested_numeric_array,
                                 '{å,魚,текст}'::text[] as 配列
                             UNION ALL
                             SELECT '{}', NULL, array[NULL]
+----------------+--------------------------------------------------------------+---------------------------------------------------------+
| bigint_array   | nested_numeric_array                                         | 配列                                                    |
|----------------+--------------------------------------------------------------+---------------------------------------------------------|
| [1L, 2L, 3L]   | [[Decimal('1'), Decimal('2')], [Decimal('3'), Decimal('4')]] | [u'\xe5', u'\u9b5a', u'\u0442\u0435\u043a\u0441\u0442'] |
| []             | NULL                                                         | [None]                                                  |
+----------------+--------------------------------------------------------------+---------------------------------------------------------+
```
After:
```
vagrant@vagrant:vagrant> SELECT
                                 array[1, 2, 3]::bigint[] as bigint_array,
                                 '{{1,2},{3,4}}'::numeric[] as nested_numeric_array,
                                 '{å,魚,текст}'::text[] as 配列
                             UNION ALL
                             SELECT '{}', NULL, array[NULL]
+----------------+------------------------+--------------+
| bigint_array   | nested_numeric_array   | 配列         |
|----------------+------------------------+--------------|
| {1,2,3}        | {{1,2},{3,4}}          | {å,魚,текст} |
| {}             | NULL                   | {NULL}       |
+----------------+------------------------+--------------+
```

## Checklist


koljonen:
@tsroten You've been pretty involved in these parts of the code, so mind having a look?

fpietka:
@koljonen I think you forgot the tests here, some are failing. Or is it a WIP?

pgcli/main.py
```diff
@@ -946,14 +946,27 @@ def format_output(title, cur, headers, status, settings):
     case_function = settings.case_function
     formatter = TabularOutputFormatter(format_name=table_format)
 
+    def format_array(val):
+        if val is None:
+            return settings.missingval
+        if not isinstance(val, list):
+            return val
+        return '{' + ','.join(unicode(format_array(e)) for e in val) + '}'
```
fpietka: `unicode()` is Python2 only. This will fail with Python3.

koljonen: Thanks, pushed a fix for that.

tsroten:
@koljonen Sure, I'll take a look tonight 😄

pgcli/main.py
```diff
@@ -64,6 +64,9 @@
 
 from collections import namedtuple
 
+if sys.version_info[0] > 2:
+    unicode = str
```
tsroten: I think there is already a Python 2/3 check in `pgcli/encodingutils.py`.

So, you might consider doing what we do at mycli and CLI Helpers. In a different module, handle the Python 2 vs. 3 logic so that it doesn't end up in `main.py`. Then, you just import `text_type` (nothing magical about that variable name, it's just a popular name choice that many projects use). Here it is in mycli:
https://github.com/dbcli/mycli/blob/master/mycli/encodingutils.py#L9

I think it's good to stay away from using the standard library names (e.g. `unicode` and `str`) so as to prevent confusion about what class we're actually referencing.

koljonen: Sure, makes sense.

pgcli/main.py
```diff
@@ -946,14 +949,27 @@ def format_output(title, cur, headers, status, settings):
...
     formatter = TabularOutputFormatter(format_name=table_format)
 
+    def format_array(val):
+        if val is None:
+            return settings.missingval
+        if not isinstance(val, list):
+            return val
+        return '{' + ','.join(unicode(format_array(e)) for e in val) + '}'
+
+    def format_arrays(data, headers, column_types, **kwargs):
```
tsroten: Since we're not using `column_types` or `kwargs`, it might be more appropriate to make it clear we aren't using any other arguments via the variable name `_`:
```python
def format_arrays(data, headers, **_):
```

tests/test_main.py
```diff
@@ -61,12 +117,30 @@ def test_format_output_auto_expand():
...
+        'Title',
+        [('abc', 'def')],
+        ['head1', 'head2'],
+        'test status',
+        settings._replace(max_width=1)
+    )
     expanded = [
-        'Title', u'-[ RECORD 1 ]-------------------------\nhead1 | abc\nhead2 | def\n', 'test status']
+        'Title',
+        u'-[ RECORD 1 ]-------------------------\n'
```
tsroten: It can sometimes lead to odd debugging situations when we mix unicode and byte strings. So, instead of sometimes using the `u` prefix and sometimes not, how about we just do `from __future__ import unicode_literals` at the top of the module so everything is unicode?

tsroten:
@koljonen The new output looks much better 👍 .

The tests are failing. I'll have time to help look at it a bit later tonight. Good work so far!

koljonen:
As for the tests failing, I'm rather confused. They're passing for me on this setup: `platform darwin -- Python 2.7.10, pytest-2.9.1, py-1.4.31, pluggy-0.3.1`. I've never had the tests work locally and then fail on Travis's 2.7 before. The Travis test output looks like the rows are missing from the result sets.

tsroten:
@koljonen So, in CLI Helpers, generators are passed around internally, so `row[:]` is not actually assigning the row back into `data`.

I'll push an update to the branch that should fix it.

tsroten:
Ok, tests are passing, now there is a minor lint issue

tsroten:
@koljonen Ok, I think the only thing left is the lint issue: https://travis-ci.org/dbcli/pgcli/jobs/261656805#L673

tsroten:
@koljonen Thanks for your work on this! The new array output is so much better than the old 👍 .

PR #773, opened 2017-08-06
WIP Add \conninfo command

### LINKED ISSUE #716: Allow \c to take in more options and add support for \conninfo
Copied from the mailing list.

## Description

I am having difficulty connecting to a different DB as a different user after successfully connecting to postgres.  psql exhibits the behavior I expect, however pgcli does not.

1. connecting to new database with new user,  same host, port, etc

From Postgres documentation, 
https://www.postgresql.org/docs/9.6/static/app-psql.html

\c or \connect [ -reuse-previous=on|off ] [ dbname [ username ] [ host ] [ port ] | conninfo ]
Establishes a new connection to a PostgreSQL server. The connection parameters to use can be specified either using a positional syntax, or using conninfo connection strings as detailed in Section 32.1.1.

Where the command omits database name, user, host, or port, the new connection can reuse values from the previous connection. By default, values from the previous connection are reused except when processing a conninfo string. Passing a first argument of -reuse-previous=on or -reuse-previous=off overrides that default. When the command neither specifies nor reuses a particular parameter, the libpq default is used. Specifying any of dbname, username, host or port as - is equivalent to omitting that parameter.

If the new connection is successfully made, the previous connection is closed. If the connection attempt failed (wrong user name, access denied, etc.), the previous connection will only be kept if psql is in interactive mode. When executing a non-interactive script, processing will immediately stop with an error. This distinction was chosen as a user convenience against typos on the one hand, and a safety mechanism that scripts are not accidentally acting on the wrong database on the other hand.

sah209: ~$ psql -p 5436 postgres postgres
Password for user postgres: 
psql (9.6.2)
Type "help" for help.

postgres=# \conninfo
You are connected to database "postgres" as user "postgres" via socket in "/var/run/postgresql" at port "5436".

postgres=# \c account_bill account_bill
Password for user account_bill: 
You are now connected to database "account_bill" as user "account_bill".

account_bill=> \conninfo
You are connected to database "account_bill" as user "account_bill" via socket in "/var/run/postgresql" at port "5436".

sah209: ~$ pgcli -p 5436 postgres postgres
Password: 
Version: 1.5.1
Home: http://pgcli.com
postgres@(none):postgres> 

postgres@(none):postgres> \c account_bill 
You are now connected to database "account_bill" as user "postgres" 

But I want to connect as -U account_bill

postgres@(none):account_bill> \c account_bill account_bill
Connection reset. Reconnect (Y/n): Y
Reconnected!
Try the command again.

Any suggestions on what I am doing wrong or if this is not yet implemented would be appreciated.

2.  Is there an pgcli equivalent to psql's \conninfo  ?

pgcli ->

postgres@(none):account_bill> \conninfo
syntax error at or near "\"
LINE 1: \conninfo
        ^

## Your environment

pgcli --version 
Version: 1.5.1

OS:
Distributor ID: Ubuntu 
Description: Ubuntu 16.04.2 LTS 
Release: 16.04 
Codename: xenial

Postgres version = 9.6.2-1.pgdg16.04+1
host = localhost
port = 5436

fpietka:
## Description
Fix #716 

## Checklist

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -203,6 +203,15 @@ class PGCli(object):
                                 'Execute commands from file.')
         self.pgspecial.register(self.write_to_file, '\\o', '\\o [filename]',
                                 'Send all query results to file.')
+        self.pgspecial.register(self.info_connection, '\\conninfo',
+                                '\\conninfo', 'Get connection details')
+
+    def info_connection(self, **_):
+        yield (None, None, None, 'You are connected to database "%s" as user '
+               '"%s" on host "%s" at port "%s".' % (self.pgexecute.dbname,
+                                                    self.pgexecute.user,
+                                                    self.pgexecute.host,
+                                                    self.pgexecute.port))
 
     def change_db(self, pattern, **_):
         if pattern:


fpietka:
So far it kinda work, except the error handling of the `\c` command I'm not happy with.

pgcli/main.py
```diff
@@ -203,11 +203,27 @@ def register_special_commands(self):
...
+               '"%s" on host "%s" at port "%s".' % (self.pgexecute.dbname,
+                                                    self.pgexecute.user,
+                                                    self.pgexecute.host,
+                                                    self.pgexecute.port))
 
     def change_db(self, pattern, **_):
         if pattern:
-            db = pattern[1:-1] if pattern[0] == pattern[-1] == '"' else pattern
-            self.pgexecute.connect(database=db)
+            infos = re.findall(r'"[^"]*"|[^"\'\s]+', pattern)
```
j-bennet: A comment would help to explain what you're capturing with this regex.

pgcli/main.py
```diff
@@ -203,11 +203,27 @@ def register_special_commands(self):
...
+            infos.extend([None] * (4 - len(infos)))
+            map(lambda s: s.strip('"'), infos)
+            db, user, host, port = infos
+            try:
+                self.pgexecute.connect(database=db, user=user, host=host,
+                                       port=port)
+            except Exception as e:
+                print("Previous connection kept")
```
j-bennet: You probably want to log this exception here also. And instead of `print`, better use `click.echo` (or `click.secho` if you want to style the error message).

fpietka:
@j-bennet thanks for your input, it was really helpful.

fpietka:
For `\conninfo` I couldn't find a way of getting socket directory (if connected through socket) with psycopg2.

j-bennet:
According to the docs [here](https://www.postgresql.org/docs/current/static/functions-info.html), if the user is connected via unix socket, functions `inet_client_addr()`, `inet_client_port()`, `inet_client_port()`, `inet_server_port()` will all return empty.

Once you know that, you can retrieve `unix_socket_directories` setting with a query:

```
postgres=# select name, setting from pg_settings where name = 'unix_socket_directories';
          name           | setting
-------------------------+---------
 unix_socket_directories | /tmp
(1 row)
```

fpietka:
Nice! Wasn't aware of the second part. I'll do the changes later today. Thanks Le 7 août 2017 7:52 AM, Irina Truong <notifications@github.com> a écrit :According to the docs here, if the user is connected via unix socket, functions inet_client_addr(), inet_client_port(), inet_client_port(), inet_server_port() will all return empty.
Once you know that, you can retrieve unix_socket_directories setting with a query:
postgres=# select name, setting from pg_settings where name = 'unix_socket_directories';
          name           | setting
-------------------------+---------
 unix_socket_directories | /tmp
(1 row)

fpietka:
@j-bennet thanks for the query, I made the changes

pgcli/pgexecute.py
```diff
@@ -215,6 +221,9 @@ def connect(self, database=None, user=None, password=None, host=None,
         self.host = host
         self.port = port
 
+        if not self.host:
+            self.host = self.socket_directory()
```
j-bennet: If you assign socket directory to `self.host`, it makes `self.host` mean different things at different times. It may also interfere with the current behavior if you mess with `self.host` - possibly affect reconnecting on timeout. I think it's better if we have no host assigned when connected via socket.

pgcli/pgexecute.py
```diff
@@ -491,6 +500,13 @@ def full_databases(self):
             headers = [x[0] for x in cur.description]
             return cur.fetchall(), headers, cur.statusmessage
 
+    def socket_directory(self):
```
j-bennet: I would suggest renaming the method to `get_socket_directory`, and adding a new property `socket_directory` to pgexecute class, that would be populated if host is empty.

pgcli/pgexecute.py
fpietka: I assumed that since it was postgres behavior, it would be an issue: https://www.postgresql.org/docs/9.6/static/libpq-connect.html#LIBPQ-CONNECT-HOST

But if you feel more confortable, we can separate both. It'll require some work since when you use `-h /socket_dir` you'll have a socket directory in the `host` property already.

My changes are there mostly to handle default behavior, when no host is passed to the connect method.

j-bennet: I didn't know that postgresql does that. Can you verify that you'd be able to connect with psycopg2 if you provide `host=/tmp`, and that reconnect behavior doesn't suffer when you run pgcli without providing hostname? If yes, we can keep that behavior.

pgcli/pgexecute.py
fpietka: I did a fixup for that.

pgcli/pgexecute.py
fpietka: When calling `pgcli with -h /tmp` it does exactly that.
As for reconnect, I came across a problem (even on master) as when I restart the database `psycopg2.OperationalError.pgcode` is `None`. This wasn't handled resulting in a `KeyError` exception within psycopg2.

Now everything is working fine.

fpietka:
Reference to pgcode being sometimes `None`: http://initd.org/psycopg/docs/module.html#psycopg2.Error.pgcode

j-bennet:
Thank you! 🍨

PR #775, opened 2017-08-06
Add COLUMN as possible keyword after DROP

### LINKED ISSUE #769: 'COLUMN' is not suggested after 'DROP' in 'ALTER TABLE' query
## Description
PG's [`ALTER TABLE`](https://www.postgresql.org/docs/9.6/static/sql-altertable.html) supports dropping individual columns with

```
ALTER TABLE foo DROP COLUMN bar;
```

where `bar` is the name of a column on the table `foo` but pgcli does not suggest `COLUMN` after `DROP`:

screen shot 2017-08-05 at 7 26 49 pm

## Your environment

macOS 10.12.4 (16E195)
pgcli 1.7.0
Installed via brew: Poured from bottle on 2017-08-05 at 12:43:38

fpietka:
## Description
Fix #769
Add missing keyword COLUMN after DROP.

## Checklist

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -11,6 +11,7 @@ Bug Fixes:
 ----------
 
 * Fix the way we get host when using DSN (issue #765) (Thanks: `François Pietka`_)
+* Add missing keyword COLUMN after DROP (issue #769) (Thanks: `François Pietka`_)
 
 1.7.0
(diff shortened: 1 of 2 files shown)


amjith:
🚀

PR #777, opened 2017-08-08
use cursor from psycopg in pgcli

meeuw:
## Description
use cursor from psycopg in pgcli

I'm not sure if you should/can use hardcoded numbers to match the `type_code` of Cursor.description

## Checklist

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -63,6 +63,7 @@ except ImportError:
 
 from getpass import getuser
 from psycopg2 import OperationalError, InterfaceError
+import psycopg2
 
 from collections import namedtuple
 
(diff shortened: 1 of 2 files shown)


pgcli/main.py
```diff
@@ -984,13 +984,24 @@ def format_arrays(data, headers, **_):
...
+        if max_width is not None:
+            cur = list(cur)
+        column_types = None
+        if hasattr(cur, 'description'):
+            column_types = []
+            for d in cur.description:
+                if d[1] == 23:
+                    column_types.append(int)
+                elif d[1] == 1700:
+                    column_types.append(float)
```
fpietka: `psycopg2.extensions.cursor` has a method `cast()` that can be used with oid's. I didn't find a method to get the type itself though, but with those magic numbers you might also miss some types.

pgcli/main.py
```diff
@@ -984,13 +984,24 @@ def format_arrays(data, headers, **_):
...
+                else:
+                    column_types.append(type(""))
```
tsroten: The variable `text_type` is available in pgcli's `main.py`, so you might try using that instead.

meeuw: fixed, thanks

pgcli/main.py
meeuw: @fpietka I've changed it to some (as far as I know) undocumented constants. Do you think this is better?

fpietka: I found it more explicit.
Kudos for finding those (I didn't, as there is nothing in the documentation).

j-bennet:
@meeuw Looks like this is ready to merge?

meeuw:
Sure, I'm sure the undocumented `psycopg2.extensions.*` constants will cause a regression but I hope that will give us a suggesion how to do this properly. :crossed_fingers:

j-bennet:
Hmm. I suppose we could pin `psycopg2` to a specific version, but I'd rather not do that. We can fix it when it breaks.

j-bennet:
🌸

PR #786, opened 2017-09-10
Drop wcwidth dependency

meeuw:
## Description
Drop wcwidth dependency (this is only used in cli_helpers which is a dependency for pgcli)

## Checklist


j-bennet:
👍

PR #794, opened 2017-09-24
Coloured and customizable boolean output

koljonen:
## Description
Similar to the configurable NULL output @avdd added last year, this adds two config settings, `false_string` and `true_string`, which control how bools are displayed. The defaults are a red "false" and a green "true", respectively.

## Checklist

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -1,7 +1,7 @@
 Upcoming
 ========
 
-TODO
+* Colored and customizable booleans in output (Thanks: `Joakim Koljonen`_)
 
 1.8.1
 =====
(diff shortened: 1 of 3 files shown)


amjith:
@koljonen I've been reviewing the PR and I have a few concerns about the way the coloring in implemented. 

I'd prefer to use a more standard way of coloring the output, say using the pygments library instead of using the terminal escape keys. They are not portable and they make it cryptic for the user to change in the future.

You've seen how we provide a way for the user to override the colors of our drop down menu in the config file (see [here](https://github.com/dbcli/pgcli/blob/master/pgcli/pgclirc#L131)). I would prefer to take that approach.

I know that @tsroten has implemented a similar concept for mycli. https://github.com/dbcli/mycli/pull/462. It would be nice to match the behavior. 

Thoughts?

tsroten:
I wonder if we should modify the [`style_output` preprocessor](https://github.com/dbcli/cli_helpers/blob/master/cli_helpers/tabular_output/preprocessors.py#L145) in CLI Helpers to be more generic and be able to handle field-by-field coloring instead of just row-by-row coloring.

PR #796, opened 2017-10-03
Use termporary config in tests

oxitnik:
## Description
Use temporary directory as config directory.

## Checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -34,4 +34,6 @@ def exception_formatter():
 
 @pytest.fixture(scope="session", autouse=True)
 def temp_config(tmpdir_factory):
+    # this function runs on start of test session.
+    # use temporary directory for config home so user config will not be used
     os.environ['XDG_CONFIG_HOME'] = str(tmpdir_factory.mktemp('data'))
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,3 +1,4 @@
+import os
 import pytest
 from utils import (POSTGRES_HOST, POSTGRES_USER, POSTGRES_PASSWORD, create_db, db_connection,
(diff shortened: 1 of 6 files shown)


tests/conftest.py
```diff
@@ -29,3 +30,8 @@ def executor(connection):
 @pytest.fixture
 def exception_formatter():
     return lambda e: str(e)
+
+
+@pytest.fixture(scope="session", autouse=True)
+def temp_config(tmpdir_factory):
```
j-bennet: Where is this fixture being used?

tests/features/environment.py
```diff
@@ -96,6 +98,10 @@ def before_all(context):
 
     context.fixture_data = fixutils.read_fixture_files()
 
+    context.env_config_home = tempfile.mkdtemp(prefix='pgcli_home_')
+
+    os.environ['XDG_CONFIG_HOME'] = context.env_config_home
```
j-bennet: Do you have to restore this variable back to previous value after tests are run?

tests/conftest.py
oxitnik: It is used automatically on session start

tests/features/environment.py
oxitnik: I don't see why. Env vars process-only anyway.

j-bennet:
@oxitnik This looks fine, but it look me some time to understand what's going on. Please add some comments to make it easier for other people working with this code later.

oxitnik:
@j-bennet  Added some comments. Also made XDG_CONFIG_HOME variable to be restored after test.

j-bennet:
Very nice. Thanks!

🍫

Issue #797, opened 2017-10-06
Issuing notee command writes an error to output file


j-bennet:
Here is a `pgcli` session output when setting an output file name to `zzz`, writing some output and issuing `notee`:

```
(pgcli) --- src/pgcli» pgcli world
irina@/tmp:world> \o zzz
Writing to file "/Users/irina/src/pgcli/zzz"
Time: 0.001s
irina@/tmp:world> select 'foo'
Time: 0.001s
irina@/tmp:world> notee
Time: 0.001s
irina@/tmp:world>
Goodbye!
```

Let's see what we wrote to file:

```
(pgcli) --- src/pgcli» cat zzz
select 'foo'
+------------+
| ?column?   |
|------------|
| foo        |
+------------+
SELECT 1

notee
syntax error at or near "notee"
LINE 1: notee
        ^
```

pgcli errors out on `notee` here:

```
notee
syntax error at or near "notee"
LINE 1: notee
```

Issue #795, opened 2017-09-25
Build constantly breaks on the same test case in end-to-end tests


j-bennet:
pgcli build breaks on `tee` test case all the time:

https://github.com/dbcli/pgcli/blob/915ffb70a70a855bfc8cbc2d3b7ab9828e102775/tests/features/iocommands.feature#L10-L17

We need to figure out how to make that test better.

meeuw:
Maybe we should raise all timeouts to 5 in:
https://github.com/dbcli/pgcli/blob/915ffb70a70a855bfc8cbc2d3b7ab9828e102775/tests/features/steps/iocommands.py

j-bennet:
I haven't seen this one lately. Closing for now.

PR #798, opened 2017-10-06
Fix for tests breaking after tee.

j-bennet:
## Description

This is an attempt to fix the `tee` test breaking: https://github.com/dbcli/pgcli/issues/795.

Also, this fixes the error in `notee` test which should not be `notee` (that's a mycli command) but `\o`: https://github.com/dbcli/pgcli/issues/797.

## Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,8 @@ python:
   - "3.5"
   - "3.6"
 
-install: 
-  - pip install . docutils pytest mock codecov==1.5.1 behave pexpect==3.3
+install:
+  - pip install . docutils pytest mock codecov==1.5.1 behave pexpect==3.3 retrying>=1.3.3
   - pip install git+https://github.com/hayd/pep8radius.git
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,8 @@ python:
   - "3.6"
 
 install:
-  - pip install . docutils pytest mock codecov==1.5.1 behave pexpect==3.3 retrying>=1.3.3
-  - pip install git+https://github.com/hayd/pep8radius.git
+  - pip install .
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -6,8 +6,8 @@ python:
   - "3.5"
   - "3.6"
 
-install: 
-  - pip install . docutils pytest mock codecov==1.5.1 behave pexpect==3.3
+install:
+  - pip install . docutils pytest mock codecov==1.5.1 behave pexpect==3.3 retrying>=1.3.3
   - pip install git+https://github.com/hayd/pep8radius.git
```
tsroten: I think you could probably replace both of these pip install lines with these two so we don't have to keep them up-to-date. Just add `docutils` and `codecov` to the requirements file.
```
pip install .
pip install -r requirements-dev.txt
```

j-bennet: Good point.

tests/features/environment.py
```diff
@@ -134,6 +134,7 @@ def after_scenario(context, _):
                 '{0}> '.format(dbname),
                 timeout=5
             )
+        context.cli.sendcontrol('u')
```
meeuw: for mycli I had to use `sendcontrol('c')` to have always a "clean" prompt... ctrl-u removed the current line and there might be a multi-line statement in the prompt...

j-bennet: Well, after scenario is done, there shouldn't be anything in prompt, that's the point.

meeuw: while using ctrl-u there might be something left at the prompt though...

j-bennet: Didn't know that. In that case I better switch to Ctrl + C.

Issue #799, opened 2017-10-11
Autocomplete adds extra quotes to database name


lazycoder-ru:
pgcli==1.8.1
Tryin to connect to database:
`\c "Kkar.Billing"`
and get this error:
`FATAL:  database ""Kkar.Billing"" does not exist`

And if I connect this way:
`\c Kkar.Billing`
there is no error

The problem is that database autocomplete adds quotes by default if db name contains mixed case.
![image](https://user-images.githubusercontent.com/6605234/31445033-f8048536-aea5-11e7-8c5c-b492ca89adac.png)

amjith:
This is an oversight on our part. pgcli tries to be helpful by quoting mixed case table names and column names but we didn't think about database names. :disappointed: 

I'll see what I can do about this. Thanks for reporting the bug. :)

lazycoder-ru:
It has appeared in 1.8.1, in 1.8.0 everything was fine.

cfournie:
Fixed by #800

PR #800, opened 2017-10-11
Do NOT escape database names.

amjith:
## Description
Don't quote the database names regardless of the mixed case.

## Checklist

### DIFF
--- a/pgcli/pgcompleter.py
+++ b/pgcli/pgcompleter.py
@@ -138,7 +138,6 @@ class PGCompleter(Completer):
         return [self.escape_name(name) for name in names]
 
     def extend_database_names(self, databases):
-        databases = self.escaped_names(databases)
         self.databases.extend(databases)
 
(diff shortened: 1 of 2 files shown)


amjith:
Addresses #799

pgcli/pgcompleter.py
```diff
@@ -138,7 +138,6 @@ def escaped_names(self, names):
         return [self.escape_name(name) for name in names]
 
     def extend_database_names(self, databases):
-        databases = self.escaped_names(databases)
```
j-bennet: Is the problem here that we escape database names, or that `escape_name` does not check for the name being already quoted?

amjith: I think we shouldn't be escaping database names because having MiXedCasE database names are valid in Postgres. So `\c MiXedCasE` is a perfectly legal command and it works.

Issue #805, opened 2017-10-13
Cannot start program after installing with pip


torkelrogstad:
## Description

`pgcli` crashes immediately after starting when I install it with `pip`.
``` 
$ pgcli
zsh: command not found pgcli

$ pip install pgcli
...
Successfully installed pgcli-1.8.1

$ pgcli
'NoneType' object is not subscriptable
```

After installing with `apt install pgcli` I can run the program, but then I run into the problem described in #466.

## Your environment

OS: Linux Mint 18.2 
Python version: 3.5.2
CLI version: 1.8.1
```
$ pip freeze
alabaster==0.7.9
anaconda-clean==1.0
anaconda-client==1.5.1
anaconda-navigator==1.3.1
argcomplete==1.9.2
astroid==1.4.7
astropy==1.2.1
Babel==2.3.4
backports.csv==1.0.5
backports.shutil-get-terminal-size==1.0.0
beautifulsoup4==4.5.1
bitarray==0.8.1
blaze==0.10.1
bokeh==0.12.2
boto==2.42.0
Bottleneck==1.1.0
cffi==1.7.0
chest==0.2.3
cli-helpers==0.2.3
click==6.7
cloudpickle==0.2.1
clyent==1.2.2
colorama==0.3.7
conda==4.3.29
conda-build==2.0.2
configobj==5.0.6
contextlib2==0.5.3
cryptography==1.5
cycler==0.10.0
Cython==0.24.1
cytoolz==0.8.0
dask==0.11.0
datashape==0.5.2
decorator==4.0.10
dill==0.2.5
docutils==0.12
dynd==0.7.3.dev1
et-xmlfile==1.0.1
fastcache==1.0.2
filelock==2.0.6
Flask==0.11.1
Flask-Cors==2.1.2
gevent==1.1.2
greenlet==0.4.10
h5py==2.6.0
HeapDict==1.0.0
humanize==0.5.1
idna==2.1
imagesize==0.7.1
ipykernel==4.5.0
ipython==5.1.0
ipython-genutils==0.1.0
ipywidgets==5.2.2
itsdangerous==0.24
jdcal==1.2
jedi==0.9.0
Jinja2==2.8
jsonschema==2.5.1
jupyter==1.0.0
jupyter-client==4.4.0
jupyter-console==5.0.0
jupyter-core==4.2.0
lazy-object-proxy==1.2.1
line-profiler==2.0
llvmlite==0.13.0
locket==0.2.0
lxml==3.6.4
[...]

PR #808, opened 2017-10-21
The unix_socket_directories setting may not be present.

j-bennet:
## Description

An attempt to fix https://github.com/dbcli/pgcli/issues/805. The `unix_socket_directories` setting may not be present in `pg_settings` view. I'm not sure if this will fully fix the problem, as I was not able to completely reproduce the conditions. But the safeguard is needed when retrieving the setting anyway.

## Checklist


meeuw:
Right, `fetchone` can return `None` so that should be handled. Nice fix.

PR #809, opened 2017-10-22
[Proposal] Fpietka/pep8

fpietka:
## Description
Attempt to solve all PEP8 related issues

## Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,7 @@ script:
   - cd ..
   # check for changelog ReST compliance
   - rst2html.py --halt=warning changelog.rst >/dev/null
-  # check for pep8 errors, only looking at branch vs master. If there are errors, show diff and return an error code.
-  - pep8radius master --docformatter --diff --error-status
+  - pep8 .
   - set +e
 
 after_success:
--- a/changelog.rst
+++ b/changelog.rst
@@ -5,6 +5,7 @@ Internal changes:
 -----------------
 * Use temporary dir as config location in tests
 * Fix errors in the ``tee`` test (#795 and #797). (Thanks: `Irina Truong`_)
+* Fully PEP8 compliant (Thanks: `François Pietka`_)
 
 Bug Fixes:
 ----------
--- a/pgcli/completion_refresher.py
+++ b/pgcli/completion_refresher.py
@@ -116,6 +116,7 @@ def refresh_views(completer, executor):
     completer.extend_relations(executor.views(), kind='views')
     completer.extend_columns(executor.view_columns(), kind='views')
 
+
 @refresher('types')
 def refresh_types(completer, executor):
     completer.extend_datatypes(executor.datatypes())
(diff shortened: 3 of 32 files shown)


fpietka:
@dbcli/pgcli-core I took care of splitting each type of change so we can discuss together if this is something we want, or partially want.

This is of course open for discussion.

j-bennet:
Whoa @fpietka, that's one large PR! Out of curiosity, what prompted it?

j-bennet:
@fpietka I don't mind the rest of the fixes, but I would rather not enforce [binary operator position](https://github.com/dbcli/pgcli/pull/809/commits/64771cb8e4a875bfbaf083df2c08ed543c08ef9a). @dbcli/pgcli-core - any strong opinions here? We can move the discussion to slack if you prefer.

meeuw:
@j-bennet my opinion is that we should avoid "multi-line-one-liners" in the first place.

j-bennet:
@meeuw What do you replace those with? :)

meeuw:
It really depends, `if`'s with long conditions could be rewritten by introducting some variables:
```
if (prev_tok and prev_tok.value and
        prev_tok.value.lower().split(' ')[-1] == 'using'):
```
could be rewritten to (A):
```
lastword = prev_tok.value.lower().split(' ')[-1]
if (prev_tok and prev_tok.value and lastword == 'using'):
```
or, using nested `if`'s (B)
```
if (prev_tok and prev_tok.value):
    if prev_tok.value.lower().split(' ')[-1] == 'using'):
```
I think this improves the readability and could help when running the code in a debugger.

I'd prefer (A) btw.

Oh I think this is mostly nitpicking, I'd propose to rewrite this only when doing serious refactors and even then; I think I'd be guilty to add conditions "as a oneliner" as well...

bersace:
@meeuw 👍 

*Readability counts.*

bersace:
> Oh I think this is mostly nitpicking, I'd propose to rewrite this only when doing serious refactors and even then; I think I'd be guilty to add conditions "as a oneliner" as well...

👍 too. Don't pollute commit history with lintwar :-)

fpietka:
@j-bennet what triggered it is one part my editor prompting me the issues, one (big) part my OCD ^^ also I found pep8radius failing me from time to time.
Concerning the binary operator position, maybe we can than change/ignore the rule.

@dbcli/pgcli-core concerning multiline splits I didn't make any decision, nor have strong preference. I think it would be nice to agree on a pattern for the code to come (and we can discard what I've done here) since linter do not enforce anything.

I wasn't intending a lint-war, so I limited my changes to the lint rule of the project and I have no problem discarding part/all of the changes. For myself I'm more interested in the discussion we are now having to know where everybody stand 😁

j-bennet:
To me, these fixes can be split into 2 main groups:

1. Changes that improve code readability.
* Blank lines.
* Whitespaces.
* Adding two spaces before inline comment.
* Removing accidental duplicated spaces.
* Removing unnecessary backslash.
* Fixing long lines.
* Removing unused imports/vars.

2. Changes that are a matter of taste.
* Binary operator position.
* Replacing lambda with def

I'd fix Group 1 and skip Group 2.

PR #810, opened 2017-10-22
Increase timeout for quitting pgcli

meeuw:
## Description
Increase timeout for quitting pgcli

## Checklist


fpietka:
looks safe to me 👍 
time will tell if that issue is definitely gone

PR #815, opened 2017-11-17
Use other prompt (prompt_dsn) when connecting using --dsn parameter

msztolcman:
## Description
I use `--dsn`/`-D` and dsn_aliases very often, and I am connecting often to databases called identically but on different hosts (for example on testing env, on stage env etc). There is no host specified in this case (dsn_alias isn't parsed by pgcli itself, but passed to PGExecute, so it's unusable). But I found dsn alias very useful in prompt. I decided to use different prompts for both cases (connecting with --dsn option and without).

## Checklist

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -68,6 +68,7 @@ Contributors:
     * Donnell Muse
     * Andrew Speed
     * Dmitry B
+    * Marcin Sztolcman
 
 Creator:
 --------
--- a/changelog.rst
+++ b/changelog.rst
@@ -1,5 +1,6 @@
 Upcoming
 ========
+* Use other prompt (prompt_dsn) when connecting using --dsn parameter
 
 Internal changes:
 -----------------
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -93,6 +93,7 @@ OutputSettings.__new__.__defaults__ = (
 class PGCli(object):
 
     default_prompt = '\\u@\\h:\\d> '
+    default_prompt_dsn = '\\dsn> '
     max_len_prompt = 30
 
     def set_default_pager(self, config):
@@ -118,12 +119,13 @@ class PGCli(object):
 
     def __init__(self, force_passwd_prompt=False, never_passwd_prompt=False,
                  pgexecute=None, pgclirc_file=None, row_limit=None,
-                 single_connection=False, less_chatty=None, prompt=None,
+                 single_connection=False, less_chatty=None, prompt=None, prompt_dsn=None,
                  auto_vertical_output=False):
 
         self.force_passwd_prompt = force_passwd_prompt
         self.never_passwd_prompt = never_passwd_prompt
         self.pgexecute = pgexecute
+        self.dsn_name = None
 
         # Load config.
         c = self.config = get_config(pgclirc_file)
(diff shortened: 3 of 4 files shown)


j-bennet:
@msztolcman When connecting with `-D` or dsn, pgexecute does not know the connection details, so it reads them:

https://github.com/dbcli/pgcli/blob/5f88e8a042900337e55efb5083e84410c37cbd3c/pgcli/pgexecute.py#L208-L216

So after that read, you're supposed to have host, user etc. to display in prompt. Is it not working?

j-bennet:
@msztolcman Thank you for the PR! It is a useful feature. Could we perhaps keep the part with parsing `\\dsn` in command prompt and replacing it with dsn name, but not add a new config option? I'm reluctant to add new config options, as that makes `pgclirc` grow out of control and become intimidating. Instead, you could use an alias script connect to these databases with `pgcli --prompt "\dsn >" -D dsn_name`.

What do you think?

msztolcman:
Hi @j-bennet,
Sorry, I missed your previous answer :(
So, answering first question:

> So after that read, you're supposed to have host, user etc. to display in prompt. Is it not working?

No, it's not. But theoretically I can parse DSN by myself and set this data, but I don't know there is no caveats for it. Or use library like https://github.com/gruns/furl (or similar), to not reinvent the wheel :) What do you think?

About second one:
There is few solutions.
First (and IMHO best one) I described above (parsing dsn). It gives us all tokens we need and support in "normal" prompt.
Second is proposed by you, but it's a little uncomfortable, because maintaining this require always to add new dsn to pgcli's config file and create new shell alias/script for this.
Third is to leave my original version.
Fourth is to allow to use environment variables for this kind of stuff you don't want to put in config file :)
And last one is to expand prompt parser to handle some conditions (if dsn use dsn, instead use other one). It has some advantages (do not show '(none)' as port if none is specified for example), but also complexity of this is a little to big like for me.

j-bennet:
@msztolcman So what do you see in prompt when connecting with dsn alias?

> Second is proposed by you, but it's a little uncomfortable, because maintaining this require always to add new dsn to pgcli's config file and create new shell alias/script for this.

For the solution I had in mind, you'd only need one generic alias, for example:

`alias pgclid="pgcli --prompt '\\dsn >'"`

Then you'd call `pgclid -D some_alias` instead of `pgcli -D some_alias`, and `\\dsn` placeholder would be replaced with the alias.

You would have to add dsn aliases to pgcli config file, yes, but you already do that.

> Third is to leave my original version.

This is still an option.

> Fourth is to allow to use environment variables for this kind of stuff you don't want to put in config file :)

Yeah, those two are equally bad. :)

> And last one is to expand prompt parser to handle some conditions (if dsn use dsn, instead use other one). It has some advantages (do not show '(none)' as port if none is specified for example), but also complexity of this is a little to big like for me.

I was thinking about always using dsn alias in prompt if one is specified, but that seems like a feature a lot of people may dislike.

msztolcman:
@j-bennet OK, it's a little bit embarrassing... it looks like it works, but with small issue :( I missed it, and you pointed me in a right direction.

It works fine. I missed there is database name in my prompt after logging into my database using dsn aliases. But my vhost name is quite long, and whole prompt is longer then PGCli.max_len_prompt, and my promt is shortened to only '\d'.

So, there is other issue: there is a condition that is verifying prompt is default or not. My is default, so it just change my prompt :( And even passing prompt as argument to pgcli it doesn't use it, just use '\d'. But I will report it in other issue.

j-bennet:
@msztolcman I think `dsn_alias` in prompt can still be a useful change!

msztolcman:
@j-bennet OK, I've changed this a little and it's in other PR: https://github.com/dbcli/pgcli/pull/820

Issue #823, opened 2017-12-20
Optionally including username to password promp


delicb:
This is suggestion for improvement, not a bug.

I am using `pgcli` quite often. My terminal app is iTerm2 which has builtin password manager. It recognizes regex in terminal and, when triggered, offers to enter password read from KeyChain. Quite handy feature.

Problem is that, with `pgcli`, there is no single regex to use for different databases, since prompt is "Password:". I would like to make distinction based on user I am using to connect.

Currently I patch local installation of `pgcli` to include username (which solves my problem, but more generic solution would probably include both username and database, maybe even host) in password prompt. 

These are two places that I have modified:
https://github.com/dbcli/pgcli/blob/master/pgcli/main.py#L368
https://github.com/dbcli/pgcli/blob/master/pgcli/main.py#L386

My modification looks something like this:
```python
passwd = click.prompt('%s password' % user, hide_input=True,
                       show_default=False, type=str)
```

I understand if you do not want to change default behavior, but would you consider adding flag that would modify password prompt to include username (and maybe database and host)? If yes, I might be able to provide PR.

Thanks.

j-bennet:
Certainly! This is a valid use case. If you wanted to change the default behavior to ask for `Password (<username>):` (or something along those lines) instead of `Password:`, I don't think it would be a problem either. It's a one-time prompt that quickly goes away, and even if someone does not like extra verbosity, it's not staying on screen long enough to be annoying.

amjith:
@delicb If you have a local patch that you're currently using why not make it a PR. If we have modifications or suggestions we can discuss it in the PR. It is easier (faster) to get a feature in if there is a PR. 😄

delicb:
@j-bennet Great, thank you. I will go with changing default behavior then, since it seem as most straight forward way, no unnecessary complications. 

@amjith Certainly, I will prepare PR promptly. 

Thank you both for being so cooperative :).

j-bennet:
Included in 1.8.2 (just released).

PR #824, opened 2017-12-20
Include username to password prompt. #823

delicb:
## Description
As described in ticket #823, this is useful for matching prompt for
different usernames and auto populating passwords. iTerm2 has password
manager that is capable of doing this, but other terminal managers might
be able to do it as well.

`behave`, `pytest` and `pep8radius` are passing locally. I have not tried building `deb` or `rpm` packages, but these changes should not affect that in any way.

I am opened for any feedback you might have. 

## Checklist

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -365,7 +365,7 @@ class PGCli(object):
         # If we successfully parsed a password from a URI, there's no need to
         # prompt for it, even with the -W flag
         if self.force_passwd_prompt and not passwd:
-            passwd = click.prompt('Password', hide_input=True,
+            passwd = click.prompt('Password for %s' % user, hide_input=True,
                                   show_default=False, type=str)
 
         # Prompt for a password after 1st attempt to connect without a password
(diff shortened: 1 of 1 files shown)


pgcli/main.py
```diff
@@ -365,7 +365,7 @@ def connect(self, database='', host='', user='', port='', passwd='',
         # If we successfully parsed a password from a URI, there's no need to
         # prompt for it, even with the -W flag
         if self.force_passwd_prompt and not passwd:
-            passwd = click.prompt('Password', hide_input=True,
+            passwd = click.prompt('Password [%s]' % user, hide_input=True,
```
amjith: One tiny nitpick can you add the word `for` before the user name in the square brackets? 
eg: 
```
    passwd = click.prompt('Password [for %s]' % user, hide_input=True,
```
This is to avoid the user from thinking that we're showing some default value.

pgcli/main.py
```diff
@@ -383,8 +383,9 @@ def connect(self, database='', host='', user='', port='', passwd='',
             except (OperationalError, InterfaceError) as e:
                 if ('no password supplied' in utf8tounicode(e.args[0]) and
                         auto_passwd_prompt):
-                    passwd = click.prompt('Password', hide_input=True,
-                                          show_default=False, type=str)
+                    passwd = click.prompt('Password [%s]' % user,
```
amjith: Same as above.

pgcli/main.py
j-bennet: If you're adding a `for`, I think it's better to remove the `[]`. `Password for XXX` will look better, more like a coherent phrase.

delicb: Sure thing, in a minute.

delicb: Done.

amjith:
👾

PR #826, opened 2017-12-21
Added prompt_toolkit 2.0 support.

jonathanslenders:
Attempt to convert to prompt_toolkit 2.0.

There is still an issue with the completer. After accepting the first input, the completer stops working. I'm not sure what's going on.

Prompt_toolkit 2.0 is not yet released, but this is an experiment to see what it takes to convert an application.

Questions I still have:
1. Shall we convert all prompt_toolkit application to 2.0? I'd very much like to see that.
2. How to deal with the incompatibilities? Linux distribution maintainers don't like this. I could rename the package to prompt_toolkit2 during a transition period. I don't like it that much, but I'm willing to do it if this is what works best for the community. The problem is that right now you can't have both prompt_toolkit 1 and 2 installed in the system wide Python installation, and this is often where these command line tools go...


jonathanslenders:
See: https://github.com/jonathanslenders/python-prompt-toolkit/issues/568

j-bennet:
@jonathanslenders Oh wow. Yes, it would be good to convert to 2.0 as soon as we can (provided we fix the "completer not working" problem. :)).

The compatibility problem is tough. We always recommend installing pgcli/mycli in a venv, but for some, it's not an option. I''d like to know if @amjith, @tsroten and @meeuw have thoughts on that.

amjith:
Hey Jonathan, It is great to see v2.0 progressing. :)

I want to encourage you to experiment with pgcli and keep it in a branch for now. We're happy to give you write permission to the repo so you can make a branch and do the development directly in pgcli repo instead of a fork. 

I'm wary of releasing a new pgcli version with prompt-toolkit-2.0. I've come to think of pgcli is a piece of software that people rely on to do their day to day job. It has grown out of being a toy side project. So I'd like to be cautious when we release new versions. Especially new versions with breaking changes. 

But we have a healthy set of beta testers (core devs and a few fervent followers). We can ask them to try out the branch build and gather feedback. 

Thoughts?

jonathanslenders:
Hi @amjith,

I completely agree. The same applied to IPython.
Maybe give this pull request a couple of months if needed. There's no hurry.

I added a second commit. This fixes the issue that I had with the completer. The problem was that the ``completer`` attribute was set while the prompt was running, but the ``prompt()`` function restores the settings when it's done. Using the ``DynamicCompleter`` made the fix pretty easy.

Please let me know if somebody comes across any issues.

jonathanslenders:
Hi @j-bennet, I updated this merge request so that it's compatible with the ptk 2.0 version that is released and I rebased it on your latest master commit.

The completer issue that I noticed earlier, is fixed. (I'm not sure what happened back then.)

One issue, which still needs to be fixed is that if you press semi-colon in a multiline input mode, that the line continuation disappears. This is because the ``pg_is_multiline`` function in ``pgbuffer.py`` return ``False`` at that point. Further, I think this requires some more testing.

pgcli/main.py
```diff
@@ -845,12 +822,6 @@ def _swap_completer_objects(self, new_completer, persist_priorities):
                 # Leave the new prioritizer as is
                 pass
 
-            # When pgcli is first launched we call refresh_completions before
```
j-bennet: @jonathanslenders Was this lost accidentally or intentionally?

j-bennet:
Closed in favor of https://github.com/dbcli/pgcli/pull/930 (cherry-picking was easier than rebasing).

PR #830, opened 2018-01-03
Change format output

fraoustin:
## Description

add command "\t" for change format output

usage: "\t rst"

## Checklist


fraoustin:
Hi i need some help: i don't understand why test is ko

fraoustin:
just python3.3 is ko but i don't understand ...

fraoustin:
Youpi !!!

j-bennet:
@fraoustin Looks like you figured out the test problem. One question - why `\t` for changing format?

fraoustin:
Hi, on mycli the command is \T (http://www.mycli.net/commands) ... but on pgcli all command is lowercase so "\t". I use that command when i need save the result in file (I change format for plain or csv)

j-bennet:
It looks like `psql` has both `\t` and `\T`:

```
\t
Toggles the display of output column name headings and row count footer. This command is equivalent to \pset tuples_only and is provided for convenience.

\T table_options
Specifies attributes to be placed within the table tag in HTML output format. This command is equivalent to \pset tableattr table_options.
```

I think it's better to use `\T` here, because it aligns with `mycli`, but also because psql uses `\T` to configure table format.

fraoustin:
I changed \t by \T but i have a problem of Timeout. I don't understand the problem ... can you help me please

fraoustin:
it is ok ... but I have generate a commit "null" , no change ... mistake

So \T is a new command for change format table

pgcli/main.py
```diff
@@ -212,6 +212,24 @@ def register_special_commands(self):
...
+                raise ValueError()
+            self.table_format = pattern
+            yield (None, None, None,
+                   'Changed table format to {}'.format(pattern))
+        except ValueError:
+            msg = 'Table format {} not recognized. Allowed formats:'.format(
+                pattern)
+            for table_type in TabularOutputFormatter().supported_formats:
+                msg += "\n\t{}".format(table_type)
+            msg += '\ncurrently table format: %s' % self.table_format
```
j-bennet: `currently table format` sounds a bit awkward. How about `Current value: XXX` or `Currently set to: XXX`?

j-bennet:
Thank you for the PR!

🍬

PR #834, opened 2018-01-11
Add feature Color of table. 

fraoustin:
## Description

Add feature Color of table by parameter

I add the possibility of use a name of token of syntax style (Token.Name.Tag, Token.Literal.String, ..) for parameter. The color of table is function of syntax style

## Checklist


tsroten:
@fraoustin Nice work! It might be nice to move the style-related code to the `pgstyle.py` file, that way, we have very little color-related logic in the main pgcli itself. What do you think?

Here is the example from Mycli: 
https://github.com/dbcli/mycli/pull/462/files

pgcli/pgclirc
```diff
@@ -154,6 +154,12 @@ Token.Toolbar.Arg.Text = 'nobold'
 Token.Toolbar.Transaction.Valid = 'bg:#222222 #00ff5f bold'
 Token.Toolbar.Transaction.Failed = 'bg:#222222 #ff005f bold'
 
+# color of table
+# you can use token or custom colors 
+Token.Output.Header = "Token.Name.Tag"
+Token.Output.OddRow = ""
+Token.Output.EvenRow = "Token.Literal.String"
```
tsroten: I could see the use of other token names here a little confusing to the user. What about using HTML color codes or ANSI color names so it's very clear what colors these are?

tsroten:
I'm posting this info here so we have it in the comments and can either refer to it later or put it in the sample config file:

```
# How the styles are rendered is highly dependent on which terminal emulator you are
# using and its settings. Here are the recognized style rules:
# * bold - Render text as bold (some terminals render bold as a bright color instead).
# * italic - Render text as italic.
# * underline - Render text with an underline.
# * bg: - Transparent background.
# * bg:#000000 - Background color (in this case, black).
# * #ff0000 - text color (in this case, red).
#
# You can also use ANSI colors, specified as #ansi*, where * is one of the colors below.
# These are the "light" colors that will be rendered bold by most terminals:
# darkgray, red, green, yellow, blue, fuchsia, turquoise, white
# These are the "dark" colors that will be rendered non-bold:
# black, darkred, darkgreen, brown, darkblue, purple, teal, lightgray
```

j-bennet:
@fraoustin I believe this is ready to merge - please add the contribution to changelog!

j-bennet:
Looks good. Merging!

🍫

PR #843, opened 2018-02-15
Fix tests

pgiraud:
## Description

Currently the build tests are broken. This PR fixes them.
There was a wrong assertion on the context when using `ex` external editor.

Also there seem to a bug in the last version of autopep8. Pinning the version to 1.3.3 seem to work.

## Checklist

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -73,6 +73,7 @@ Contributors:
     * Bojan Delić
     * Chris Vaughn
     * Frederic Aoustin
+    * Pierre Giraud
 
 
 Creator:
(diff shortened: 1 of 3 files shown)


j-bennet:
Thanks for the PR! You can also now switch to installing `pep8radius` from pypi, like here:

https://github.com/dbcli/mycli/commit/1da113b5491dbf76a5dcd7c452be27e61e91462a

pgiraud:
@j-bennet I tried this already but we need a version of pep8radius more recent than the one available on pypi since we rely on the `--error-status` option.

j-bennet:
I see. Well done!

PR #856, opened 2018-03-21
Ported release script improvements from mycli.

j-bennet:
I honestly stole all @tsroten's release script tweaks.


meeuw:
I've checked for differences in {mycli,pgcli}/release.py and found only replacements of pgcli/mycli :+1:

Issue #858, opened 2018-03-23
Customized prompt not used


akuchling:
## Description

I was puzzled when connecting to one server produced the `user@server:dbname>` prompt I expected, but another server produced just `dbname>`.  

It turns out that if the prompt ends up longer than `self.max_len_prompt`, which is set to 30 and isn't configurable, pgcli falls back to just `dbname>`.    I was using just 'docker' for one hostname and the other was 'db-dev.domainname.tld', which pushed the whole prompt over the limit.  (My username is always 'akuchling', so there's 9 characters right there!)

Two ideas:

1) bash's prompt formatting supports both the FQDN and just the local hostname.  Would you like me to write a patch adding a similar new formatting character?  Unfortunately bash uses \h for the up-to-the-first-'.' and \H for the full name.  This would be an incompatible change for pgcli, where \h is currently the full server name.  Would we want to change the meaning of \h in pgcli, have the opposite of bash's variables (\h for full, \H for shortened), or just pick some different character?

2) We could also make the max_len_prompt configurable.  Does that seem worth doing?

## Your environment

pgcli version 1.9.0.

j-bennet:
A "short host name" formatting character would be great. I think it would be ok if we switch the meaning of `\h` to "short host name", and use `\H` for "full host name" (users might even appreciate shorter prompt). Perhaps with this change you would not need to increase `max_len_prompt`.

akuchling:
Closing this issue, because a fix was merged.

PR #861, opened 2018-03-28
#858: make \h in prompt return the first part of the hostname; add \H to return whole hostname

akuchling:
## Description
This patch makes \h in the prompt only return the hostname up to the first period (i.e. db.orgname.org will be displayed as just 'db'), and adds \H which will return the whole hostname.  If there's no '.' in the hostname, \h and \H will produce identical results.

The `get_prompt()` method doesn't seem to have a test suite that exercises all of the format characters; should it?  I did verify that `get_prompt()` is at least invoked by the behave tests.

## Checklist

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -74,6 +74,7 @@ Contributors:
     * Chris Vaughn
     * Frederic Aoustin
     * Pierre Giraud
+    * Andrew Kuchling
 
 
 Creator:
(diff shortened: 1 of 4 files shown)


akuchling:
Thanks for pointing that out.  I've added another commit to update pgclirc.

j-bennet:
Works great! Merging.

PR #866, opened 2018-04-05
Mark tests requiring a running database server as dbtest

meeuw:
## Description
Mark tests requiring a running database server as dbtest

## Checklist


j-bennet:
Funny enough, both of those tests are for output formatting, so we could have mocked execution and not gone to database. But 👍 .

Issue #849, opened 2018-03-07
Not sure about psycopg2-wheels dependency


dvarrazzo:
Hello, I'm psycopg2 maintainer.

I have developed wheel package of psycopg as a way to make tools like yours easier to install, but unfortunately the psycopg2 wheels has complex problems hard or impossible to eliminate, which forced us, in order to not get rid altogether of the wheel, to deploy them under a different pip name, as you have found seeing the requirements in your `setup.py`.

In psycopg/psycopg2#674 someone has reported though that the dependencies of pgcli cause problems to shared virtual envs. My suggestion would be to not have pgcli installed in the share virtualenv of other projects as it's a standalone executable, not a library. However I'm not sure about that. The existence of a -wheel package is a new strategy for us so I'd like to find the best way to tackle the problem.

Do you think there are scenarios where `pgcli` is meant to be installed as a Python library for other projects? If so how can we solve conflicts where other projects depend on the naked `psycopg2`? Glad to know your opinion.

j-bennet:
`pgcli`'s primary purpose is to be an interactive tool, not a library. We worked on extracting shared functionality into libraries such as `pgspecial` and `cli_helpers`. However, many users install `pgcli` and `mycli` into venvs, to avoid changes to their system-wide packages. I wish we could specify `psycopg2` as either binary or source distribution, to be controlled with `pip install` flags, but I'm not sure if there's an existing way to do that.

@dbcli/pgcli-core @dbcli/mycli-core What do you think?

(Related issue: https://github.com/dbcli/pgcli/issues/844.)

dvarrazzo:
If your userbase is already used to installing psycopg from source is maybe better to specify the dependency on the source package. Here is [how to do it in a requirements.txt file](http://initd.org/psycopg/docs/install.html#disabling-wheel-packages-for-psycopg-2-7) but I'm not sure it works in `setup.py` too.

bersace:
@dvarrazzo how do you manage a `OR` dependency with setuptools ? Something like `psycopg2 | psycopg2-binary` ? How to let *user* choose whether they use binary or build from source ?

bastianh:
This is a bit annoying .. I install pgcli into a django docker container that runs in aws to use it as the "manage.py dbshell" program so I don't need to install the postgres client itself.

Now, both.. psycopg2 and psycopg-binary get installed .. but it looks like it still prefers the non-binary version so I now get the warning also for the django and celery processes that also run in the container.

[...]

PR #867, opened 2018-04-24
Rename psycopg2 to psycopg2-binary

gabfl:
## Description

The package `psycopg2` is being renamed to `psycopg2-binary` and currently throws the following warning:

```
>>> import psycopg2
/usr/local/lib/python3.5/dist-packages/psycopg2/__init__.py:144: UserWarning: The psycopg2 wheel package will be renamed from release 2.8; in order to keep installing from binary please use "pip install psycopg2-binary" instead. For details see: <http://initd.org/psycopg/docs/install.html#binary-install-from-pypi>.
```

See more information at http://initd.org/psycopg/docs/install.html#binary-install-from-pypi

## Checklist


j-bennet:
We already did this, for a brief period of time, in https://github.com/dbcli/pgcli/pull/846, and then reverted back in https://github.com/dbcli/pgcli/pull/859, because of https://github.com/dbcli/pgcli/issues/849.

There's a subset of users that can't run with psycopg2-binary installation. So we're currently not forcing it. This means that until 2.8 (first source-only version) is released, psycopg2 will spit out a warning. It's annoying, but at least it's not breaking things. Once 2.8 is released, the warning should go away. As far as I understand, users that wish to use psycopg2-binary will still be able to install that over psycopg2. If my understanding is not correct, we'll have to figure out a way for users to select binary vs source installation after 2.8 is out.

Thank you for the PR!

gabfl:
gotcha, thanks for the quick reply

PR #869, opened 2018-04-26
Add application name

### LINKED ISSUE #868: Add application_name parameter for Postgres 9.x & 10.x connections
## Feature request
Hi, would it be possible to add `application_name="pgcli"` connection parameter to all connections, for the sake of Postgres 9.x or 10.x? See https://www.postgresql.org/docs/9.0/static/runtime-config-logging.html#GUC-APPLICATION-NAME

With this, database administrators will be able to see "pgcli" connections explicitly in the list of connections for debugging reasons (ie. `SELECT * FROM pg_stat_activity`).

FYI, `psql` and many more UI clients add the application_name by default too.

Thanks for any comments! And thanks for this project, I use it every day 👍

fpietka:
## Description
Adding `application_name` to help identify `pgcli` connections to postgres.

Fix #868

## Checklist

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -5,6 +5,7 @@ Internal changes:
 -----------------
 
 * Mark tests requiring a running database server as dbtest (Thanks: `Dick Marinus`_)
+* Add ``application_name`` to help identify pgcli connection to database (issue #868) (Thanks: `François Pietka`_)
 
 1.9.1:
 ======
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -260,7 +260,7 @@ class PGCli(object):
             db, user, host, port = infos
             try:
                 self.pgexecute.connect(database=db, user=user, host=host,
-                                       port=port)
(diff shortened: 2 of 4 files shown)


fpietka:
There might be a better way, since I'm adding this a few times. I'm not sure we need it on simple `connect()` calls, so I'm gonna run some more tests, and update the PR with my results.

fpietka:
In the end I only kept this parameter in case we are connecting to a specified host, as other cases are working fine without it (hence are displaying the `application_name` correctly)

meeuw:
I'm trying to write a test to run `SELECT * FROM pg_stat_activity WHERE application_name = 'pgcli'` but this needs to be run by `pgcli/main.py` and not `tests/utils.py`.

j-bennet:
@meeuw A behave test can do that.

meeuw:
nice work @fpietka!

VojtechVitek:
Well done, thank you guys! 🎉

PR #871, opened 2018-05-12
Respect \pset pager on expected behavior

### LINKED ISSUE #813: Option to intelligently enable pager based on output size (pset is not working)
## Description
It gets annoying that the pager is either on (set to `less`) or off (set to `cat`). When the query has a large number of output lines, it's helpful to have the pager on, but when doing a bunch of small queries in preparation for a larger query, it's helpful to have the pager off so you can see the query results in scrollback.

Would it be possible to have a pager option similar to `\x auto` that decides whether to use the pager based on the size of the output?

## Your environment

- OS: Mac OS 10.12.6
- pgcli version: 1.8.1

maxrothman:
`\pset pager` has three possible values: "always", "on", and "off".
pgcli previously treated all non-"off" values as "always". This change
implements the expected behavior, which is to use the pager when the
output is larger than the terminal height (See \pset pager in
https://www.postgresql.org/docs/9.2/static/app-psql.html). Pgcli adds
to this and also uses the pager when the output is wider than the
terminal width.

Fixes #813

## Checklist

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -42,7 +42,7 @@ from prompt_toolkit.auto_suggest import AutoSuggestFromHistory
 from pygments.lexers.sql import PostgresLexer
 from pygments.token import Token
 
-from pgspecial.main import (PGSpecial, NO_QUERY, PAGER_OFF)
+from pgspecial.main import (PGSpecial, NO_QUERY, PAGER_OFF, PAGER_LONG_OUTPUT)
 import pgspecial as special
 try:
     import keyring
@@ -76,6 +76,9 @@ import psycopg2
 from collections import namedtuple
 
 from textwrap import dedent
+
+# Ref: https://stackoverflow.com/questions/30425105/filter-special-chars-such-as-color-codes-from-shell-output
+COLOR_CODE_REGEX = re.compile(r'\x1b(\[.*?[@-~]|\].*?(\x07|\x1b\\))')
 
 # Query tuples are used for maintaining history
 MetaQuery = namedtuple(
@@ -929,6 +932,15 @@ class PGCli(object):
     def echo_via_pager(self, text, color=None):
         if self.pgspecial.pager_config == PAGER_OFF or self.watch_command:
             click.echo(text, color=color)
+        elif self.pgspecial.pager_config == PAGER_LONG_OUTPUT:
+            lines = text.split('\n')
+
+            # The last 4 lines are reserved for the pgcli menu and padding
+            if len(lines) >= self.cli.output.get_size().rows - 4 \
+                    or any(len(COLOR_CODE_REGEX.sub('', l)) > self.cli.output.get_size().columns for l in lines):
+                click.echo_via_pager(text, color=color)
+            else:
+                click.echo(text, color=color)
         else:
             click.echo_via_pager(text, color)
 
--- a/.gitignore
+++ b/.gitignore
@@ -66,3 +66,6 @@ target/
 # Generated Packages
 *.deb
 *.rpm
+
+.vscode/
+venv/
--- a/AUTHORS
+++ b/AUTHORS
@@ -84,6 +84,7 @@ Contributors:
     * Saif Hakim
     * Artur Balabanov
     * Kenny Do
+    * Max Rothman
 
 
 Creator:
--- a/changelog.rst
+++ b/changelog.rst
@@ -1,3 +1,11 @@
+Upcoming
+========
+
+Features:
+---------
+
+* Respect `\pset pager on` and use pager when output is longer than terminal height (Thanks: `Max Rothman`_)
+
 1.10.3
 ======
 
@@ -868,3 +876,4 @@ Improvements:
 .. _`Saif Hakim`: https://github.com/saifelse
 .. _`Artur Balabanov`: https://github.com/arturbalabanov
 .. _`Kenny Do`: https://github.com/kennydo
+.. _`Max Rothman`: https://github.com/maxrothman
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,4 +1,5 @@
-pytest>=2.7.0
+# The maximum version requirement can be removed once Python 3.4 goes EOL
+pytest>=2.7.0,<=3.0.7
 mock>=1.0.1
 tox>=1.9.2
 behave>=1.2.4
--- a/tests/features/steps/wrappers.py
+++ b/tests/features/steps/wrappers.py
@@ -3,6 +3,7 @@ from __future__ import unicode_literals
 
 import re
 import pexpect
+from pgcli.main import COLOR_CODE_REGEX
 
 
 def expect_exact(context, expected, timeout):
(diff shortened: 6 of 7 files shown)


maxrothman:
I realize a behave test is probably in order, but I'm not terribly familiar with behave. Could someone point me in the right direction?

j-bennet:
Ah, very interesting. Your test is a special ❄️ kind of test. In our other behave tests, we use this fake pager:

https://github.com/dbcli/pgcli/blob/master/tests/features/wrappager.py

This is because we 1) want to see full output to check it against expected, i.e. no paging and 2) want to add "boundary" characters on top and bottom of the output we are testing, to easily find it. This fake pager is set here:

https://github.com/dbcli/pgcli/blob/2f344ca6d27312bc47c2caff35a273590fb7b2ac/tests/features/environment.py#L69-L72

I pushed a change into your fork to work around the fake pager and fix the tests. It may be enough if you add a unit test like one of those here that test `PGCli` object:

/https://github.com/dbcli/pgcli/blob/2f344ca6d27312bc47c2caff35a273590fb7b2ac/tests/test_main.py#L170-L178)

because testing paging with `behave` may turn out to be a hassle.

maxrothman:
Huh, interesting. I'll take a look at adding that test.

How hard would it be to change the setup for the behave tests so they set `pset pager always` or `off`? That would avoid the need for the special case.

maxrothman:
Alright, I've got a few tests written. There's a fair amount of repetition; I'm sure the repeated setup could be factored into a fixture somehow, but I'm not super familiar with pytest, so I'm very much open to suggestions.

maxrothman:
It looks like tests are failing because `pytest.param` isn't available in the version of `pytest` that's being installed for Python 3.4. I could remove the uses of it, but it'd result in less clear test IDs. Before I go and do that, do you know if there's a different workaround?

j-bennet:
We may want to pin pytest for the time being. Python 3.4 did not reach end-of-life yet.

j-bennet:
I missed this question from you:

> How hard would it be to change the setup for the behave tests so they set pset pager always or off?

For tests, we use a temp dir as config location (see this PR: https://github.com/dbcli/pgcli/pull/796/files to figure out how it works). There's nothing in that dir, which means a default config will be used.

What you want to do is write a config file into that dir with the values you want. But there are drawbacks:

* tests run with default config because we want to make sure things work with default config.
* writing and then deleting your own config file for behave tests may be more complex than what you're doing.

maxrothman:
Fair enough! Seems like this is the best approach then.

j-bennet:
Nevermind, there's an easier way. Do this in `before_feature` handler:

```
context.cli.sendline('\pset pager always')
wrappers.wait_prompt(context)
```

maxrothman:
@j-bennet are you able to restart the Travis build? It looks like there's a flaky behave test failure and it appears that I'm not allowed to kick it off myself:

```
Closed connection: port=5432 host=localhost dbname=pgcli_behave_tests_2_7_14 user=postgres.
Created connection: port=5432 host=localhost dbname=postgres user=postgres.
HOOK-ERROR in after_all: OperationalError: database "pgcli_behave_tests_2_7_14" is being accessed by other users
DETAIL:  There is 1 other session using the database.
ABORTED: By user.
Failing scenarios:
  features/iocommands.feature:10  tee output from query
```

maxrothman:
> Nevermind, there's an easier way. Do this in `before_feature` handler:

`before_feature` or `before_all`?

j-bennet:
I think `before_all` will not give you `context.cli` object yet. Try it. If object is already there, then yes. Otherwise, `before_feature` looks like the next best thing.

j-bennet:
I'll take a look tonight.

j-bennet:
@maxrothman So, I'm pretty sure there's a way to fix the problem, but it may not be as easy as I thought. Unfortunately, I'll be away for two weeks and have no time to help. Perhaps @meeuw or @tsroten or @amjith could take a look.

I did find the right place for `pset pager always`. It's the `run_cli` method in `basic_commands.py`. That's where we create the cli. Things still break though, even with setting pager to `always`.

maxrothman:
Would it be better to leave the hack in for now and file a ticket to fix it
later?

amjith:
I see one issue with this approach. 

Right now you're checking just the height, not the width. If you have a really wide table with just a single row, we use the pager to display the output which lets the user scroll left/right using the arrow keys. 

But with the new change that you've done the default behavior is changed to not use a pager if the width is too large to fit in the window.

maxrothman:
Sorry for the radio silence, I hit a busy patch. I'll try to pick this up again in a week or so.

maxrothman:
I'm back! That was a long "week or so".

I've rebased against latest master, updated `AUTHORS` and the changelog appropriately, and addressed @amjith's concern about paging long horizontal outputs. A bunch of behave tests are failing, but I don't really understand why.

j-bennet:
I'll try to take a look at those tests in the next few days.

maxrothman:
Ping, any progress on this?

j-bennet:
I still can't figure out the problem. I'll have some time tomorrow to look again.

pgcli/main.py
```diff
@@ -926,9 +926,21 @@ def get_last_query(self):
         """Get the last query executed or None."""
         return self.query_history[-1][0] if self.query_history else None
 
+    # Ref: https://stackoverflow.com/questions/30425105/filter-special-chars-such-as-color-codes-from-shell-output
+    COLOR_PATTERN = re.compile(r'\x1b(\[.*?[@-~]|\].*?(\x07|\x1b\\))')
```
j-bennet: We're doing the exact same thing when reporting an error in behave tests:

https://github.com/dbcli/pgcli/blob/5834259535e1cf5a75204305b0533ac5559ff344/tests/features/steps/wrappers.py#L13

See if maybe you can move it into a common method? or at least have a shared constant for this regexp.

j-bennet:
It's odd that so many tests stopped working, but I guess that's what you get when releases / builds don't happen very often. 🤷‍♀️  I got tests to pass for this PR. I have a minor comment as to sharing the color code trimming regexp. Other than that, this should be fine to merge.

maxrothman:
I'll take a look at that this weekend. In the meantime, I'm planning to squashing most (all?) of my commits down before merging, do you want to stick the new ones that make tests pass on master and have me rebase on them, or not a big deal?

j-bennet:
That might be better. I'll cherry-pick those things into a separate branch and remove from yours.

j-bennet:
Moved my fixes in https://github.com/dbcli/pgcli/pull/941 and merged into master.

j-bennet:
@maxrothman Yeah, that was pep8 fixer for your new code. If you like, you can drop my commit and run pep8 fixer yourself on this branch:
`pep8radius master --docformatter -i`
You'll need to `pip install -r requirements-dev` for that.

maxrothman:
Apparently the regex I added is slightly more accurate than the one in the tests (or at least, the regex in the tests failed the test I wrote), so I used it instead of the existing one.

maxrothman:
Looks like [this behave test](https://travis-ci.org/dbcli/pgcli/jobs/432154279#L865) is failing on Python 2.7, but it seems unrelated to my change. Mind taking a look?

j-bennet:
The `tee` test fails intermittently. Fixed by restarting the build.

maxrothman:
Done! Would you like me to squash, or do you want to control that process?

j-bennet:
Yes, go ahead and squash it.

maxrothman:
Looks like the same test failed again. Mind kicking it?

j-bennet:
Merged! Nice work - thank you for sticking with it!

🍒

maxrothman:
Thanks for sticking with me! Any idea what the ETA for release will be?

j-bennet:
I'll release within the next few days.

PR #873, opened 2018-05-14
Remember prompted passwords using keyring

cfournie:
## Description
This PR adds support for remembering prompted passwords using [`keyring`](https://github.com/jaraco/keyring).

TODO:
  - [ ] Behave test
  - [ ] Add a way to disable using keyring in case a suitable backend isn't configurable
      - Via config file so it doesn't need to be specified each time it's invoked (document in README), CLI option/env var documents nicely but is cumbersome to specify on each usage

## Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,6 @@ python:
   - "3.5"
   - "3.6"
 
-before_install:
-  - sudo apt-get install python-dbus
-
 install:
   - pip install .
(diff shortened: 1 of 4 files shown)


amjith:
@cfournie Would you prefer if I merged this now or would you prefer to add the behave tests to the PR before I merge it?

amjith:
Thank you for this nice feature. 

:trumpet:

PR #880, opened 2018-05-15
Feature/get last sql query

RishiRamraj:
## Description
Refactor the command execution pipeline to flag MetaQuery with is_special. If this flag is True, the command is a special command, otherwise it's plain SQL.

## Checklist

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -78,6 +78,7 @@ MetaQuery = namedtuple(
         'db_changed',       # True if any subquery changed the database
         'path_changed',     # True if any subquery changed the search path
         'mutated',          # True if any subquery executed insert/update/delete
+        'special',          # True if the query is a special command
     ])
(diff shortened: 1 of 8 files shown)


pgcli/main.py
```diff
@@ -472,7 +473,7 @@ def handle_editor_command(self, cli, document):
             cli.application.pre_run_callables = saved_callables
         return document
 
-    def execute_command(self, text, query):
```
amjith: I believe we still need to pas in the `query` arg. 

Line 535 returns the `query` var, but it can go uninitialized if a user hits Ctrl-C during the execution of a command.

I can show you what you mean when we meet in the sprint room.

amjith:
@RishiRamraj You can fix the lint errors using pep8radius. 

`pep8radius master --docformatter --in-place`

That command will fix the style issues in the current diff. 

You will have to `pip install pep8radius`.

amjith:
I'm going to merge this and fix the pep8 violations.

PR #883, opened 2018-05-15
Add git-blame-ignore revs in anticipation of hyper-blame.

amjith:
## Description
https://commondatastorage.googleapis.com/chrome-infra-docs/flat/depot_tools/docs/html/git-hyper-blame.html

This will be useful when we use hyper-blame to ignore certain revs (such as running `black` on the entire repo).


fpietka:
PR is failing because of keyring 12.2.1 (12.2.0 works fine).
The release is young (2 days), I'll need more time to find out what's going on there.

fpietka:
Found it!
@amjith just rebase your branch, the fix is already in the master (we need `keyrings-alt` for travis)

PR #886, opened 2018-05-21
\ev and \ef commands

catherinedevlin:
## Description

Supports \ev (edit view) and \ef (edit function) commands, using the same SQL that psql uses.

To work properly, depends on a pull request to `pg_special`: https://github.com/dbcli/pgspecial/pull/63; however, nothing will break without that pull request (these commands simply won't work).

Courtesy PyCon 2018 sprints.

## Checklist

### DIFF
--- a/pgcli/pgexecute.py
+++ b/pgcli/pgexecute.py
@@ -162,11 +162,18 @@ class PGExecute(object):
         FROM pg_settings
         WHERE name = 'unix_socket_directories'
     '''
+
     view_definition_query = '''
         SELECT table_schema, table_name, view_definition 
         FROM   information_schema.views 
         WHERE  table_schema LIKE %s 
         AND    table_name LIKE %s'''
+
+    function_definition_query = '''
+        WITH f AS 
+            (SELECT %s::pg_catalog.regproc::pg_catalog.oid AS f_oid)
+        SELECT pg_catalog.pg_get_functiondef(f.f_oid)
+        FROM f'''
 
     def __init__(self, database, user, password, host, port, dsn, **kwargs):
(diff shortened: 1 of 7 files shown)


pgcli/pgexecute.py
```diff
@@ -384,6 +405,38 @@ def search_path(self):
...
+            sql = self.view_definition_query
+            _logger.debug('View Definition Query. sql: %r\nspec: %r',
+                          sql, spec)
+            try:
+                cur.execute(sql, (spec, ))
+            except psycopg2.ProgrammingError:
+                raise RuntimeError('View {} does not exist.'.format(spec))
+            result = cur.fetchone()
+            view_type = 'MATERIALIZED' if result[2] == 'm' else ''
+            return template.format(*result, view_type)
```
amjith: This syntax will work in Python 3.5 and above but it won't work in Python 2.7. 

```
SyntaxError: only named arguments may follow *expression 
```

amjith:
I have tested this PR in Python 3.6 and it works beautifully. But there is a syntax error when I try to use it in Python 2.7.

j-bennet:
The PR is rebased, and things seem to work great. I'm going to merge it.

@catherinedevlin Nice work - thanks a lot!

❤️

mkcinu:
How can I quite the editor opened through \ef ... without applying the changes? It openes VI for me, then I :quit and the function is displayed in the prompt waiting for me to hit Enter. How can I discard the query? Ctrl+C triggers AssertionError. Bug?

j-bennet:
@mkcinu Sounds like a bug, yes - please open an issue so we can track it.

PR #888, opened 2018-05-26
make keyring optional

meeuw:
## Description
The keyring code seems to fail if you have gnome-keyring installed but aren't running the gnome-keyring-daemon.

Stack trace:
```
2018-05-26 11:22:50,043 (28714/MainThread) pgcli.main ERROR - traceback: 'Traceback (most recent call last):
  File "pgcli/main.py", line 436, in connect
    keyring.set_password(\'pgcli\', key, passwd)
  File "/lib/python3.6/site-packages/keyring/core.py", line 47, in set_password
    _keyring_backend.set_password(service_name, username, password)
  File "/lib/python3.6/site-packages/keyring/backends/SecretService.py", line 80, in set_password
    collection = self.get_preferred_collection()
  File "/lib/python3.6/site-packages/keyring/backends/SecretService.py", line 61, in get_preferred_collection
    raise InitError("Failed to unlock the collection!")
keyring.errors.InitError: Failed to unlock the collection!
```

## Checklist

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -433,7 +433,10 @@ class PGCli(object):
                 pgexecute = PGExecute(database, user, passwd, host, port, dsn,
                                       application_name='pgcli', **kwargs)
                 if passwd:
-                    keyring.set_password('pgcli', key, passwd)
+                    try:
+                        keyring.set_password('pgcli', key, passwd)
+                    except keyring.errors.InitError:
+                        pass
             except (OperationalError, InterfaceError) as e:
                 if ('no password supplied' in utf8tounicode(e.args[0]) and
                         auto_passwd_prompt):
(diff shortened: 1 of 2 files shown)


pgcli/main.py
```diff
@@ -433,7 +433,10 @@ def connect(self, database='', host='', user='', port='', passwd='',
                 pgexecute = PGExecute(database, user, passwd, host, port, dsn,
                                       application_name='pgcli', **kwargs)
                 if passwd:
-                    keyring.set_password('pgcli', key, passwd)
+                    try:
+                        keyring.set_password('pgcli', key, passwd)
+                    except keyring.errors.InitError:
```
j-bennet: What happens when we silence this `InitError`? It looks like we don't fall back to anything else, the connection just silently fails?

meeuw: No, if I understand #873 correctly it stores the password as a convenience. The down side of this fix is that the user doesn't know why his password isn't stored. On the other side, if the user didn't read the source of pgcli he probably wouldn't even know his password is being saved.

I think this feature (saving passwords) need some work, this is a quick workaround for a failing pgcli on my system. (I should have mentioned this first, sorry)

j-bennet: Got it. It may not be perfect, better than it was before, for sure. Let's merge it.

PR #892, opened 2018-06-01
pgcli bash completion script

antoniotrkdz:
# Pgcli bash completion
Bash completion script for pgcli.
## Description
It allows to <tab> complete database names, users and hosts, plus showing the other available options when calling pgcli from the terminal prompt.
In order to work it steals a couple of functions from psql (postgres CLI client installed by default). 
It is intended to be packaged with the debian package and to be copied on /usr/share/bash-completion/completions/pgcli (filename changes for consistency) upon installation of the package.
Please make the debian package mantainer (Lennart Weller lhw@ring0.de as per `apt-cache show`) aware of it and include it in the relevant section of the installation scripts/package.
I hope it is useful.

## Checklist

### DIFF
--- a/pgcli-completion.bash
+++ b/pgcli-completion.bash
@@ -1,0 +1,61 @@
+_pg_databases()
+{
+    # -w was introduced in 8.4, https://launchpad.net/bugs/164772
+    # "Access privileges" in output may contain linefeeds, hence the NF > 1
+    COMPREPLY=( $( compgen -W "$( psql -AtqwlF $'\t' 2>/dev/null | \
+	    awk 'NF > 1 { print $1 }' )" -- "$cur" ) )
+}
+                                                                                                               
+_pg_users()
+{
+    # -w was introduced in 8.4, https://launchpad.net/bugs/164772
+    COMPREPLY=( $( compgen -W "$( psql -Atqwc 'select usename from pg_user' \
+        template1 2>/dev/null )" -- "$cur" ) )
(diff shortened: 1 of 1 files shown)


antoniotrkdz:
Hi
this is not a part of the pgcli code itself, but a script for unix systems to get useful pgcli command auto completion in the shell.
So if the pgcli command is called from the bash command line, when you press <tab><tab> it autocompletes the names of the databases that are present on the local machine.
It has some context too so it can complete the database names and the user names based on the command option.

I don't know if or how it can be tested with your CI... 

However, you can test it by simply dropping the file into
`/usr/share/bash-completion/completions/pgcli`
(only if you are using bash)
and try to press <tab><tab> when invoking pgcli from the shell to operate on local databases.

mattbrictson:
@antoniotrkdz I'm getting an error trying to use your script on macOS with this version of bash:

```
GNU bash, version 5.0.11(1)-release (x86_64-apple-darwin18.6.0)
```

When a type `pgcli <TAB>` I see:

```
pgcli -bash: _init_completion: command not found
```

Any suggestions?

mattbrictson:
Applying a patch similar to this one fixed it for me on macOS: https://gist.github.com/eparis/fd17b8fb4eb58efc2c12

Thanks for the script!

meeuw:
I think this is very useful! Wouldn't it be better to use click for this purpose?
https://click.palletsprojects.com/en/7.x/bashcomplete/

j-bennet:
Should have merged this long ago. Thank you @ antoniotrkdz.

PR #895, opened 2018-06-06
Hotfix/not enough values to unpack

gma2th:
## Description

Error `not enough values to unpack` is raise when statement is an empty string.

## Checklist


j-bennet:
Nice catch - thank you!

PR #897, opened 2018-06-09
Hotfix/output unbound local error

gma2th:
## Description

Fix unbound local error when destructive_warning is false.

## Checklist


meeuw:
Thank you for your contribution! This is really an improvement although I personally don't like the tri-state of destroy (True, False, None). Could you please explain how to reproduce the error your fixing?

Please don't get me wrong, again, your code really is an improvement over the current situation.

gma2th:
The error is raised when `destructive_warning` is set to False in `pgclirc`.
`confirm_destructive_query` should only be called if the query is destructive, what do you think?

meeuw:
Right, thanks!

PR #901, opened 2018-07-04
Avoid reconnect prompt after error if connection is still valid

### LINKED ISSUE #807: pgcli says "connection reset" on any OperationalError
## Description

On any OperationalError, such as an invalid SQL query, pgsql says "Connection reset. Reconnect?", even though there should be no need to reconnect.

Here is an example of issuing a query to modify `pg_settings` view, for the parameter that cannot be modified in real time:

```
2017-10-21 10:29:05,537 (45148/MainThread) pgcli.pgexecute ERROR - traceback: 'Traceback (most recent call last):\n  File "/Users/irina/src/pgcli/pgcli/pgexecute.py", line 319, in run\n    yield self.execute_normal_sql(sql) + (sql, True)\n  File "/Users/irina/src/pgcli/pgcli/pgexecute.py", line 355, in execute_normal_sql\n    cur.execute(split_sql)\nOperationalError: parameter "unix_socket_directories" cannot be changed without restarting the server\n\n'
2017-10-21 10:29:05,537 (45148/MainThread) pgcli.main ERROR - sql: u"update pg_settings set setting = null where name = 'unix_socket_directories'", error: OperationalError('parameter "unix_socket_directories" cannot be changed without restarting the server\n',)
```

## Your environment

pgcli 1.8.1 from master (https://github.com/dbcli/pgcli/tree/6b75bc272768879ba1043da2acf84913477bf7a0).

saifelse:
## Description
Instead of whitelisting all errors that do not require reconnecting, we simply
only reconnect if we detect a disconnect has occurred.

psql notably behaves in a similar way: https://git.io/fbxuc#L1461

Fixes #807

## Testing
1. In one shell:
    ```
    CREATE TABLE hello ( id integer, PRIMARY KEY(id));
    BEGIN;
    LOCK TABLE hello IN ACCESS EXCLUSIVE MODE;
    ```

2. In another shell:
    ```
    BEGIN;
    LOCK TABLE hello IN ACCESS EXCLUSIVE MODE;
    ```
3. Hitting Ctrl-C, results in: `^Ccanceling statement due to user request`
4. Hitting a lock timeout:
    ```
    ROLLBACK;
    BEGIN;
    SET lock_timeout = 3;
    LOCK TABLE hello IN ACCESS EXCLUSIVE MODE;
    ```
    similarly results in `canceling statement due to lock timeout`

5. Attempting to perform another statement behaves as expected:
    ```
    postgres@localhost:mytestdb> SELECT count(*) FROM hello;
    current transaction is aborted, commands ignored until end of transaction block
   ```
6. Running `SELECT count(*) FROM hello;` after stopping the database
    correctly prompts `Connection reset. Reconnect (Y/n):`

## Checklist

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -81,6 +81,7 @@ Contributors:
     * Rishi Ramraj
     * Matthieu Guilbert
     * Alexandr Korsak
+    * Saif Hakim
 
 
 Creator:
--- a/changelog.rst
+++ b/changelog.rst
@@ -20,7 +20,7 @@ Internal changes:
 Bug Fixes:
 ----------
 * Disable pager when using \watch (#837). (Thanks: `Jason Ribeiro`_)
-* Don't offer to reconnect when we can't change a param in realtime (#807). (Thanks: `Amjith Ramanujam`_)
+* Don't offer to reconnect when we can't change a param in realtime (#807). (Thanks: `Amjith Ramanujam`_ and `Saif Hakim`_)
 * Make keyring optional. (Thanks: `Dick Marinus`_)
 * Fix ipython magic connection (#891). (Thanks: `Irina Truong`_)
 * Fix not enough values to unpack. (Thanks: `Matthieu Guilbert`_)
--- a/changelog.rst
+++ b/changelog.rst
@@ -837,3 +837,4 @@ Improvements:
 .. _`Rishi Ramraj`: https://github.com/RishiRamraj
 .. _`Matthieu Guilbert`: https://github.com/gma2th
 .. _`Alexandr Korsak`: https://github.com/oivoodoo
+.. _`Saif Hakim`: https://github.com/saifelse
--- a/pgcli/pgexecute.py
+++ b/pgcli/pgexecute.py
@@ -353,22 +353,17 @@ class PGExecute(object):
 
     def _must_raise(self, e):
         """Return true if e is an error that should not be caught in ``run``.
-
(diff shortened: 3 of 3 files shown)


saifelse:
I realize #807 is already closed, but I think this fix avoids the issue more generally —  https://github.com/dbcli/pgcli/issues/807#issuecomment-392115598 reveals that the original fix doesn't handle Ctrl-C

j-bennet:
This sounds like a more sensible way to handle the problem. Very nice. Let me test it out, and I'll merge.

saifelse:
@j-bennet : One worry is that this code path is no longer exercised:  https://codecov.io/gh/dbcli/pgcli/pull/901/changes

Unfortunately,  I don't have enough context to know if that is an unintended consequence of my change. From what I gather, if KeyboardException does cause the connection to close, and we're catching it in pgexecute.py, then we'd wouldn't be able to catch it in main.py.

j-bennet:
This works very well. I am going to merge it. We can deal with any new `KeyboardInterrupt` problems later, if they do come up.

Also, this PR wins the "best first contributor PR" award, because the description is amazing and I love the test cases being included! Here it is: 🥇

Thank you!

saifelse:
😊

Also, just wanted to say thanks for making it so easy to contribute and for building such a great tool. We recently made the switch from psql to pgli... and you can literally hear the delight around the office "What, we have syntax highlighting and autocomplete now?! This is the best."

PR #916, opened 2018-07-21
make keyring optional

meeuw:
## Description
make keyring optional

## Checklist

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -44,7 +44,10 @@ from pygments.token import Token
 
 from pgspecial.main import (PGSpecial, NO_QUERY, PAGER_OFF)
 import pgspecial as special
-import keyring
+try:
+    import keyring
+except ImportError:
+    keyring = None
 from .pgcompleter import PGCompleter
 from .pgtoolbar import create_toolbar_tokens_func
 from .pgstyle import style_factory, style_factory_output
@@ -72,6 +75,7 @@ import psycopg2
 
 from collections import namedtuple
 
+from textwrap import dedent
 
 # Query tuples are used for maintaining history
 MetaQuery = namedtuple(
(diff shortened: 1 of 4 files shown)


pgcli/main.py
```diff
@@ -440,11 +461,22 @@ def connect(self, database='', host='', user='', port='', passwd='',
             try:
                 pgexecute = PGExecute(database, user, passwd, host, port, dsn,
                                       application_name='pgcli', **kwargs)
-                if passwd:
+                if passwd and keyring and self.config["main"].get("keyring", True):
                     try:
                         keyring.set_password('pgcli', key, passwd)
-                    except (keyring.errors.InitError, RuntimeError):
-                        pass
+                    except (
```
j-bennet: This keyring-related handling is getting long enough to be extracted into methods, or perhaps a class that `main` can delegate things to. Our `main` is very clattered and it's becoming difficult to follow the logic.

pgcli/main.py
```diff
@@ -440,11 +461,22 @@ def connect(self, database='', host='', user='', port='', passwd='',
             try:
                 pgexecute = PGExecute(database, user, passwd, host, port, dsn,
                                       application_name='pgcli', **kwargs)
-                if passwd:
+                if passwd and keyring and self.config["main"].get("keyring", True):
```
j-bennet: I would add the default `keyring = True` to `pgclirc`, for better discoverability, and I'd also put in a comment explaining what it is. People tend to rely on `pgclirc` to tell them what `pgcli` can do.

meeuw: Ah yes, thanks for reminding. I forgot to change pgclirc

pgcli/main.py
meeuw: Hmm yes, but I don't like to generalize a generalization (which I think keyring is). You're right about it's getting long and `main.py` is flooded with exceptions like this. But I don't know how that can be fixed.

I've moved `keyring.set_password` so it only has to be called once.

j-bennet:
Merging! Release coming up.

Issue #410, opened 2015-11-11
Transactions


ThiefMaster:
It would be nice if pgcli had some fancy handling for transactions. Things I could imagine:
- Always start a transaction on startup. COMMIT/ROLLBACK would automatically start a new one.
- Start a transaction implicitly as soon as you run any SQL command that modifies data (might be a bit obscure though).

When always having a transaction an automated ROLLBACK+BEGIN in case of an SQL error would be useful - that's actually one of the things I miss most in psql. Sometimes I want to try things and having to rollback+begin after a typo etc it somewhat annoying.

An undo feature could also be useful. Basically it'd ROLLBACK+BEGIN and then execute all statements since the start of the transaction besides that last one / last n ones again.

amjith:
Interesting request. 

But I'm still unclear when pgcli would commit. For example: When a user launches pgcli we start a transaction, then they explore using select, change data using insert or update or drop, then explore some more using select. In this workflow where does the commit happen? Do we keep all these commands in a transaction until they quit pgcli or type in COMMIT/ROLLBACK themselves? 

@darikg I was thinking your latest changes to add the extra field for denoting success/failure of a query can be used to achieve the undo feature.

ThiefMaster:
> But I'm still unclear when pgcli would commit. 

I would only commit explicitly, i.e. by issuing a `COMMIT` statement. Of course this would need to be displayed on startup when transactions are enabled (and there could be an *optional* prompt on exit if there are uncommitted changes).

fernandomora:
I think this would be very useful as a mode that can be enabled/disabled with its status shownin the bottom bar like smart completion and multiline mode. When autocommit mode is ON its behaviour would be the standard one when no transaction is explicly started and when it is OFF, a commit is explicitly needed.

Improving autocommit off mode with an automatic start of a new transaction (ROLLBACK+BEGIN) when an errors fail as @ThiefMaster purpose would be great.

j-bennet:
This sounds like a nice feature to have. I like the idea of autocommit mode being displayed in the bottom status bar. We could also add a little indicator (like a red `*` or `!`) to show that user has uncommited changes - in addition to a warning on exit.

fedragon:
I would love to see this feature coming to pgcli, I'm currently dealing with `begin; ... commit;` on my own but that makes it really prone to error, especially when dealing with production databases...

mkataja:
[...]

Issue #535, opened 2016-06-24
Command history configuration/HISTFILE option


Seanstoppable:
In my psqlrc file, I have the following:

\set HISTFILE ~/.psql_history- :DBNAME

This allows for a history file per database I connect to, making it easy to get the history of commands for that database.

It would be nice if that were supported, ideally from parsing .psqlrc, to make switching easier.

Similarly, it would be nice if the history were done the same way as psql's (which is unfortunately not time stamped), again to ease transition and allow a new user to just start up pgcli and get their existing history.

hunsakerbn:
For me a history-file-per-database ability is a must.  I connect to a lot of different databases and I don't want history from other DBs showing up when working with my current DB.  This is going to keep me from using this great product.

anthonydb:
Adding my +1 for this as well as for `HISTCONTROL ignoredups` support. Love pgcli!

eliottwiener:
Here's a dumb, hacky workaround for those who want history-file-per-database with the current release version of pgcli.

If, like me, you formerly used the environment variable `PSQL_HISTORY` (or any other variable) to configure your history-file-per-database setup, you can launch pgcli like so:

```
pgcli --pgclirc =(sed 's@history_file =.*@history_file = '$PSQL_HISTORY'@' ~/.config/pgcli/config) your_database_name
```

To explain briefly: The above creates a new temporary config file that is the same as the actual pgcli config file, only with the `history_file = ...` setting replaced to point to our desired history file (`$PSQL_HISTORY` in this case). Then pgcli is launched with the temporary config file.

The `=(...)` is a zsh-ism (sorry) for creating a temporary file, and is used in a similar fashion to `<(...)` of bash and others. In this case `<(...)` seems to not be suitable, possibly due to the way pgcli reads the config file. `=(...)` can easily be emulated in other shells. See: http://zsh.sourceforge.net/Doc/Release/Expansion.html#Process-Substitution

Create an alias or shell function incorporating the above approach, and you're good to go.

A more convenient solution would be to create a shell function that extracts the db name (and maybe host, etc) from the command line, and incorporates that into the `history_file` path used.

viniciusd:
I am currently struggling switching among different databases. A per-database history would be neat

laixintao:
I'd like to pick this four-years-ago feature request up. 😀

laixintao:
I have been thinking this feature for days. So here's what I want to do:

[...]

PR #917, opened 2018-07-22
Autocommit mode

gma2th:
## Description

Add autocommit mode. 
Add on error rollback. 
When autocommit_mode is set to false, users can manually commit or rollback the current transaction.
When on_error_rollback is set to true, if a statement in a transaction block generates an error, the error is ignored and the transaction continues.
In order to deactivate autocommit for users transactions, without messing up with intern pgcli transactions, we better use two different connections.
Requested here https://github.com/dbcli/pgcli/issues/410

## Limitation
Users are not able to activate/disable autocommit from the UI.
Autocommit has to be activated in order to use on error rollback.
No tests to cover the change.

## Ressource 
- https://www.postgresql.org/docs/current/static/ecpg-sql-set-autocommit.html
- https://www.postgresql.org/docs/current/static/tutorial-transactions.html
- https://www.endpoint.com/blog/2015/02/24/postgres-onerrorrollback-explained

## Checklist


j-bennet:
Unit tests are failing with a unicode error there. Are you able to run them locally (running `py.test` in pgcli directory)?

lelit:
I think that the reason of the failure is that the new `get_new_connection()` method should install the same typecasters as done by `connect()`... Maybe some refactor could be made, to avoid duplicating basically the same code to create the connection. 

Also, I'm not sure if those typecasters should be moved only onto the new `user_conn`: I would expect that only _user_ queries will involve `JSON` values...

Finally, in the method `execute_normal_sql()` (the only place where the new `user_conn` gets used), there's a loop that pops possible `notices`: shouldn't it targeting the same `user_conn` connection?

gma2th:
Thank you for having a look at this PR.
I fixed the unicode error (due to typecasters) and notices, thank you.
It definitely need some refactor, in order to init all connections in the same place. I can have a look in the next days. This could be done in another PR, up to you.

amjith:
@gma2th I looked through the PR and I have 2 questions. Let me preface these comments by saying, I would like to minimize the number of configuration options in pgcli. It is wise to choose sensible defaults wherever possible. 

1. What is the purpose of surfacing the autocommit mode as a config? Is there an advantage in disabling autocommit mode? From my basic understanding we want to have the autocommit mode enabled at all times and the user can specify the `BEGIN` command to explicitly start a transaction.

2. Why is the on_error_rollback set to `off` by default? Wouldn't it make sense to have that `on` all the time?

amjith:
Hrm. Even when I set the `on_error_rollback` set to `True` it ends up aborting the transaction. I haven't traced the issue yet. Can you please verify if it works for you?

Gollum999:
Any updates on this?  I've been interested in this feature for quite a while.

samskeller:
@amjith @gma2th any updates on getting this through? Not being able to turn off autocommit for `pgcli` is basically a dealbreaker for me and many others who want some safety from fat-fingering!

quezak:
Hello, could we get this merged? I want to be able to set autocommit off when accessing _important_ databases :)

j-bennet:
@quezak This PR has a couple of outstanding questions, and a rebase on master. We've heard nothing from @gma2th in a long time, and I don't think they are interested in finishing the work.

Perhaps you'd be interested to pick up and continue working on this feature, by forking pgcli master and porting over these changes?

quezak:
I see. Let's see if we're able to solve the questions and I could be able to help if it's not too complicated :)

> What is the purpose of surfacing the autocommit mode as a config? 

I'd like to have autocommit always disabled when I access a production database that's generally not operated manually -- just in case I hit enter without thinking twice :) But on a dev or local instance I'm fine with having autocommit on.

>  I would like to minimize the number of configuration options

I agree, but also I'd like to have _some_ way to disable autocommit for some databases as described above. I think it would be at least good to limit the number of CLI parameters -- this PR proposes two of them, #535 another one, and at a glance I see more cases in open issues. Maybe it would be nice to come up with a more general solution to allow overriding _some_ config vars from CLI and/or per different databases? I've dropped a few ideas in https://github.com/dbcli/pgcli/issues/535#issuecomment-726653656, but I don't know where would be a good place to discuss it.

Oliver-Fish:
If anyone could work on this and get it merged, it would be greatly appreciated. This increased safety would be beneficial. I am not super familiar with Python. Otherwise, I would consider jumping and trying this myself.

Issue #919, opened 2018-07-24
PG 11 changed the structure of pg_catalog.pg_proc


lelit:
The method `PGExecute.functions` will need some tweak to support PG 11 (now in beta): it queries the `pg_catalog.pg_proc` relation to extract a list of known functions, and in the coming version the `proisagg` and `proiswindow` columns have been collapsed into a single `prokind` enum, see [doc](https://www.postgresql.org/docs/11/static/catalog-pg-proc.html).

I can't promise, but I'll try to propose a fix if I find a couple of spare hours :smile:

PR #921, opened 2018-07-24
Fix issue #919

lelit:


lelit:
AFAICT the failure is not related to my changes... that OTOH I think aren't even tested by Travis: we should activate a _matrix_ to check against different version of PostgreSQL. One project I know doing that is [asyncpg](https://github.com/MagicStack/asyncpg/blob/master/.travis.yml#L40).

j-bennet:
@lelit Yes, that test failure is intermittent, things are fine now. Would you be able to rebase this PR on top of master? Otherwise, it looks good to me.

lelit:
There it is, thank you!

j-bennet:
👍 Looks good. Merging!

PR #928, opened 2018-07-27
Tab press on an empty line increases the indentation instead of triggering the auto-complete pop-up

arturbalabanov:
## Problem

Writing complicated queries on multiple lines can be quite frustrating because the `Tab` key _always_ triggers the auto-complete pop-up, instead of increasing the indentation when on empty lines.

This seems to be undesired behaviour as at these cases one would rarely want a general auto-completion window open. Far more useful would be to increase the indentation without resorting to the vi mode or using an external editor.

## Solution

This MR changes the `Tab` mapping behaviour, so that if it's pressed on an empty line (with or without whitespace), it simply inserts a `\t` character. The auto-completion behaviour is triggered in all the other cases.

Furthermore, since some users would prefer spaces instead of `\t` characters (for copy-pasting queries for example), an `expand_tab` option has been introduced: it it's enabled, 4 spaces are inserted instead.

## Checklist


j-bennet:
Thank you for the PR! Your build fails on pep8 checks. You can fix pep8 problems with this command:
```
pip install git+https://github.com/hayd/pep8radius.git
pep8radius master --docformatter -i
```

arturbalabanov:
@j-bennet, fixed the PEP8 problems, now the build should succeed.

j-bennet:
@arturbalabanov I have a couple of problems with how this currently works:

* I can see how inserting whitespace can be useful in multiline mode. But not in single-line mode.
* In both single-line and multi-line mode, when the cursor is immediately at prompt (like `user@localhost> [HERE]`), I think the correct behavior is autocomplete, not insert whitespace.

arturbalabanov:
@j-bennet, thank you for the feedback!

> In both single-line and multi-line mode, when the cursor is immediately at prompt (like `user@localhost> [HERE]`), I think the correct behavior is autocomplete, not insert whitespace.

Yes, this does make sense, fixed in 7602998 -- on the first line autocomplete is triggered always.

> I can see how inserting whitespace can be useful in multiline mode. But not in single-line mode.

I'm a bit reluctant to restrict the new `Tab` key behaviour to multi-line mode only -- the above fix should capture any reasonable single-line mode editing, and there is one weird edge case:

* Go to single-line mode
* Turn on vi-mode
* Hit `Esc` to go to Vi's normal mode
* Hit `o` to insert a new line

In this scenario, even though we're in single-line mode, the query consists of multiple lines and it seems reasonable and intuitive to me to allow the `Tab` key to increase the indentation. Though, I am not sure whether this is the correct behaviour of vi-mode. Please let me know your thoughts on that.

j-bennet:
I did not forget about this PR, but I'm working on switching to prompt-toolkit 2.0: https://github.com/dbcli/pgcli/pull/930.

With 2.0, the multiline buffer behavior is not quite the same. I am going to leave this PR open until 2.0 is merged and then see if this change still makes sense / can be applied.

pgcli/pgclirc
```diff
@@ -143,6 +143,9 @@ enable_pager = True
 # Use keyring to automatically save and load password in a secure manner 
 keyring = True
 
+# When the <Tab> key is pressed on an empty line, use 4 spaces instead of \t
+expand_tab = False
```
amjith: Can we pick one and make it a default? I really hate adding new config options. 

Given that pgcli is a REPL, I honestly don't think users will care if we choose tab or spaces.

amjith:
Merging this for now. I will remove the config option after the merge and I'll pick expand_tab as the default.

PR #930, opened 2018-07-28
Cherry-picked prompt-toolkit 2.0 changes.

j-bennet:
Changes for prompt-toolkit 2.0, based on current master.

Cherry-picked from https://github.com/dbcli/pgcli/pull/826.

@jonathanslenders I cherry-picked because I have no permissions to rebase your branch, but we can merge either PR, provided it's in line with current `pgcli` master.

**Problems noticed so far**:

* ~styles are not applied~ fixed
* cli takes longer to start. I had to increase "show help" integration test timeout.
* ~cli won't run under 2.7~ fixed, it was unicode literals
* ~`iocommands` integration test fail on editor command~ fixed
* ~`named_queries` integration test fail~ fixed

Named query integration test failed because of this warning printed under `pexpect` (which is not a real terminal):

https://github.com/jonathanslenders/python-prompt-toolkit/blob/b8d09e971e5fbad76d3c5bda4b47063e4c0644b8/prompt_toolkit/application/application.py#L706

@jonathanslenders Is there a way to suppress this kind of error output for tests? perhaps is you use `warnings` vs writing to stdout?


j-bennet:
Here is what I mean by "styles not applied"

pgcli with ptk 1.0:

pgcli-1 0-style

pgcli with ptk 2.0:

pgcli-2 0-style

j-bennet:
I tried to debug the problem with styles again today. I did not get very far, but it seems to me that styles are kind of applied, albeit incorrectly. At least in case of bottom toolbar, they seem reversed. For example, when styling bottom toolbar with red text and blue background, here is what I get:

red-blue-toolbar

here are my config colors:

```
Token.Toolbar = '#000000 bg:#0000ff'
Token.Toolbar.Off = '#000000 bg:#0000ff' 
Token.Toolbar.On = '#ff0000 bg:#0000ff' 
```

j-bennet:
I found the reversing bit:

https://github.com/jonathanslenders/python-prompt-toolkit/blob/b7ab3ba67df2e7dbf63463fb98f817878dfd256a/prompt_toolkit/styles/defaults.py#L129

and added an override for that.

Now to figure out what's wrong with other styles...

jonathanslenders:
Hi @j-bennet,

The bottom toolbar is reversed by default, in order to make it stand out on both dark and light background terminals. You can indeed override it like that.

The backward-compatible changes that you've added there, is that to support custom color schemes in the configuration files? I think it looks fine. Ideally (in the future), you want to switch to the new style dictionaries that don't use Pygments tokens.

j-bennet:
@jonathanslenders Yes, I added changes to convert from old-style Pygments tokens to new-style class names, so that people don't have to change their existing config files. But I need you to tell me what other token names I should convert to what class names. I migrated 1):

```
Token.Menu.Completions.Completion.Current = 'bg:#ffffff #000000'
Token.Menu.Completions.Completion = 'bg:#008888 #ffffff'
Token.Menu.Completions.Meta.Current = 'bg:#44aaaa #000000'
Token.Menu.Completions.Meta = 'bg:#448888 #ffffff'
Token.Menu.Completions.MultiColumnMeta = 'bg:#aaffff #000000'
Token.Menu.Completions.ProgressButton = 'bg:#003333'
Token.Menu.Completions.ProgressBar = 'bg:#00aaaa'
```

but not 2):

```
Token.SelectedText = '#ffffff bg:#6666aa'
Token.SearchMatch = '#ffffff bg:#4444aa'
Token.SearchMatch.Current = '#ffffff bg:#44aa44'
Token.Toolbar = 'bg:#222222 #aaaaaa'
Token.Toolbar.Off = 'bg:#222222 #888888'
Token.Toolbar.On = 'bg:#222222 #ffffff'
Token.Toolbar.Search = 'noinherit bold'
Token.Toolbar.Search.Text = 'nobold'
Token.Toolbar.System = 'noinherit bold'
Token.Toolbar.Arg = 'noinherit bold'
Token.Toolbar.Arg.Text = 'nobold'
Token.Toolbar.Transaction.Valid = 'bg:#222222 #00ff5f bold'
Token.Toolbar.Transaction.Failed = 'bg:#222222 #ff005f bold'
```

What class/style names can I use for the list 2) above?

jonathanslenders:
This contains a good overview of all tokens that had a certain style: https://github.com/jonathanslenders/python-prompt-toolkit/blob/1.0/prompt_toolkit/styles/defaults.py#L16
This contains the new default style rules:

The following tokens match to these class names:
```
Token.SelectedText -> 'selection'
Token.SearchMatch -> 'search'
Token.SearchMatch.Current -> 'search.current'
```

The other toolbar tokens were never defined in prompt_toolkit. They correspond to custom pgcli styling. I see that the bottom toolbar is wrapped in ``PygmentsTokens``. This means that these will be turned into the following prompt_toolkit classnames:
```

Token.Toolbar.Off = 'pygments.toolbar.off'
Token.Toolbar.On = 'pygments.toolbar.on'
Token.Toolbar.Search = 'pygments.toolbar.search'
Token.Toolbar.Search.Text = 'pygments.toolbar.search.text'
...
```
Ideally, in the future (or now?), we should convert that to a list of prompt_toolkit ``(style, text)`` tuples, where ``style`` is something like ``"class:toolbar.off"``. In that case, the pygments prefix is dropped.

j-bennet:
@jonathanslenders Thanks, that helps a lot, especially the link to old token definitions! I re-mapped everything to new style names, even though we're still backwards compatible with tokens. I also got rid of `PygmentsTokens` when creating the cli toolbar.

@amjith @meeuw @tsroten I would appreciate another look from you at this PR.

@tsroten would you be able to migrate `cli_helpers` to 2.0? Right now I'm still passing a dict of `{pygments token -> style string}` to `formatter.format_output`, but we should probably get rid of tokens in `cli_helpers` now that they are phased out of `prompt_toolkit`. This does not have to be done right now.

meeuw:
Hmm in emacs mode `ctrl-x-ctrl-e` (open in editor) doesn't seem to work anymore (works in master)

Other than that I haven't found any problems, nice work!

j-bennet:
@meeuw Nice find! It looks like it takes a non-default flag to open things in editor in 2.0.

pgcli/main.py
```diff
@@ -506,24 +503,25 @@ def handle_editor_command(self, cli, document):
...
+        :param text: Document
         :return: Document
         """
+        app = self.prompt_app.app
+
         # FIXME: using application.pre_run_callables like this here is not the best solution.
         # It's internal api of prompt_toolkit that may change. This was added to fix #668.
         # We may find a better way to do it in the future.
-        saved_callables = cli.application.pre_run_callables
+        saved_callables = app.pre_run_callables
```
meeuw: @jonathanslenders removed this workaround in the mycli branch for python prompt toolkit, is this fixed in prompt toolkit 2.0?

pgcli/main.py
```diff
@@ -692,68 +682,54 @@ def prompt_tokens(_):
                     len(prompt) > self.max_len_prompt):
                 prompt = self.get_prompt('\\d> ')
 
-            return [(Token.Prompt, prompt)]
```
meeuw: In mycli branch this was rewitten as `return [('class:prompt', prompt)]` I don't know if that would be better.

pgcli/main.py
```diff
@@ -215,8 +213,7 @@ def __init__(self, force_passwd_prompt=False, never_passwd_prompt=False,
         self._completer_lock = threading.Lock()
         self.register_special_commands()
 
-        self.eventloop = create_eventloop()
-        self.cli = None
+        self.prompt_app = None
```
meeuw: Maybe this can be removed (or changed to a prompt_app_initialized boolean) as we're using a `get_app()` singleton now.

pgcli/main.py
j-bennet: 👍

pgcli/main.py
j-bennet: I'm not sure about that one. Seems to me like one of those things that's not broken and no reason to fix.

j-bennet:
I think this is ready to merge. Can I get some 👍 or 👎 from @dbcli/pgcli-core ?

lelit:
I'm sorry Irina, but I could do only a very quick and simple check on the branch (unfortunately where I'd use it most there's another tool that depends on PT 1, so they cannot co-exist). I did not hit any problem, though :-)

meeuw:
From tomorrow I'll be testing these changes I've ported to mycli (I don't have any postgresql production database :cry:).

meeuw:
Pressing `F2` gives me `Exception 'DynamicCompleter' object has no attribute 'smart_completion'`

amjith:
> From tomorrow I'll be testing these changes I've ported to mycli (I don't have any postgresql production database).

I've been using https://www.elephantsql.com/ for testing pgcli. 

Their free plan is good enough for a toy database.

meeuw:
>I've been using https://www.elephantsql.com/ for testing pgcli.
>
>Their free plan is good enough for a toy database.
That looks nice, thanks! I already have a test postgresql database using https://hub.docker.com/_/postgres/ but what I meant was that I don't have a "normal workflow" with a production postgresql. When I test postgresql it's a bit like dry land swimming.

meeuw:
`ctrl-c` seems to Abort `pgcli`, I use this often when I want to cancel the query I'm typing.

pgcli/main.py
meeuw: sorry for the confusion; prompt_app is a `prompt_toolkit.shortcuts.PromptSession`, which cannot be reached from the get_app() singleton.

j-bennet: ah ok, good to know!

j-bennet:
@eirnym prompt toolkit supports formatted text in prompt:

https://python-prompt-toolkit.readthedocs.io/en/master/pages/asking_for_input.html#coloring-the-prompt-itself

and lots of way to create formatted text:

https://python-prompt-toolkit.readthedocs.io/en/master/pages/printing_text.html#formatted-text.

`pgcli` uses prompt toolkit to build a cli, so we support whatever is supported there.

meeuw:
whoops, @j-bennet you're right, I've found some Vi-mode regression.

Steps to reproduce:
- Enable Vi-mode (F4)
- enter `select`
- press `ESC` `v`
- exit editor

Expected result:
Vi commands should still work (like enter `select` and press `ESC` again)

Actual result:
Vi command don't work, Vi-mode is still enabled according to toolbar

j-bennet:
@meeuw Nice find!

Jonathan also links to a pdf here (about vi mode in readline):

https://github.com/jonathanslenders/python-prompt-toolkit/blob/fc4d8a0f21077d25285779ed3abe750ea7ae2d28/prompt_toolkit/key_binding/bindings/vi.py#L304

http://www.catonmat.net/download/bash-vi-editing-mode-cheat-sheet.pdf

I'll try and see which of those cheatsheet commands work (if any).

j-bennet:
@meeuw It's not just editor, it looks like the vi mode is never applied to this prompt session after the first invocation. This commit seems to fix the problem:

https://github.com/dbcli/pgcli/pull/930/commits/31b6eb1b2987e70314e24631df33f0f28a8602f6

but I think this fix should not be needed. Once we change vi mode here:

https://github.com/dbcli/pgcli/blob/31b6eb1b2987e70314e24631df33f0f28a8602f6/pgcli/key_bindings.py#L32

shouldn't it propagate from `Application` back to our `PromptSession`?

@jonathanslenders can you advise?

meeuw:
Good find @j-bennet! I've posted a PR for prompt_toolkit: https://github.com/jonathanslenders/python-prompt-toolkit/pull/720

eirnym:
@j-bennet I've asked as you can request library to disable or enable some features programatically and I want to be aware about the particular one.

j-bennet:
After more poking at this PR, I don't see anything else broken. I feel like merging it. What do you think @meeuw? Sure, once it's released, people may find other problems, but at least we'll have more eyes out looking for those problems.

meeuw:
I think it's good to go!

j-bennet:
Linking another regression here: https://github.com/dbcli/pgcli/pull/970

PR #959, opened 2018-10-29
Set default port in `connect_uri` when none is given, fixes #958

### LINKED ISSUE #958: \p in prompt resolves to `None`
When no port in the connection string is given - like in...

`pgcli postgresql://postgres:password@localhost`

...`\p` resolves to `None` in the printed prompt:

`postgres@localhost:None/postgres pgcli>`

`Mssql-cli` and `Mycli` don't show this behaviour.

DanEEStar:
## Description
When there is no port given in a connection URI the sytem should assume the default port 5432.

## Checklist

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -895,7 +895,8 @@ class PGCli(object):
         short_host, _, _ = host.partition('.')
         string = string.replace('\\h', short_host)
         string = string.replace('\\d', self.pgexecute.dbname or '(none)')
-        string = string.replace('\\p', str(self.pgexecute.port) or '(none)')
+        string = string.replace('\\p', str(
+            self.pgexecute.port) if self.pgexecute.port is not None else '5432')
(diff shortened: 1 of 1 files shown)


j-bennet:
Hi, and thank you for the PR! I think this is fixing the problem in the slightly wrong place. We do not need to set the port before connecting, because `psycopg2` will connect without provided port just fine. We just need to change what we display in this case here:

https://github.com/dbcli/pgcli/blob/master/pgcli/main.py#L898

DanEEStar:
@j-bennet thanks for the hints. I changed it at the corresponding place in `get_prompt`.
As far as I researched also `psycopg2` will assume port 5432 when none is given, so display port 5432 in this case should make sense.

j-bennet:
Should not have been so quick to merge - forgot about changelog. :) Fixed here: https://github.com/dbcli/pgcli/pull/960

Thanks again!

PR #977, opened 2018-12-02
Time execution separately.

j-bennet:
## Description

This adds a separate timing for query execution. It is useful to see how long things take on database side, and how long on pgcli side (formatting etc.). I am slightly worried that seeing this may make some people unhappy, but on the other hand, it can help us make pgcli better.

## Checklist


amjith:
When I pull down this branch and try it out I don't see the two separate timing output. Am I supposed to see two timing output?

j-bennet:
You'll see `Time: xxx, executed in: yyy` only if total time > 1 sec.

amjith:
```
Time: 2.074s (2 seconds), executed in: 2.067s (2 seconds)
```

Looks good to me. 

:anchor:

PR #978, opened 2018-12-05
Support multihost connection string

wronglink:
## Description
Hi there. I've found out that pgcli doesn't support postgres multihost connection string ([pg docs](https://www.postgresql.org/docs/current/libpq-connect.html#id-1.7.3.8.3.6)):
```
$ pgcli postgresql://host1:123,host2:456/somedb?target_session_attrs=any
  File "/usr/local/bin/pgcli", line 11, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python3.5/dist-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.5/dist-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.5/dist-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.5/dist-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.5/dist-packages/pgcli/main.py", line 1028, in cli
    pgcli.connect_uri(database)
  File "/usr/local/lib/python3.5/dist-packages/pgcli/main.py", line 392, in connect_uri
    port=fixup_possible_percent_encoding(uri.port),
  File "/usr/lib/python3.5/urllib/parse.py", line 158, in port
    port = int(port, 10)
ValueError: invalid literal for int() with base 10: '123,host2:456'
```

I've found out that `urlparse` used for connection URI parsing doesn't support multiple hosts. I think that it would be convenient to switch to URI parsing function that comes with psycopg2 as pgcli already uses it as pg driver. I also provided extra test for multihost connection string.

## Checklist


j-bennet:
Very nice. Thank you for the PR! Tests are especially appreciated. 🍓 

I noticed that the prompt looks a little weird when specifying multiple hosts:

```
(pgcli) --- src/pgcli ‹wronglink-master*?› » pgcli postgresql://localhost,lo.co.com,127.0.0.1/test
Server: PostgreSQL 10.6
Version: 2.0.1
Home: http://pgcli.com
irina@localhost,lo:test>
```
Maybe it makes sense to just display the first value for the host and port?

wronglink:
Thanks for a quick answer.

> Maybe it makes sense to just display the first value for the host and port?

There could be a problem displaying the wrong host (not the one you have connected to), like if the first host didn't answer or if it is filtered by some query parameters. For example, you can add `?target_session_attrs=read-write` to connect only read-write replica. I think it would make more sense to display the actual host you have connected to. I'll dig how to support it.

wronglink:
I've investigated that there is no correct way to identify current connected host in psycopg yet. There is closed issue about it: https://github.com/psycopg/psycopg2/issues/726 and after psycopg v2.8 released, pgcli can switch to [connection.info](
http://initd.org/psycopg/docs/connection.html#connection.info) property that returns [ConnectionInfo](http://initd.org/psycopg/docs/extensions.html#psycopg2.extensions.ConnectionInfo) object and init (or event proxy) PGExecute connection properties through it. So I implemented your idea to pick up the first host from the list.

I also slightly refactored PGExecute init method, to specify all the connection parameters after the connection is established and moved the `short_host` generation logic to a `PGExecute.short_host` property. What do you think of these changes?

pgcli/pgexecute.py
```diff
@@ -186,15 +186,15 @@ class PGExecute(object):
...
+        self.conn = None
+        self.dbname = None
+        self.user = None
+        self.password = None
+        self.host = None
+        self.port = None
+        self.dsn = None
         self.server_version = None
-        self.connect()
+        self.connect(database, user, password, host, port, dsn, **kwargs)
```
j-bennet: Why this change?

pgcli/pgexecute.py
```diff
@@ -276,6 +265,15 @@ def connect(self, database=None, user=None, password=None, host=None,
         register_json_typecasters(self.conn, self._json_typecaster)
         register_hstore_typecaster(self.conn)
 
+    @property
```
j-bennet: +1 for the property.

tests/test_pgexecute.py
```diff
@@ -400,3 +400,13 @@ def test_nonexistent_view_definition(executor):
         result = executor.view_definition('there_is_no_such_view')
     with pytest.raises(RuntimeError):
         result = executor.view_definition('mvw1')
+
+
+@dbtest
```
j-bennet: 👍

pgcli/pgexecute.py
wronglink: It looked to me quite strange that we:
1. set properties at `__init__`
2. than pass it to `connect` method that also accepts all the arguments or takes if from object properties
3. and finally in `connect` sometimes sets properties at the end of it execution

I think it would be much easier to get the data flow if we set all the parameters at `__init__` method and use them in the `connect` method leaving it without parameters or if we don't set anything in the `__init__` and pass all the parameters to the `connect`.

What do you think of this change? If I didn't convince you I'll but the current logic back.

j-bennet: The current pgexecute does what you describe in 1st scenario. It sets class member variables in `__init__`, and then calls `connect()` without parameters:

https://github.com/dbcli/pgcli/blob/7b03f8e2043682aa2d1c06e0f7e8258b2f9c43c5/pgcli/pgexecute.py#L188-L197

Your version conforms to the 2nd scenario (we don't set anything in `__init__` and pass all the parameters to `connect`), so I was wondering why this is better?

The only difference I see between the two is that in 2nd scenario, if the initial connection fails, you can't call `connect()` without parameters to retry. Otherwise it's pretty much the same.

wronglink: I think I could miss the point of what `PGExecute` properties are used for? Which of cases is more relevant, use them as connection parameters to establish new one or represent the state of an already established connection?

> The current pgexecute does what you describe in 1st scenario. It sets class member variables in __init__, and then calls connect() without parameters:

Yes, but it does it twice: once at the `__init__` and second at the `connect`. When I first read the code it was harder for me to predict behaviour of those 2 methods. Which of them sets the state of an object and which initiates the connection? If i don't pass `database` parameter  to `connect` what value would be used for connection? Now it's quite clear to me, but that was the first impression :-)

> Your version conforms to the 2nd scenario (we don't set anything in __init__ and pass all the parameters to connect), so I was wondering why this is better?

The idea was also to prepare to get the actual values of connection for the lower psycopg API (as v2.8 is released). What should we put into the `host` property of an `PGExecute` in case of multiple hosts? The original multi-host string or the final (the host we have connected to? What if we connected via DSN string should we reconnect using the same DSN string or using the parameters we parsed from it?

> if the initial connection fails, you can't call connect() without parameters to retry

I think it depends on design of `PGExecute`. If it stores initial values - it can reconnect itself, otherwise we need to store initial parameters somewhere outside.

j-bennet:
I'm not against the refactor. I'm just trying to figure out the impact of it, i.e. how the new flow is different from the old one, and if it's going to break somewhere unexpected.

The old pgexecute was written in a way that would allow it to support the following use cases:

* establish a connection on `__init__`
* store connection properties (user, database etc) within `pgexecute` and return them to the cli as needed
* execute queries with established connection
* connect to a different database (if asked) without recreating `pgexecute` object and store the new state
* reconnect with the same parameters if connection was dropped

Does the refactor affect any of the use cases?

wronglink:
> Does the refactor affect any of the use cases?

It brakes the only one case: 

> reconnect with the same parameters if connection was dropped

as it doesn't store the original parameters. In common cases initial parameters are the same as parameters of established connection, but they could be different in case of multi-host connection. Reconnection would require the list of all initial hosts, but the object now stores only the connected one. The current `PGExecute` implementation has a problem of storing 2 values in one place: the initial values (before connection) and the actual one (after connection).

j-bennet:
So looks like there's 2 ways we could handle the dropped connection use case in your refactor PR:

1. Reconnect to the same host that we last connected to (no additional code changes)
2. Store the original parameters (multiple hosts and all) and use them to reconnect (more code changes).

Of course, 1 is already better than what we currently have in pgcli (since we do not handle the multi-host connection at all). But in my personal opinion, 2 is preferable. What do you think?

wronglink:
Hi threre, sorry for long response. I also like the idea of storring the original connection options and reuse it in case of reconnection or altering the connection.

I see these options:
1. Store all initial data as `PGExecute` properties, which is a little bit strange, as we would store 2 copy of all initial parameters (initial and actual) and will lead to some misunderstandings
2. Store data passed to `psycopg2.connect` in a dict and use it for next reconnections. Something like `connection_params`.

In that case there would be 2 kinds of connection properties: initial that are stored as is and used for connection and actual that are taken from the real established connection. After we switch to psycopg2 v2.8, we can turn them into read-only public properties. What do you think?

j-bennet:
I like it. I think this way is best:

> 2. Store data passed to psycopg2.connect in a dict and use it for next reconnections. Something like `connection_params`.

because we don't need to parse and re-parse things more than once.

It may be tricky to do this change and keel the code readable and the intent clear, but it's certainly doable.

wronglink:
I'd like to make a try. What do you think, keep it in current PR or make a separate one?

j-bennet:
Same PR is fine.

wronglink:
@j-bennet I've updated PR and added some logic to support initial connection parameters storing and added ability to update some parts of connection through additional `executor.connect()` calls. What do you think of this?

j-bennet:
Nice! Looks clean, and +1 for the test. Please add some comments when you `.copy()` things to explain why this is done. Otherwise, I think it's ready to go!

wronglink:
> Nice! Looks clean, and +1 for the test. Please add some comments when you `.copy()` things to explain why this is done. Otherwise, I think it's ready to go!

Sorry, I didn't get you well. Do you mean adding some comments where the `copy()` is being called (actually the only place is `pgcli/completion_refresher.py` and it has [comment](https://github.com/dbcli/pgcli/pull/978/files#diff-632d320c84eb6e42c9ef7847b274c828R56)) or adding docstring to `PGExecute.copy()`?

j-bennet:
@wronglink Yes, pgexecute's new `copy` method.

wronglink:
@j-bennet done :-)

j-bennet:
Looks good to me! Please rebase on top of master, and I can merge it.

wronglink:
@j-bennet done

Issue #979, opened 2018-12-12
Lag when typing in v2


jwhitmarsh:
## Description
It's hard to demonstrate, but I have noticable lag when typing out a query in v2.0.1 compared to v1.11.0. It's usually at the end of a query, so i'll get as far as `update table set colum = value where...` and then it will start stuttering. Please let me know what information would be helpful to diagnose the problem.

## Your environment

- [ ] Please provide your OS and version information.
- [ ] Please provide your CLI version.
- [ ] What is the output of ``pip freeze`` command.

MacOS `10.13.6`
pgcli `v2.0.1`

```
➜ pip freeze
awscli==1.16.13
boto3==1.7.70
botocore==1.12.3
certifi==2018.8.24
chardet==3.0.4
cli-helpers==1.1.0
Click==7.0
colorama==0.3.9
configobj==5.0.6
docutils==0.14
httpie==0.9.9
httpie-jwt-auth==0.3.0
humanize==0.5.1
idna==2.7
jmespath==0.9.3
pgcli==1.11.0
pgspecial==1.11.4
prompt-toolkit==1.0.15
psycopg2==2.7.6.1
pyasn1==0.4.4
Pygments==2.2.0
python-dateutil==2.7.3
python-magic==0.4.15
pytz==2018.5
PyYAML==3.13
requests==2.19.1
rsa==3.4.2
s3cmd==2.0.2
s3transfer==0.1.13
setproctitle==1.1.10
six==1.11.0
sqlparse==0.2.4
tabulate==0.8.2
terminaltables==3.1.0
urllib3==1.23
wcwidth==0.1.7
You are using pip version 18.0, however version 18.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
```

j-bennet:
I think it's most likely a prompt-toolkit issue. Migrating to prompt-toolkit 2.0 was the only change in `pgcli` 2.0. @jonathanslenders can you advise?

jonathanslenders:
One change in prompt_toolkit 2.0 is that completions are now synchronous by default. If they are expensive, they have to be wrapped into ``prompt_toolkit.completion.ThreadedCompleter``, otherwise they'll block the event loop and make the UI unresponsive. Could this be the cause?

amjith:
Is there a reason for synchronous completion being the default?

j-bennet:
> If they are expensive, they have to be wrapped

[...]

PR #982, opened 2018-12-27
Remove some functions completions

j-bennet:
## Description

Fix for the lag problem in v2 (https://github.com/dbcli/pgcli/issues/979).

## Checklist

### DIFF
--- a/pgcli/packages/parseutils/meta.py
+++ b/pgcli/packages/parseutils/meta.py
@@ -80,7 +80,7 @@ class FunctionMetadata(object):
         self.is_aggregate = is_aggregate
         self.is_window = is_window
         self.is_set_returning = is_set_returning
-        self.is_extension = is_extension
+        self.is_extension = bool(is_extension)
         self.is_public = (self.schema_name and self.schema_name == 'public')
 
     def __eq__(self, other):
@@ -92,8 +92,8 @@ class FunctionMetadata(object):
 
     def _signature(self):
         return (
-            self.schema_name, self.func_name, self.arg_names, self.arg_types,
-            self.arg_modes, self.return_type, self.is_aggregate,
+            self.schema_name, self.func_name, self.arg_names,
+            self.arg_types, self.arg_modes, self.return_type, self.is_aggregate,
             self.is_window, self.is_set_returning, self.is_extension, self.arg_defaults
         )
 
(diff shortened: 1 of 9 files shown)


j-bennet:
Cleaned up and ready to merge. This does not fix the problem completely, but it's an improvement. There's also https://travis-ci.org/dbcli/pgcli/jobs/472740029 to address the same problem.

pgcli/packages/parseutils/meta.py
```diff
@@ -79,6 +80,8 @@ def __init__(
         self.is_aggregate = is_aggregate
         self.is_window = is_window
         self.is_set_returning = is_set_returning
+        self.is_extension = bool(is_extension)
```
amjith: I'm having a hard time seeing how this is used.

j-bennet: Fixed in https://github.com/dbcli/pgcli/pull/982/commits/ed5dc2e68ac373d69a949f13be48ab20b4d2161f.

pgcli/pgcompleter.py
```diff
@@ -636,22 +637,33 @@ def list_dict(pairs): # Turns [(a, b), (a, c)] into {a: [b, c]}
...
     def get_function_matches(self, suggestion, word_before_cursor, alias=False):
+
         if suggestion.usage == 'from':
             # Only suggest functions allowed in FROM clause
-            def filt(f): return not f.is_aggregate and not f.is_window
+
+            def filt(f):
+                return (not f.is_aggregate and
+                        not f.is_window and
+                        not f.is_extension and
```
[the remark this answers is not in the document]
j-bennet: I'll get you that info in a bit. There might have been a reason I left `is_extension` out of this PR.

j-bennet:
Here are the results that I got when removing either extension functions, or non-public functions, or both, from completions:

https://gist.github.com/j-bennet/f6181e28460916910433882abb8d2b75

In my case (I have postgis installed)

* initially, I get up to 3603 completions
* removing extensions gets rid of 1059 completions
* removing non-public functions gets rid of 2374 completions
* removing both gets rid of 3434 completions.

I did not add a filter for underscored function names. It was not necessary, they went away once I filtered out extensions.

I handle a case when the user explicitly types a non-public schema name here:

https://github.com/dbcli/pgcli/blob/ed5dc2e68ac373d69a949f13be48ab20b4d2161f/pgcli/pgcompleter.py#L648

The one case I'm not handling here is filtering extensions. What if a user explicitly wants an extension function and starts typing it? What is a good way to handle that?

j-bennet:
Going to merge this and deal with edge cases after.

PR #956, opened 2018-10-14
Clean up and add behave logging

meeuw:
## Description
Clean up and add behave logging

## Checklist

- [x] I've added this contribution to the `changelog.rst`.
- [x] I've added my name to the `AUTHORS` file (or it's already there).

PR #996, opened 2019-01-21
Use pexpect.TIMEOUT instead of pexpect.exceptions.TIMEOUT

saper:
All pexpect submodules have been moved into the pexpect package as of version 3.0.

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -4,6 +4,7 @@ Upcoming:
 Bug fixes:
 ----------
 * Avoid error message on the server side if hstore extension is not installed in the current database (#991). (Thanks: `Marcin Cieślak`_)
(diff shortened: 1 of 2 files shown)


j-bennet:
@saper Nice catch! Looks like we're already pinning to 3.3 here:

https://github.com/dbcli/pgcli/blob/ef8aac61c476bd46747d3d4910a4d494a59fa375/requirements-dev.txt#L6

I'm surprised we didn't run into an error until now!

saper:
> @saper Nice catch! Looks like we're already pinning to 3.3 here:

Yes, what I mean that behaviour has been changed in 3.0 already. But this all comes from the #956 patch :(

PR #999, opened 2019-01-21
VISUAL overrides the value of EDITOR for the \e command

saper:
Also revert unfinished logging change from #956

### DIFF
--- a/tests/features/environment.py
+++ b/tests/features/environment.py
@@ -22,6 +22,7 @@ def before_all(context):
     os.environ['COLUMNS'] = "100"
     os.environ['PAGER'] = 'cat'
     os.environ['EDITOR'] = 'ex'
+    os.environ['VISUAL'] = 'ex'
 
     context.package_root = os.path.abspath(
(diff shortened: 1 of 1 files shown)


meeuw:
Uh, I'm fine with adding VISUAL and I see why you're reverting my extended logging, I forgot to port some code from mycli. These lines need to be copied: 

https://github.com/dbcli/mycli/blob/2538d3fe56eb7b274280cb9102abdd201ca74ccc/test/features/steps/wrappers.py#L82

Would you like to do this or shall I create a new PR and close this one?

meeuw:
I don't remember why I used a separate context.logfile, you could try to remove it and see if it still works.

In my opinion adding a logfile isn't testing, I agree the behave/expect tests are hard to debug (that's what I was trying to solve using the logfile).

Please rewrite this PR to only include the VISUAL addition and optionally add the logfile which I've missed while porting this feature from mycli.

meeuw:
:+1: thanks, also for finding the bug with logfile, I'll try to post a PR this evening to fix it.
I'll also add a contribution to the changelog for this.

PR #1007, opened 2019-02-11
keybinding for dismissing autocompletelist

verfriemelt-dot-org:
## Description

i added the esc key in `pgcli/key_bindings.py` to dismiss the autocomplete. The main reason for this, that it sometimes tends to get in the way of the source, especially if you have multiline selects pasted into pgcli.

since im not that profound into python and its libs you could maybe improve the current solution.

thank you for maintaining this greate piece of software!

## Checklist

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -89,6 +89,7 @@ Contributors:
     * Ignacio Campabadal
     * Mikhail Elovskikh (wronglink)
     * Marcin Cieślak (saper)
+    * easteregg
 
 Creator:
 --------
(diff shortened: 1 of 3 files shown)


j-bennet:
Thank you for the PR! I can't help but feel that autocomplete is the best part of pgcli, but people have different needs. :) It seems to work fine. Merging!

🍫

verfriemelt-dot-org:
the autocomplete is greate :) but if you dump a big function into the cli and want to verify it again, before hitting enter, the autocomplete tends to be in the way.

Issue #1014, opened 2019-03-04
cannot quit application without reconnecting to database?


gantsevdenis:
## Description
```
Connection reset. Reconnect (Y/n): \q
Error: \q is not a valid boolean
Connection reset. Reconnect (Y/n): n
wakapi_depl1> \q                                                                                                                                                                                                                             
connection already closed
wakapi_depl1>                                                                                                                                                                                                                                
wakapi_depl1> exit                                                                                                                                                                                                                           
connection already closed
wakapi_depl1> exit()                                                                                                                                                                                                                         
connection already closed
wakapi_depl1> quit                                                                                                                                                                                                                           
connection already closed
```
## Your environment
```
pgcli version: 2.0.2
postgres version: 10.5
OS: Linux 4.10.0-38-generic

```

PR #1009, opened 2019-02-14
Add reconnect automatically enhancement

808sAndBR:
## Description

Adding the enhancement requested in #989 to automatically reconnect to the database when the server closes the connection.

I couldn't figure out how to add testing for this since I couldn't find a way to force a disconnect to trigger the auto reconnection. 

## Checklist

- [x] I've added this contribution to the `changelog.rst`.
- [x] I've added my name to the `AUTHORS` file (or it's already there).

PR #1017, opened 2019-03-05
Fix for https://github.com/dbcli/pgcli/issues/1014.

j-bennet:
## Description

Fixes https://github.com/dbcli/pgcli/issues/1014.

## Checklist

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -13,6 +13,7 @@ Bug fixes:
 * All pexpect submodules have been moved into the pexpect package as of version 3.0. Use pexpect.TIMEOUT (Thanks: `Marcin Cieślak`_)
 * Resizing pgcli terminal kills the connection to postgres in python 2.7 (Thanks: `Amjith Ramanujam`_)
 * Fix crash retrieving server version with ``--single-connection``. (Thanks: `Irina Truong`_)
+* Cannot quit application without reconnecting to database (#1014). (Thanks: `Irina Truong`_)
 
 Internal:
 ---------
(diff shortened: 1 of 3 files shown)


j-bennet:
> What will happen in this case if the user executes a pgspecial command that needs a cursor?

It will throw an exception further down the line, in `pgspecial`. Right now, it just throws an exception earlier.

> I notice that \dp is NO_QUERY, but it might actually need a connection to execute. I guess it's fine since it's not actually implemented? I didn't notice any other pgspecial commands that might be problematic.

Whatever actually needs a connection, will break, but I think it's fine, because it also breaks now.

> Should some tests be added for what happens when the connection is closed but the prompt is still open? Should pgcli just exit at that point?

No need to exit. I recently merged a PR to auto-reconnect on connection dropped:

https://github.com/dbcli/pgcli/pull/1009

Automatic reconnect is a feature that `mycli` already had for a while, and it's much more convenient than asking the user if he wants to reconnect every time connection times out. After #1009 was merged, this one is really only handling the edge case where auto-reconnect did not work.

Good point about testing. I will see if I can add a unit or integration test here. It may be a bit of a hassle, what with the additional complexity of killing the connection.

j-bennet:
@maxrothman I added a unit test, would appreciate another look!

amjith:
@j-bennet The test you added seems to be failing. Can you take a look?

j-bennet:
Will do.

j-bennet:
Missed the pep8 problem. Perhaps it's time to black-ify this repo, just like you did for litecli. All good now!

amjith:
:+1:

maxrothman:
Sorry I didn't have time to follow up with your comments. Thanks for getting this one across the finish line @amjith!

PR #1023, opened 2019-03-16
Remove the expand_tab option from pgclirc.

amjith:
## Description
This is a follow up to #928.

Picking 4 spaces whenever a user presses the tab key. Removes the option from the config file.


j-bennet:
@amjith Were you going to add a changelog message for this? If not, go ahead and merge.

Issue #1034, opened 2019-04-08
Services from .pgservice.conf don't work in the new version


Qu4tro:
## Description

Using services don't work (but they did in the previous release):
`PGSERVICE=spacex pgcli` or `pgcli service=spacex`
By not working I mean: It tries to get a connection with my user account and this is the error message that I get: `FATAL:  database "me" does not exist`

Might be helpful to know the PR that caused the issue: #978 

I also, kind of rely on pgcli (thanks for this gem), so I came up with a quick patch that might be helpful for anyone having the same issue:

```
From 49582ccb7c4ec98aa9736c02d300096952d065fe Mon Sep 17 00:00:00 2001
From: Xavier Francisco <xavier@yarilabs.com>
Date: Mon, 8 Apr 2019 19:46:29 +0100
Subject: [PATCH] Fix usage of pgservice

---
 pgcli/main.py      | 9 +++++----
 pgcli/pgexecute.py | 2 +-
 2 files changed, 6 insertions(+), 5 deletions(-)

diff --git a/pgcli/main.py b/pgcli/main.py
index d7853e47..adf42283 100644
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -398,11 +398,12 @@ class PGCli(object):
                 dsn='', **kwargs):
         # Connect to the database.
 
-        if not user:
-            user = getuser()
+        if not dsn:
+            if not user:
+                user = getuser()
 
-        if not database:
-            database = user
+            if not database:
+                database = user
 
         # If password prompt is not forced but no password is provided, try
         # getting it from environment variable.
diff --git a/pgcli/pgexecute.py b/pgcli/pgexecute.py
index 8c3bfa82..6d94ccc6 100644
--- a/pgcli/pgexecute.py
+++ b/pgcli/pgexecute.py
@@ -233,7 +233,7 @@ class PGExecute(object):
         }
         new_params.update(kwargs)
         conn_params.update({
-            k: unicode2utf8(v) for k, v in new_params.items() if v is not None
+            k: unicode2utf8(v) for k, v in new_params.items() if v
         })
 
         if 'password' in conn_params and 'dsn' in conn_params:
-- 
2.21.0
```

First part removes defaults, if dsn is set. Second part cleans the dict properly (as empty attributes are mostly empty strings and not None objects).

I've opened an issue instead of a PR, since I'm not sure of the consequences these changes would have for other use cases. "Works for me" would be a good title for the patch.

j-bennet:
[...]

PR #1035, opened 2019-04-08
Fix usage of pgservice

Qu4tro:
## Description

Using services don't work (but they did in the previous release):
`PGSERVICE=spacex pgcli` or `pgcli service=spacex`
By not working I mean: It tries to get a connection with my user account and this is the error message that I get: `FATAL:  database "me" does not exist`

Might be helpful to know the PR that caused the issue: #978 

This PR implements a quick fix for this issue.

Changes:
     First part removes defaults, if dsn is set. Hopefully they're not needed since we're using dsn (might be totally in the wrong here).
     Second part cleans the dict properly (as empty attributes are mostly empty strings and not None objects). Seems like an issue. Not sure if this was triggered by my use-case.

Like mentioned in #1034, this PR lacks some love at this moment. Feel free to suggest / push for improvements.

## Checklist

### DIFF
--- a/pgcli/pgexecute.py
+++ b/pgcli/pgexecute.py
@@ -232,14 +232,21 @@ class PGExecute(object):
             'dsn': dsn,
         }
         new_params.update(kwargs)
+
+        if 'dsn' in new_params:
+            new_params = {
+                'dsn': new_params['dsn'],
+                'password': new_params['password']
+            }
+
+            if 'password' in new_params:
+                new_params['dsn'] = "{0} password={1}".format(
+                    new_params['dsn'], new_params.pop('password')
+                )
+
         conn_params.update({
             k: unicode2utf8(v) for k, v in new_params.items() if v
         })
-
-        if 'password' in conn_params and 'dsn' in conn_params:
-            conn_params['dsn'] = "{0} password={1}".format(
-                conn_params['dsn'], conn_params.pop('password')
-            )
 
         conn = psycopg2.connect(**conn_params)
         cursor = conn.cursor()
--- a/pgcli/pgexecute.py
+++ b/pgcli/pgexecute.py
@@ -233,7 +233,7 @@ class PGExecute(object):
         }
         new_params.update(kwargs)
         conn_params.update({
-            k: unicode2utf8(v) for k, v in new_params.items() if v is not None
+            k: unicode2utf8(v) for k, v in new_params.items() if v
         })
 
         if 'password' in conn_params and 'dsn' in conn_params:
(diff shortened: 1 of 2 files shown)


pgcli/pgexecute.py
```diff
@@ -233,7 +233,7 @@ def connect(self, database=None, user=None, password=None, host=None,
         }
         new_params.update(kwargs)
         conn_params.update({
-            k: unicode2utf8(v) for k, v in new_params.items() if v is not None
+            k: unicode2utf8(v) for k, v in new_params.items() if v
```
j-bennet: Would this change alone be enough to fix the problem, or `if not dsn:` change is necessary as well?

Qu4tro: Both changes are necessary. If the dsn check removed, both user and database will be present on conn_params, triggering the issue.

j-bennet: I think this was the spot where #978 went wrong:

https://github.com/dbcli/pgcli/pull/978/files#diff-8859be39408e4fbeea82b923784d9e65L227

Previously, we connected to database in 2 different ways:

* if we had a `dsn`, we updated it with `password` and called `psycopg2.connect` with only one parameter: `dsn=xxx`
* otherwise, we called `psycopg2.connect` with the whole set of parameters (`host`, `user`, etc).

I would suggest that we go back to the same logic. If we have `dsn` provided, our `conn_params` should only contain `dsn` (possibly updated with `password`). Otherwise, include all parameters. This logic can be handled by `pgexecute.py`, without the need for the change in `main.py.

What do you think?

Qu4tro: Sounds good. I'll add a commit with the changes.

Qu4tro: If the tests still pass, I'll add a couple of tests to the services feature.

pgcli/pgexecute.py
```diff
@@ -232,15 +232,22 @@ def connect(self, database=None, user=None, password=None, host=None,
             'dsn': dsn,
         }
         new_params.update(kwargs)
+
+        if 'dsn' in new_params:
```
j-bennet: This is not going to work right. We always have `dsn` key in `new_params`, because we just created `new_params` with that key and other keys above. You want to do this only if `dsn` is non-empty.

pgcli/pgexecute.py
```diff
@@ -232,15 +232,22 @@ def connect(self, database=None, user=None, password=None, host=None,
...
+            new_params = {
+                'dsn': new_params['dsn'],
+                'password': new_params['password']
+            }
+
+            if 'password' in new_params:
```
j-bennet: Same here: you only want to add `password` into dsn string if password is non-empty.

amjith:
I'm going to fix this up and update the PR.

amjith:
Looks like I can't update the existing PR. I'm going to create a new one.

Issue #1041, opened 2019-04-22
PGCLI takes a long time to start


nhumrich:
## Description
Anytime I run pgcli, it takes forever to do anything. It almost seems like it hangs, but if I wait long enough, it eventually works. I initially suspected it was doing something with the keyring because the first time I left it running for a while, the kwallet menu came up. I tried to disable the keyring in the config file and it still takes a long time to load. It even takes a while when connecting to a local database. Everything connects instantly using psql. 

Even just running `pgcli --version` takes forever to load. It somehow looks like its still trying  to do something with the keyring, if I ctrl+c in the middle of hanging I get a stack trace that looks like this:

```python
^CTraceback (most recent call last):
  File "/usr/bin/pgcli", line 11, in <module>
    load_entry_point('pgcli==2.1.0', 'console_scripts', 'pgcli')()
  File "/usr/lib/python3.7/site-packages/pkg_resources/__init__.py", line 489, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/usr/lib/python3.7/site-packages/pkg_resources/__init__.py", line 2843, in load_entry_point
    return ep.load()
  File "/usr/lib/python3.7/site-packages/pkg_resources/__init__.py", line 2434, in load
    return self.resolve()
  File "/usr/lib/python3.7/site-packages/pkg_resources/__init__.py", line 2440, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/usr/lib/python3.7/site-packages/pgcli/main.py", line 49, in <module>
    import keyring
  File "/usr/lib/python3.7/site-packages/keyring/__init__.py", line 1, in <module>
    from .core import (
  File "/usr/lib/python3.7/site-packages/keyring/core.py", line 187, in <module>
    init_backend()
  File "/usr/lib/python3.7/site-packages/keyring/core.py", line 91, in init_backend
    keyrings = filter(limit, backend.get_all_keyring())
  File "/usr/lib/python3.7/site-packages/keyring/util/__init__.py", line 21, in wrapper
    func.always_returns = func(*args, **kwargs)
  File "/usr/lib/python3.7/site-packages/keyring/backend.py", line 206, in get_all_keyring
    return list(rings)
  File "/usr/lib/python3.7/site-packages/keyring/util/__init__.py", line 31, in suppress_exceptions
    for callable in callables:
  File "/usr/lib/python3.7/site-packages/keyring/util/properties.py", line 26, in __get__
[...]

PR #1046, opened 2019-04-28
lazy load keyring.

zzl0:
## Description

Load keyring only when keyring is enabled in the config file.
fix issue: https://github.com/dbcli/pgcli/issues/1041

## Checklist

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -20,7 +20,7 @@ import datetime as dt
 import itertools
 from time import time, sleep
 from codecs import open
-
+keyring = None  # keyring will be loaded later
 
 from cli_helpers.tabular_output import TabularOutputFormatter
 from cli_helpers.tabular_output.preprocessors import (align_decimals,
@@ -45,11 +45,7 @@ from pygments.lexers.sql import PostgresLexer
 
 from pgspecial.main import (PGSpecial, NO_QUERY, PAGER_OFF, PAGER_LONG_OUTPUT)
 import pgspecial as special
-try:
-    import keyring
-except:
-    # pep8 will be unhappy about this. But keyring is optional, and we better disable it if it won't load.
-    keyring = None
+
 from .pgcompleter import PGCompleter
 from .pgtoolbar import create_toolbar_tokens_func
 from .pgstyle import style_factory, style_factory_output
@@ -187,7 +183,7 @@ class PGCli(object):
         self.on_error = c['main']['on_error'].upper()
         self.decimal_format = c['data_formats']['decimal']
         self.float_format = c['data_formats']['float']
-        self.keyring_enabled = c["main"].as_bool("keyring")
+        self.initialize_keyring()
 
         self.pgspecial.pset_pager(self.config['main'].as_bool(
             'enable_pager') and "on" or "off")
@@ -383,6 +379,18 @@ class PGCli(object):
         pgspecial_logger.addHandler(handler)
         pgspecial_logger.setLevel(log_level)
 
+    def initialize_keyring(self):
+        global keyring
+
+        keyring_enabled = self.config["main"].as_bool("keyring")
+        if keyring_enabled:
+            # Try best to load keyring (issue #1041).
+            import importlib
+            try:
+                keyring = importlib.import_module('keyring')
+            except Exception as e:  # ImportError for Python 2, ModuleNotFoundError for Python 3
+                self.logger.warning('import keyring failed: %r.', e)
+
     def connect_dsn(self, dsn):
         self.connect(dsn=dsn)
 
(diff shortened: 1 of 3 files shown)


pgcli/main.py
```diff
@@ -45,11 +45,7 @@
 
 from pgspecial.main import (PGSpecial, NO_QUERY, PAGER_OFF, PAGER_LONG_OUTPUT)
 import pgspecial as special
-try:
-    import keyring
-except:
-    # pep8 will be unhappy about this. But keyring is optional, and we better disable it if it won't load.
-    keyring = None
+
```
j-bennet: IMHO keeping it as a global on top adds to readability (i.e. see at a glance which modules are, or will be, imported by the `main.py`). Even if it's just `keyring = None`.

But if you see some benefit of making it a member variable, then sure.

pgcli/main.py
```diff
@@ -383,6 +379,18 @@ def initialize_logging(self):
...
+        self.keyring = None
+
+        keyring_enabled = self.config["main"].as_bool("keyring")
+        if keyring_enabled:
+            # Try best to load keyring (issue #1041).
+            import importlib
+            try:
+                self.keyring = importlib.import_module('keyring')
+            except Exception:  # ImportError for Python 2, ModuleNotFoundError for Python 3
+                pass
```
j-bennet: A `logger.debug` may be useful here. In case someone wants to use keyring, and it won't load, and we want to see why.

pgcli/main.py
zzl0: I just feel it's a little weird to have a `keyring = None` as a global variable, then update and use it inside a class.

For the readability example, I am not sure, I would not trust those import statements at the beginning of a Python file. Because I have seen many code import modules inside functions.

I don't have a strong preference though. If you still prefer having a global `keyring = None`, I will change it in the next revision.

pgcli/main.py
zzl0: Sure, I will add it in next revision.

pgcli/main.py
j-bennet: > I just feel it's a little weird to have a keyring = None as a global variable, then update and use it inside a class.

This kind of pattern is often used with lazy loading. I would prefer that import on top. Even though it's a weak preference. :)

j-bennet:
@zzl0 P.S. you don't have to develop on a fork anymore. Branches are easier.

pgcli/main.py
```diff
@@ -383,6 +379,18 @@ def initialize_logging(self):
...
 
+    def initialize_keyring(self):
+        self.keyring = None
+
+        keyring_enabled = self.config["main"].as_bool("keyring")
+        if keyring_enabled:
+            # Try best to load keyring (issue #1041).
+            import importlib
+            try:
+                self.keyring = importlib.import_module('keyring')
```
j-bennet: @zzl0 What do you think about wrapping this call into a timeout decorator? Something like described here:

https://stackoverflow.com/questions/2281850/timeout-function-if-it-takes-too-long-to-finish

for those cases when it takes a long time to fail, or just takes too long.

pgcli/main.py
zzl0: Done, thanks.

BTW, I only have seen below pattern before.

```Python
try:
    import xxx
except ImportError:
    xxx = None
```

pgcli/main.py
zzl0: NOTE: I assume the root cause is keyring has a bug/feature.

If a user's keyring doesn't work properly, then the user probably has 3 options:

1. fix keyring, make it work properly.
2. turn off keyring in the config file.
3. don't fix keyring, pgcli wait for some time and then continue without keyring, the effect is same as option 2 and user need to wait for some time.

So I prefer not to do it. What do you think?

zzl0:
@j-bennet 
> P.S. you don't have to develop on a fork anymore. Branches are easier.

Thanks, next time I use a branch.

pgcli/main.py
j-bennet: The pattern I'm talking about usually looks something like this:

```
class MyClass:

    def __init__(self):
        self._expensive_resource = None

    @property
    def my_resource(self):
        if self._expensive_resource is None:
            self._expensive_resource = self.init_expensive_resource()
        return self._expensive_resource
```

and you would use it when you 1) only want to try loading something once, because it's slow or otherwise expensive and 2) only want to load it if it's asked for. Our case of `keyring` here, even if it's not exactly the same, is very similar.

pgcli/main.py
j-bennet: I was thinking of preventing the user from waiting too long in case of 3, and getting frustrated with pgcli. I suppose it would only happen once, after which we could tell the user to disable keyring. It's just that not every user will be able to figure out that keyring is the slow-loading part. It is likely that they will blame pgcli, and perhaps give up on pgcli.

This is still a good fix, and I'm going to merge it for now.

zzl0: Thanks for sharing your thought. Yeah, I agree with that. But I was also worry that users might think pgcli startup is slow (if their keyring always timeout and they don't check the log).

Let's see if users have the problem that you mentioned, we can fix it then.

pgcli/main.py
zzl0: @j-bennet Thanks for sharing. It looks like a [singleton pattern](https://en.wikipedia.org/wiki/Singleton_pattern#Lazy_initialization), it also similar to what I implemented in previous revision (except I put `self.xxx = None` in the a normal method)

j-bennet: Yep.

PR #1049, opened 2019-05-04
black all the things.

j-bennet:
## Description

Run black on everything and add a pre-commit hook.

For the hook to work, install pre-commit:

https://pre-commit.com/#install

And then run: `pre-commit install`

## Checklist


meeuw:
I like the code style of black but won't this pollute blame when we merge this?

Awhile ago I've created this pep8radius like tool which can use any auto-fixing-linter and I could easily add black support. Would that be a good idea?

https://github.com/meeuw/vcscheck

j-bennet:
It will pollute blame, but it will also solve our formatting / code style differences once and for all. With the additional benefit of not needing pep8radius anymore, so no need to tell people "Your PR breaks because pep8radius, you have to run this command to fix it". Even though the command is in our develop guide, people miss it.

I think it's worth it. We're not losing history here, just blame. History can still be used to look up who changed the file and why.

meeuw:
Hmm, that would indeed be very nice! But I don't see why people wouldn't forget to run/install `black` & `pre-commit` (like they do with `pep8radius`).

j-bennet:
Good point. I added that checkbox to our PR template. :)

j-bennet:
So in the end, I added a `black` check into our travis build, instead of `pep8radius` check that we had before. It only runs in 3.6 build. The reasons for that are:

* `black` cannot be run with python lower than 3.6.5 (even though it can lint python 2.7 files)
* failing one of the builds is enough, it does not really matter which one.

j-bennet:
Here... we... go.

:)

zadacka:
2 days later ... I thought that it might amuse you both to know that I was trying to find the origin of the " "psycopg2 >= 2.7.4,<2.8"," change when "<2.8" was introduced, and blame brought me here! 

That said, +1 for using black. Long term good move, for sure. Now I'll just trawl the history a little more to find the commit that introduced the change....

PR #1045, opened 2019-04-28
More intelligent dsn format

j-bennet:
## Description

Fix for https://github.com/dbcli/pgcli/issues/1043.

## Checklist

- [x] I've added this contribution to the `changelog.rst`.
- [x] I've added my name to the `AUTHORS` file (or it's already there).

Issue #1043, opened 2019-04-26
no longer works with password containing spaces


drewthorp:
## Description
Since upgrading to version 2.1.0 I get the following error when logging in with a user who's password contains a space :
`invalid dsn: missing "=" after "<???>" in connection info string`
Where `<???>` is the second word of the password. My login command is as follows :
`pgcli --host=<hostname> --port=5432 --dbname=<dbname> --password --username=<username>`

## My environment
iTerm2 v3.2.9
DB running on a local VirtualBox VM (v6.0) with a guest OS of Ubuntu 14.10
Running pgcli from host of MacOS 10.14.4
CLI is Bash version 3.2.57(1)-release (x86_64-apple-darwin18) 
I have no `pip` command (Installed pgcli with brew)

j-bennet:
Fixed via https://github.com/dbcli/pgcli/pull/1045.

PR #1051, opened 2019-05-06
 Fix usage of pgservice #1035 

amjith:
This is an update to PR #1035 

## Checklist

### DIFF
--- a/pgcli/pgexecute.py
+++ b/pgcli/pgexecute.py
@@ -233,13 +233,13 @@ class PGExecute(object):
         }
         new_params.update(kwargs)
 
-        if 'dsn' in new_params:
+        if new_params['dsn'] is not None:
             new_params = {
                 'dsn': new_params['dsn'],
                 'password': new_params['password']
             }
 
-            if 'password' in new_params:
+            if new_params['password'] is not None:
                 new_params['dsn'] = "{0} password={1}".format(
                     new_params['dsn'], new_params.pop('password')
                 )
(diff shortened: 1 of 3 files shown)


pgcli/pgexecute.py
```diff
@@ -233,14 +233,22 @@ def connect(self, database=None, user=None, password=None, host=None,
             'dsn': dsn,
         }
         new_params.update(kwargs)
+
+        if new_params['dsn']:
```
j-bennet: Looks like I am too late for this, but you want to keep that `make_dsn` call. You do not want to re-format DSN by hand like this PR does. Reason: https://github.com/dbcli/pgcli/issues/1043 (fixed by https://github.com/dbcli/pgcli/pull/1045/files).

amjith: Doh! I'll take a look at it when I land. Sorry about being trigger happy with the merge. 

How come the tests passed? :thinking:

j-bennet:
We do not have a test case for password with space. Probably should add one.

pgcli/pgexecute.py
j-bennet: We do not have a test case for password with space. Probably should add one.

amjith: I just tried to connect to a database with a space in the password it seems to connect just fine. What am I missing?

j-bennet: You'd only get into this branch when connecting with a `dsn` parameter. So for example you have this service defined:

```
$ cat ~/.pg_service.conf                                                                                                                                                                                               [service_postgres]
dbname=postgres
host=localhost
```

and you connect like this:

`pgcli service=service_postgres`

and you have a password with space in it. It will break there.

Issue #1053, opened 2019-05-09
Connecting using socket is broken in current master


j-bennet:
## Description

Connecting using socket is broken in current master.

Exception:

```
2019-05-08 17:15:32,804 (64021/MainThread) pgcli.main ERROR - traceback: 'Traceback (most recent call last):
  File "/Users/irina/src/pgcli/pgcli/main.py", line 477, in connect
    **kwargs)
  File "/Users/irina/src/pgcli/pgcli/pgexecute.py", line 203, in __init__
    self.connect(database, user, password, host, port, dsn, **kwargs)
  File "/Users/irina/src/pgcli/pgcli/pgexecute.py", line 252, in connect
    conn = psycopg2.connect(**conn_params)
  File "/Users/irina/.pyenv/versions/3.6.5/envs/pgcli/lib/python3.6/site-packages/psycopg2/__init__.py", line 130, in connect
    conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
  File "/Users/irina/src/pgcli/pgcli/pgexecute.py", line 40, in _wait_select
    state = conn.poll()
psycopg2.OperationalError: fe_sendauth: no password supplied


During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/irina/src/pgcli/pgcli/main.py", line 484, in connect
    dsn, **kwargs)
  File "/Users/irina/src/pgcli/pgcli/pgexecute.py", line 203, in __init__
    self.connect(database, user, password, host, port, dsn, **kwargs)
  File "/Users/irina/src/pgcli/pgcli/pgexecute.py", line 271, in connect
    self.host = dsn_parameters[\'host\']
KeyError: \'host\'
'
> 2019-05-08 17:15:32,804 (64021/MainThread) pgcli.main ERROR - traceback: 'Traceback (most recent call last):
  File "/Users/irina/src/pgcli/pgcli/main.py", line 477, in connect
    **kwargs)
  File "/Users/irina/src/pgcli/pgcli/pgexecute.py", line 203, in __init__
    self.connect(database, user, password, host, port, dsn, **kwargs)
  File "/Users/irina/src/pgcli/pgcli/pgexecute.py", line 252, in connect
    conn = psycopg2.connect(**conn_params)
  File "/Users/irina/.pyenv/versions/3.6.5/envs/pgcli/lib/python3.6/site-packages/psycopg2/__init__.py", line 130, in connect
[...]

PR #1054, opened 2019-05-09
Dsn parameters not always present.

j-bennet:
## Description

Fixes https://github.com/dbcli/pgcli/issues/1053.

## Checklist

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -10,6 +10,7 @@ Bug fixes:
 * Load keyring only when keyring is enabled in the config file (#1041). (Thanks: `Zhaolong Zhu`_)
 * No longer depend on sqlparse as being less than 0.3.0 with the release of sqlparse 0.3.0. (Thanks: `VVelox`_)
 * Fix the broken support for pgservice . (Thanks: `Xavier Francisco`_)
+* Connecting using socket is broken in current master. (#1053). (Thanks: `Irina Truong`_)
 
 Internal:
 ---------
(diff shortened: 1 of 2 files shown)


meeuw:
Nice and clean fix! Only `self.host = dsn_parameters.get('host')` and `self.port = dsn_parameters['port']` seem to be problematic for this issue, maybe only these should be optional? Would it make sense to also read the socket from dsn_parameters here?

j-bennet:
> Only self.host = dsn_parameters.get('host') and self.port = dsn_parameters['port'] seem to be problematic for this issue, maybe only these should be optional?

True. But there's also no reason to not use "safe" getters for the rest of parameters.

> Would it make sense to also read the socket from dsn_parameters here?

I don't know if that's actually present in parameters. When I log in via a local socket, here are the parameters I see:

```
{'application_name': 'pgcli',
 'dbname': 'postgres',
 'krbsrvname': 'postgres',
 'options': '',
 'passfile': '/Users/irina/.pgpass',
 'port': '5432',
 'sslcompression': '1',
 'sslmode': 'prefer',
 'target_session_attrs': 'any',
 'tty': '',
 'user': 'postgres'}
```

And we already read socket name here:

https://github.com/dbcli/pgcli/blob/580639904c07c4b29376a144d15fa00f4cd39b9e/pgcli/pgexecute.py#L275-L276

(in our `pgexecute`, socket name is used interchangeably with `host`).

meeuw:
Allright, I couldn't test the socket using my docker container and I don't see it in your dump either. Nice work!

PR #1056, opened 2019-05-14
Handle password spaces

amjith:
## Description
Fixed the issue with spaces in password.

Started writing some behavioral tests for some of the command line args. Haven't figured out how to pass in a DSN yet.

### DIFF
--- a/tests/features/steps/basic_commands.py
+++ b/tests/features/steps/basic_commands.py
@@ -45,6 +45,9 @@ def step_run_cli_using_arg(context, arg):
     if arg == '--password':
         arg = '--password'
         prompt_check = False
+    if arg == 'dsn_password':  # This uses the mock_pg_service.conf file in fixtures folder.
+        arg = 'service=mock_postgres --password'
+        prompt_check = False
     wrappers.run_cli(context, run_args=[arg], prompt_check=prompt_check)
 
 @when('we wait for prompt')
--- a/tests/features/steps/basic_commands.py
+++ b/tests/features/steps/basic_commands.py
@@ -46,11 +46,13 @@ def step_run_cli_using_arg(context, arg):
     if arg == '--password':
         arg = '--password'
         prompt_check = False
-    if arg == 'dsn_password':  # This uses the mock_pg_service.conf file in fixtures folder.
+    # This uses the mock_pg_service.conf file in fixtures folder.
+    if arg == 'dsn_password':
         arg = 'service=mock_postgres --password'
         prompt_check = False
         currentdb = "postgres"
-    wrappers.run_cli(context, run_args=[arg], prompt_check=prompt_check, currentdb=currentdb)
+    wrappers.run_cli(context, run_args=[
+                     arg], prompt_check=prompt_check, currentdb=currentdb)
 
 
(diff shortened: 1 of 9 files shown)


tests/features/steps/basic_commands.py
```diff
@@ -32,6 +33,19 @@ def step_see_list_databases(context):
 def step_run_cli(context):
     wrappers.run_cli(context)
 
+@when('we launch dbcli using {arg}')
+def step_run_cli_using_arg(context, arg):
+    prompt_check = False
+    if arg == '--username':
```
j-bennet: Looks like you're not doing anything special for `arg != "password"`. Could combine that into a default case.

amjith: `--password` will prompt for the password which is different behavior than just simply passing in `--username`.

amjith:
> As for DSN case, you could create a mock .pg_service.conf during tests setup stage (delete after), and connect using service=xxx.

Are there any examples of this in current behave tests? I'm trying to figure out how to specify a pg_service.conf file that is not located at `~/.pg_service.conf` location.

j-bennet:
There is no example doing exactly what you want to do (writing a deleting a test config), but take a look at this:

https://github.com/dbcli/pgcli/blob/master/tests/features/environment.py

Here is where all the `before` and `after` setting up is done.

You can specify custom location for service file using a special env var:

`PGSERVICEFILE=foo.txt pgcli service=my_service`

(from https://www.postgresql.org/docs/current/libpq-envars.html)

amjith:
I have added a new behave test for testing the password with spaces case. But unfortunately, the behave scenario that I added is failing for a completely unrelated reason and I'm not able to figure out why that is.

tests/features/steps/basic_commands.py
j-bennet: I meant that while `--password` and `dsn_password` look like your special cases, the rest of the args are not so special. So you could probably do something like:

```
if arg == "--password":
   # do my special thing
elif arg == "dsn_password":
   # another special thing
else:
   arg = "{}={}".format(arg, context.config[arg.replace('--', '')])
```

j-bennet:
I made a few changes to fix your behave test. Our testing setup has a few non-obvious things going on behind the scenes, making it difficult to debug. It could use a bit of refactor. I hope to get around to that at some point.

tests/features/steps/basic_commands.py
```diff
@@ -116,7 +119,8 @@ def step_confirm_destructive_command(context):
         context, 'You\'re about to run a destructive command.\r\nDo you want to proceed? (y/n):', timeout=2)
     context.cli.sendline('y')
 
+
 @then(u'we send password')
 def step_send_password(context):
-    wrappers.expect_exact(context, 'Password for postgres:', timeout=5)
+    wrappers.expect_exact(context, 'Password for', timeout=5)
```
amjith: expect_exact() only does a `startswith()` check?

j-bennet: I think it's looking for a substring in the output. It does not have to match from line start to end.

Issue #974, opened 2018-11-28
Getting pgcli to work on Windows 7


tinmanjk:
## Description

I am not able to install pgcli via pip on Windows.
I have tried to install a binary of psycopg2 (as suggested in issue #108) in advance but to no avail. 
Does anybody have a step-by-step guide on how to do that?
## Your environment

- [ ] Please provide your OS and version information.
Windows 7 64bit
- [ ] Please provide your CLI version.
- [ ] What is the output of ``pip freeze`` command.

j-bennet:
What kind of error are you getting?

j-bennet:
Yes, newer versions of `pgcli` bumped psycopg requirement to `psycopg2 >= 2.7.4,<2.8`:

https://github.com/dbcli/pgcli/blob/3b9041fe36798639b4316fb3e413ac7a04c43da3/setup.py#L19

and I don't see any binaries for that in linked issue (previously the problem was solved with binaries from http://www.stickpeople.com/projects/python/win-psycopg/).

@dbcli/pgcli-core Any of you use Windows / able to help?

meeuw:
I haven't tried it but I'd recommend https://www.anaconda.com/download/#windows

Which includes a compiled version of psycopg2 (https://docs.anaconda.com/anaconda/packages/py3.6_win-64/)

After installing anaconda you should start an "Anaconda Prompt"
https://conda.io/docs/user-guide/getting-started.html

And run the following command:
`pip install pgcli`

Then you should be able te start pgcli from the same anaconda prompt.

tinmanjk:
Thanks for the suggestion. Was hoping for something 'simpler' though.

dbaty:
Commit 96eb37fd1989ec8fd86150f20053abd03f50e7a4 changed the requirement to `psycopg2-binary` on Windows, which should fix the installation. (I don't use Windows and I cannot confirm.) This (old and not very detailed) issue can be closed once a new version of pgcli is released with that change.

dbaty:
The aforementioned fix has been included in version 4.1.0 of pgcli. I am closing this.

Issue #844, opened 2018-02-20
UserWarning: The psycopg2 wheel package will be renamed from release 2.8


ncFoCo:
## Description
The psycopg2 package display's a user warning when calling pgcli. They have a blog post describing why this warning is displayed: http://initd.org/psycopg/articles/2018/02/08/psycopg-274-released/.

### Warning Message
``` Text
UserWarning: The psycopg2 wheel package will be renamed from release 2.8; in order to keep installing from binary please use "pip install psycopg2-binary" instead. For details see: <http://initd.org/psycopg/docs/install.html#binary-install-from-pypi>.
```

## Your environment
- pgcli is installed in a virtual environment in my home .local folder.  
- Symlink pgcli in my .local/bin folder to the virtual instance of pgcli.  

## System Information
- [x] Please provide your OS and version information.  
`Ubuntu 16.04.3 LTS (xenial, x86_64)`  
  
- [x] Please provide your CLI version.  
`pgcli Version: 1.8.2`  
  
- [x] What is the output of ``pip freeze`` command.  
``` Text
cli-helpers==1.0.1
click==6.7
configobj==5.0.6
humanize==0.5.1
pgcli==1.8.2
pgspecial==1.9.0
prompt-toolkit==1.0.15
psycopg2==2.7.4
Pygments==2.2.0
setproctitle==1.1.10
six==1.11.0
sqlparse==0.2.4
tabulate==0.8.2
terminaltables==3.1.0
wcwidth==0.1.7
```

ncFoCo:
Downgrading to psycopg2 2.7.3.2 avoids the warning (`pip install psycopg2==2.7.3.2`). But, it would be preferable to stick with the most recent psycopg2 package.

pshlykov:
Do you guys know if this warning can be suppressed?

Thanks!

ncFoCo:
@pshlykov If you are running pglci from a virtual environment you can install psycopg2 v2.7.3.2 to get around the warning. By default, pgcli will install the [latest version][1] of psycopg2.

[1]:https://github.com/dbcli/pgcli/blob/fa248655da620a70e94935541d50cebf440ea946/setup.py#L19

amjith:
I'll change the dependency to `psycopg2-binary`. 

Thanks for reporting.

pshlykov:
Thanks @ncFoCo @amjith

j-bennet:
Released in 1.9.0.

mbertheau:
[...]

PR #1059, opened 2019-05-27
setup.py: use wheel by default, latest psycopg2

zadacka:
Motivation:
* use the The psycopg2 wheel (pre-compiled binary) for convenience where possible.
* allow use of psycopg2.8 (and greater!) now that it is available. 

The change effectively reverts the setup.py changes in dd579cd and b95ac55, although that the 'detailed instructions' section introduced in dd579cd becomes relevant. In fact the option to install with `--no-binary` is only meaningful when the default requirement is the psycopg2-binary.

I think that this is a reasonable proposal, but would like to know if there were any other reasons behind b95ac55 that would prevent using psycopg2 2.8 being usable.

## Checklist


zadacka:
Having looked a bit further around the conversation surrounding this:
#849
#974
#844

... this PR still seems like a reasonable change. The (recent) release of psycopg2 2.8 should get rid of the warning. Depending on the wheel should be best for most users. Those who specifically need to build from source can still do so using the instructions specified in the 'detailed instructions'.

'course I may still be missing something - please let me know :)

koobs:
From the psycopg docs [1]:

The psycopg2-binary package is meant for beginners to start playing with Python and PostgreSQL without the need to meet the build requirements.

If you are the maintainer of a publish package depending on psycopg2 **you shouldn’t use ‘psycopg2-binary’ as a module dependency.** For production use you are advised to use the source distribution.

The `install_requires` for this dependency should remain `psycopg2`, **not** `psychopg2-binary`

[1] http://initd.org/psycopg/docs/install.html#binary-install-from-pypi

zadacka:
@koobs  - good point

The project seemed split on which to use, with the README making it seem like the binary was the dependency, but then the actual dependency list using the source. And historically it looks like both had been used...

I'm definitely happy to abide by the official docs and change the dependency to use the source.

Given the change to depend on source, and with the README cleanup, does this now look better to you?

koobs:
@zadacka Thanks for the quick turnaround! 

The real distinction is between 'user/application deployments' and 'packaging' (the same distinction as *_requires vs requirements files). [1][2] Unfortunately the distinction has been lost among the noise of cries of "terrible packaging ecosystem" over the years, made worse by everyone doing everything their own unique new way :)

But yes, I saw no test failure delta just unpinning psycopg2 in install_requires, so it should be good to go

From my FreeBSD port commit log message:

```
Upstream has an open PR to unpin psycopg2 [2][3] but it needs to be updated
to use psycopg2 (not psycopg2-binary) in install_requires.

This change patches out the psycopg2 max version. The test results shows
no test failure delta with 2.8.2 over 2.7.7 (the version before ports
r502646).
``` 

[1] https://packaging.python.org/discussions/install-requires-vs-requirements/
[2] https://caremad.io/posts/2013/07/setup-vs-requirement/

zadacka:
@koobs Thanks! And thanks also for the links to reading notes.

Some context: I previously had some problems (now solved) when trying to install psycopg2 from source; the pip error message contained the lines:
```
    If you prefer to avoid building psycopg2 from source, please install the PyPI
    'psycopg2-binary' package instead.
```
I initially read this as 'either option is fine' which seemed to contradict the psycopg docs, but it all makes more sense to me now in the context of `install_requires` / `requirements`. 

(Side note: my problem above was `pg_config` not being in the install user's path, fixing this allowed the installation to complete successfully and pgcli + source-built psycopg2 2.8 seems to work great so far.)

koobs:
@zadacka Yeh, many C/extension based packages have the same issue as they overlap into system/thirdparty library land, and all the lovely challenges that come along with it like a multitude of include/library paths and supporting different compilers/toolchains. But, this is also the beauty of upstreams and downstreams working together, everyone gets to learn about the grass on the other side. 

We're all consumers (downstreams) after all, as your experience too shows :)

telmotrooper:
I was gonna add a merge request changing `psycopg2 >= 2.7.4,<2.8` to `psycopg2 >= 2.7.4,<=2.8.2`, just so pgcli will work on Arch Linux without any user-made changes, but it seems you guys have it covered.

If anyone is reading this while the issue hasn't been fixed, just edit that one line in `/lib/python3.7/site-packages/pgcli-2.1.0-py3.7.egg-info/requires.txt` and it should start working again.

telmotrooper:
Actually reading through all your conversation it seems you haven't decided on merging this. Guess I'll make the merge request with my workaround then.

j-bennet:
After digging through psycopg2 issues, it looks to me like the original segfault problem that caused the split of `psycopg2` and `psycopg2-binary` was fixed:

https://github.com/psycopg/psycopg2/issues/543

However, `psycopg2-binary` is still not recommended as the default choice:

https://github.com/psycopg/psycopg2/issues/921#issuecomment-494378693

The following PR bumps psycopg2 to 2.8.2:

https://github.com/dbcli/pgcli/pull/1060

I think we can merge that PR first, since jury is still out on "can wheels be default again" question.

The README update is good. The `--no-binary` instructions will not work with `psycopg2>=2.8` anymore:

http://initd.org/psycopg/docs/install.html#change-in-binary-packages-between-psycopg-2-7-and-2-8

zadacka:
Closing this out... , this PR does two things:
* set the setup.py dependency "psycopg2 >= 2.7.4", - which is the same as @telmotrooper  and you settled on in #1060, so will now result in no change to trunk!
* update the README, which I think that we're agreed is sensible following the above.

@j-bennet would you like to merge this in order to get the README update? Or would you prefer that I submit a separate PR with just the README change?

j-bennet:
Yikes! Sorry, I got tricked by the PR's title, and missed the fact that you don't actually switch to `psycopg2-binary`. My mistake.

I'm going to review https://github.com/dbcli/pgcli/pull/1061.

PR #1060, opened 2019-05-30
Allow psycopg2 up to 2.8.2

telmotrooper:
This fixes the issue with pgcli not working on Arch Linux, since the package `python2-psycopg2` is already on 2.8.2.

## Checklist

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -95,6 +95,7 @@ Contributors:
     * raylu
     * Zhaolong Zhu
     * Zane C. Bowers-Hadley
+    * Telmo "Trooper" (telmotrooper)
 
 Creator:
 --------
--- a/changelog.rst
+++ b/changelog.rst
@@ -11,7 +11,7 @@ Bug fixes:
 * No longer depend on sqlparse as being less than 0.3.0 with the release of sqlparse 0.3.0. (Thanks: `VVelox`_)
 * Fix the broken support for pgservice . (Thanks: `Xavier Francisco`_)
 * Connecting using socket is broken in current master. (#1053). (Thanks: `Irina Truong`_)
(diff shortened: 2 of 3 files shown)


j-bennet:
Thank you for the PR! This is less involved than https://github.com/dbcli/pgcli/pull/1059. I can merge it as soon as you update the [changelog](https://github.com/dbcli/pgcli/blob/master/changelog.rst) and [AUTHORS file](https://github.com/dbcli/pgcli/blob/master/AUTHORS).

We did fluctuate from `psycopg2` to `psycopg2-binary` and back to `psycopg2` in our `setup.py`. The reason to revert back to the source package was that apparently, a subset of users had problems with the wheels:

http://initd.org/psycopg/articles/2018/02/08/psycopg-274-released/

Because of that article's reasoning, source package looked like a safer default choice. But it can still be overridden with a subsequent wheel install, i.e.:

```
pip install pgcli
pip install psycopg2-binary
```

`pgcli` will run with either one.

telmotrooper:
@j-bennet I've updated both AUTHORS and changelog as requested, I didn't figure such a small contribution deserved a citation in the authors list. I'm glad to be of help though!

j-bennet:
In fact, we can probably stop pinning `psycopg2` as strictly, and simply do this:

`psycopg2 >= 2.7.4`

telmotrooper:
Sounds fair to me, presuming there won't be any breaking API changes soon. I'll update the branch.

j-bennet:
I think the strict pinning was only done because of this change, and uncertainty around how we're going to recommend installing 2.8:

http://initd.org/psycopg/docs/install.html#change-in-binary-packages-between-psycopg-2-7-and-2-8

We did not set an upper pin on `psycopg2` before that.

j-bennet:
Tests are passing. Merging. Thank you! 🍫

Issue #1057, opened 2019-05-15
Supporting mysql-like \G option for temporary extended display


assembler:
Is there a way to support `\G` option at the end of the query so that results of single query can be returned in extended format? Mysql cli has this functionality and i use it all the time. Toggling extended mode with \x before each query is tedious.

Mysql example:

```
data> SELECT 1 as 'one', 2 as 'two' \G
***************************[ 1. row ]***************************
one | 1
two | 2
```

assembler:
Awesome. Thanks for doing this so fast!

PR #1064, opened 2019-06-03
Add \G as a temporary way to use expanded mode.

amjith:
This feature is copied from mycli to enable temporary expanded mode. 

eg: `SELECT * FROM table_name \G` will show results in expanded mode. 

## Checklist

### DIFF
--- a/pgcli/pgexecute.py
+++ b/pgcli/pgexecute.py
@@ -380,7 +380,7 @@ class PGExecute(object):
             try:
                 if pgspecial:
                     # \G is treated specially since we have to set the expanded output.
-                    if sql.endswith('\\G'):
+                    if sql.endswith("\\G"):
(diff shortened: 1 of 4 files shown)


amjith:
#1057 

@assembler FYI

pgcli/pgexecute.py
```diff
@@ -378,6 +379,13 @@ def run(
 
             try:
                 if pgspecial:
+                    # \G is treated specially since we have to set the expanded output.
+                    if sql.endswith("\\G"):
```
fpietka: no big deal, but why don't we use raw string here too to avoid escaping?

amjith: Good question. I just copy pasted the relevant bits from mycli. For whatever reason we didn't use raw strings in mycli.

 :smile:

PR #738, opened 2017-06-26
Fixed installation instructions via brew

indapublic:
You should to tap the formula repo first

## Checklist

- [ ] I've added this contribution to the `changelog.md`.
- [ ] I've added my name to the `AUTHORS` file (or it's already there).

PR #1065, opened 2019-06-03
Add keybinding for alt-enter to introduce a line break

pabloab:
## Description

Use Alt-Enter to insert a multi-line statement without switching to multi-line mode. Already did it by @amjith [on mycli](https://github.com/dbcli/mycli/pull/748/).

Requested in [#738 of mycli](https://github.com/dbcli/mycli/issues/738).

## Checklist

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -97,6 +97,7 @@ Contributors:
     * Zane C. Bowers-Hadley
     * Telmo "Trooper" (telmotrooper)
     * Alexander Zawadzki
+    * Pablo A. Bianchi (pabloab)
 
 Creator:
 --------
--- a/changelog.rst
+++ b/changelog.rst
@@ -889,6 +889,7 @@ Features:
 * Cancel a command using Ctrl+C. (Thanks: https://github.com/macobo)
 * Faster startup by reading all columns and tables in a single query. (Thanks: https://github.com/macobo)
 * Improved psql compliance with env vars and password prompting. (Thanks: `Darik Gamble`_)
+* Pressing Alt-Enter will introduce a line break. This is a way to break up the query into multiple lines without switching to multi-line mode. (Thanks: https://github.com/pabloab).
 
 Bug Fixes:
 ----------
--- a/pgcli/key_bindings.py
+++ b/pgcli/key_bindings.py
@@ -90,10 +90,10 @@ def pgcli_bindings(pgcli):
         event.current_buffer.complete_state = None
         event.app.current_buffer.complete_state = None
 
-    @kb.add('escape', 'enter')
+    @kb.add("escape", "enter")
     def _(event):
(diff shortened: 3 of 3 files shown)


j-bennet:
Lovely! Thanks for the PR!

amjith:
That was fast. 

@pabloab Debian packages are not built by us, so it takes a while to get to the debian repos. 

If you'd like you can install directly from master using this command: 

```pip install -U https://github.com/dbcli/mycli/archive/master.zip```

**Comment by jonathanslenders** [comment]
Keep in mind that this can possibly break Vi mode. (make sure to test it.)

alt-enter is really the same sequence as escape followed by enter (there's no way to distinguish), and in Vi mode, escape is meant to return to navigation mode.

**Comment by amjith** [comment]
I just tried it out. It makes the escape key behave sluggish. I think it's
because it makes the escape key a non-eager event. That's a poor user
experience. Wonder how vim itself handles the scenario. I have alt key
mapped to do different things in vim and they all work fine without
interfering with the escape key.

**Comment by jonathanslenders** [comment]
I think the delay after the escape key press was already there before this PR.

The reason for the delay is that stdin is a stream, so we can't distinguish between an escape key followed very quickly by an enter key press or the alt-enter combination. Similar we also can't distinguish between a left arrow key press, and an escape key press, followed by "[D". So, if we receive the escape key sequence, we have to wait for whatever follows.

Vim handles this with the ``timeoutlen`` and ``ttimeoutlen`` options, and prompt_toolkit implements it in the same way. See: https://python-prompt-toolkit.readthedocs.io/en/master/pages/advanced_topics/key_bindings.html#timeouts I think the default values are reasonable.

**Comment by jonathanslenders** [comment]
By the way, alt+enter was already configured in prompt_toolkit to always accept the input (also in multiline mode), regardless of where the cursor position is.

Many applications (like IPython, and I guess pgcli too) do some kind of validation when enter is pressed to see whether a new line should be inserted, or whether we should accept the input. alt+enter would always accept the input.

Further, I'm thinking about adding a key binding control-j to always insert a line ending. control-j corresponds to \n where control-m corresponds to \r, so I think it could make sense, but I need to do some testing to see whether this doesn't break on any terminal. I know some terminals also send control-j when shift+enter or control+enter is pressed, but this is certainly not standardized.

**Comment by oblitum** [comment]
Folks, default delay to enter normal mode isn't great. Is there a simple setting I can use from `config` or should I patch pgcli?

**Comment by owst** [comment]
@oblitum - the delay should have been removed on master by #1115 

You can install directly from master using:
```
pip install -U https://github.com/dbcli/pgcli/archive/master.zip
```

oblitum:
@owst I'm actually using master, default prompt-toolkit `ttimeoutlen` delay of `0.5` is really bad. I've patched it to `0.05`.

Issue #1110, opened 2019-10-17
PQconninfo not available in libpq < 9.3


Ian-T-Price:
## Description
On running pgcli the above error message is given.
Various people are getting the same error:
https://github.com/psycopg/psycopg2/issues/976
The maintainers suggest that pgcli needs a better work-around

Tried to find & update libpq but I'm not having much success with that - my bad.

## Your environment

- [ Latest AWS Amazon Linux 2 AMI with yum update applied ] Please provide your OS and version information.
- [ Version: 2.1.1 ] Please provide your CLI version.
- [ ] What is the output of ``pip freeze`` command.
amazon-linux-extras==1.6.9
aws-cfn-bootstrap==1.4
awscli==1.16.102
Babel==0.9.6
backports.csv==1.0.7
backports.ssl-match-hostname==3.5.0.1
botocore==1.12.92
cffi==1.6.0
chardet==2.2.1
cli-helpers==1.2.1
Click==7.0
cloud-init==18.2
colorama==0.3.2
configobj==5.0.6
cryptography==1.7.2
docutils==0.12
ec2-hibinit-agent==1.0.0
enum34==1.0.4
futures==3.0.5
hibagent==1.1.0
humanize==0.5.1
idna==2.4
iniparse==0.4
ipaddress==1.0.16
Jinja2==2.7.2
jmespath==0.9.3
jsonpatch==1.2
jsonpointer==1.9
jsonschema==2.5.1
jwcrypto==0.4.2
kitchen==1.1.1
langtable==0.0.31
lockfile==0.9.1
MarkupSafe==0.11
oauthlib==2.0.1
pgcli==2.1.1
pgspecial==1.11.7
Pillow==2.0.0
ply==3.4
prompt-toolkit==2.0.10
psycopg2==2.8.3
pyasn1==0.1.9
pycparser==2.14
pycurl==7.19.0
Pygments==2.4.2
pygpgme==0.3
pyliblzma==0.5.3
pystache==0.5.3
python-daemon==1.6
python-dateutil==2.6.0
pyxattr==0.5.1
PyYAML==3.10
repoze.lru==0.4
requests==2.6.0
rsa==3.4.1
s3transfer==0.1.12
setproctitle==1.1.10
simplejson==3.2.0
six==1.9.0
sqlparse==0.3.0
tabulate==0.8.5
terminaltables==3.1.0
urlgrabber==3.10
urllib3==1.24.3
wcwidth==0.1.7
[...]

PR #1112, opened 2019-10-19
[WIP] Fix for PQconninfo not available.

j-bennet:
## Description

Fix for #1110 PQconninfo not available in libpq < 9.3.

## Checklist


amjith:
Why is this marked as WIP?

j-bennet:
@amjith I have no way to test it, I have newer postgresql and libpq, that's why it's WIP.

amjith:
I was able to reproduce the issue using a docker image.

I have made a small change to the PR to check for libpq version instead of psycopg2 version. 

@j-bennet If you're happy with the changes I've made, please feel free to merge it.

j-bennet:
@amjith I think that we'd still need to check psycopg2 version, not libpq version, to make sure `conn.info.dsn_parameters` is present, because that object is provided by psycopg2. What do you think?

amjith:
You might be right. We will have to check both versions. When I tried your PR yesterday, pgcli won't even start even though psycopg2 version was the latest by libpq version was 9.1. 

I'll update the PR to check for both.

amjith:
Actually we have psycopg2 version requirement in setup.py, so I don't think we need to do the psycopg2 version check during runtime. Thoughts?

j-bennet:
> Actually we have psycopg2 version requirement in setup.py, so I don't think we need to do the psycopg2 version check during runtime.

Ha. Ok, that will do it. Except that we currently require 2.7.4 and up:

https://github.com/dbcli/pgcli/blob/199a0223e044ec3116879e67d2b924050e4afa78/setup.py#L20

so we should probably change it to 2.8 and up. Otherwise we'll have an exception with versions in between 2.7.4 and 2.8 (pypi has 2.7.5, 2.7.6, 2.7.6.1, 2.7.7).

amjith:
:+1:

PR #1109, opened 2019-10-13
Handle a multi-line query on Enter key-press (fixes #1031)

owst:
## Description

Use a (conditional) Enter key-binding to force-handle a multi-line
buffer, rather than doing so by (conditionally) disabling the multiline
mode of prompt_toolkit.

This has the benefit of being more efficient (the multiline Condition
filter is called very often, which (due to the repeated query parsing)
causes editing to become slow with a large buffer that ends in a
semicolon), clearer in intent (we want to force-handle the query, rather
than (temporarily) disable multiline mode which indirectly forces the
buffer to be handled) and avoids a bug in multi-line history search
(issue #1031)

## Checklist

- [x] I've added this contribution to the `changelog.rst`.
- [x] I've added my name to the `AUTHORS` file (or it's already there).

- [x] I installed pre-commit hooks (`pip install pre-commit && pre-commit install`), and ran `black` on my code.
- [ ] Please squash merge this pull request (uncheck if you'd like us to merge as multiple commits)

PR #1114, opened 2019-10-22
Fix the condition for <enter> key.

amjith:
## Description
Fix the non-multi line case where the enter key was not recognized when the completion menu was open.

### DIFF
--- a/pgcli/key_bindings.py
+++ b/pgcli/key_bindings.py
@@ -92,7 +92,7 @@ def pgcli_bindings(pgcli):
         (accept current selection).
 
         """
-        _logger.debug("Detected enter key.")
+        _logger.debug("Detected enter key during completion selection.")
 
         event.current_buffer.complete_state = None
         event.app.current_buffer.complete_state = None
--- a/pgcli/key_bindings.py
+++ b/pgcli/key_bindings.py
@@ -102,9 +102,11 @@ def pgcli_bindings(pgcli):
     # history search, and one of several conditions are True
     @kb.add(
         "enter",
-        filter=~(has_completions | is_searching) & buffer_should_be_handled(pgcli),
+        filter=~(completion_is_selected | is_searching)
+        & buffer_should_be_handled(pgcli),
     )
     def _(event):
+        _logger.debug("Detected enter key.")
         event.current_buffer.validate_and_handle()
 
     @kb.add("escape", "enter")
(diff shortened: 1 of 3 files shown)


pgcli/pgbuffer.py
```diff
@@ -23,10 +26,13 @@ def _is_complete(sql):
 def buffer_should_be_handled(pgcli):
     @Condition
     def cond():
+        # import wdb; wdb.set_trace()
```
owst: Should this have been deleted?

pgcli/key_bindings.py
```diff
@@ -102,9 +102,11 @@ def _(event):
     # history search, and one of several conditions are True
     @kb.add(
         "enter",
-        filter=~(has_completions | is_searching) & buffer_should_be_handled(pgcli),
+        filter=~(completion_is_selected | is_searching)
+        & buffer_should_be_handled(pgcli),
     )
     def _(event):
+        _logger.debug("Detected enter key.")
```
owst: I see the same debug line a few lines above - should we make the output distinct (maybe add "- handling buffer" or similar)?

pgcli/key_bindings.py
```diff
@@ -102,9 +102,11 @@ def _(event):
     # history search, and one of several conditions are True
     @kb.add(
         "enter",
-        filter=~(has_completions | is_searching) & buffer_should_be_handled(pgcli),
+        filter=~(completion_is_selected | is_searching)
```
owst: I recall using `has_completions` rather than `completion_is_selected` in #1109 consciously, but I don't remember _why_. Can you give an example key-press sequence to show where the existing behaviour causes issues? I use multi-line and vim mode, so it's quite possible that I didn't fully test the emacs/single-line mode - apologies for that.

owst: I also tried to add a feature test for keypress handling in #1109, but ran into issues (I was having trouble sending the correct keypress sequences as I recall). 

It'd be nice if we could add some tests in this area, as I think it's important to preserve a consistent UI. Do you know if anyone has tried before - or whether it's even possible to test the key press handling such as this in a feature test?

amjith: Reproducible key sequence:

Ensure multiline mode is OFF either via the config file or by toggling it via F3 key. 

SELECT a<enter>

When you type `SELECT a` it will pop up a completion menu with possible suggestions for the letter `a`. If you press enter it should submit the command to the server but instead a newline is inserted with the completion menu kept open. 

This PR fixes that situation.

PR #1115, opened 2019-10-23
Disable escape-enter keybinding in vi mode

owst:
## Description 

Escape then enter intended as two separate key presses is a very common
sequence when using vi mode - returning to normal mode from insert mode
(escape) and then submitting the query (enter). The presence of the
escape-enter key binding overrides this behaviour by inserting
undesirable newlines (newline is be easily inserted when in insert mode
by pressing enter) or by introducing a noticable delay/lag before enter
can be pressed after escape, due to prompt toolkit not being able to
recognise escape immediately (since it may form part of an escape-enter
sequence).

### DIFF
--- a/pgcli/key_bindings.py
+++ b/pgcli/key_bindings.py
@@ -8,6 +8,7 @@ from prompt_toolkit.filters import (
     is_searching,
     has_completions,
     has_selection,
+    vi_mode,
 )
 
 from .pgbuffer import buffer_should_be_handled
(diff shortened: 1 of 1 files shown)


amjith:
I wonder if this is going to break compatibility for current users who use vi mode and rely on alt-enter to introduce line breaks.

amjith:
I guess we'll find out if this is an issue after the release. Merging!

:coffee:

PR #1116, opened 2019-10-24
Add a test to verify error message.

amjith:
## Description
Add a behave test to verify error condition.


amjith:
This is the test that I came up with for checking the regression introduced in the non-multiline mode. But the problem is I can't make this test fail using the old code. I am at a loss to try and come up with a reliable test to catch the regression.

amjith:
Merging.

PR #1119, opened 2019-10-29
setup: Simplify getting module version

mmtj:
## Description
Get version of module for setup.py in simpler way - just import it.
Unless you are planning to release postfixed versions (like x.y.z-dev) often, there is IMO no need to use regular expressions and ast to determine version number.
In case this is intentional, regular expression alone should be sufficient solution.

## Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,14 +1,7 @@
-import re
-import ast
 import platform
 from setuptools import setup, find_packages
 
-_version_re = re.compile(r"__version__\s+=\s+(.*)")
-
-with open("pgcli/__init__.py", "rb") as f:
-    version = str(
-        ast.literal_eval(_version_re.search(f.read().decode("utf-8")).group(1))
-    )
+from pgcli import __version__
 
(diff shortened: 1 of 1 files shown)


amjith:
I don't think this will work when pgcli is already installed in the virtualenv. I understand that the current code is complicated but it has served us well and I don't see a need to change it. 

I'm happy to reconsider if you have a reasonable counterpoint.

mmtj:
Could you point out case when will it break in virtualenv? I don't believe it will, but I could be mistaken.

I just tried it and `pip install . --upgrade` seems to work fine in virtualenv.
Not sure in case of `python setyp.py install` though, but combination of installing through `python setup.py` with virtualenv always lead to some interesting behaviour (at least for me), so I guess it is reasonable to expect use virtualenv & pip.

Nevertheless I see your point with "don't change what works". Also I am aware that this could break setup.py in few corner cases - if some in-package imports happens in `__init__.py` - however I didn't seen such code in commit history, so I guess this package is safe from that.

Issue #1121, opened 2019-11-19
pgcli + postgresql 12: colum def.adsrc not exists


alxchk:
## Description
Latest pgcli from pip do have issues with latest postgresql:

During startup:

```
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 754, in runs-mode     Refreshing completions...                                                                                                                                                
    self.__target(*self.__args, **self.__kwargs)
  File "/home/manager/.local/lib/python2.7/site-packages/pgcli/completion_refresher.py", line 65, in _bg_refresh
    refresher(completer, executor)
  File "/home/manager/.local/lib/python2.7/site-packages/pgcli/completion_refresher.py", line 109, in refresh_tables
    completer.extend_columns(executor.table_columns(), kind="tables")
  File "/home/manager/.local/lib/python2.7/site-packages/pgcli/pgcompleter.py", line 218, in extend_columns
    for schema, relname, colname, datatype, has_default, default in column_data:
  File "/home/manager/.local/lib/python2.7/site-packages/pgcli/pgexecute.py", line 591, in table_columns
    for row in self._columns(kinds=["r"]):
  File "/home/manager/.local/lib/python2.7/site-packages/pgcli/pgexecute.py", line 586, in _columns
    cur.execute(sql)
UndefinedColumn: column def.adsrc does not exist
LINE 7:                         def.adsrc as default
                                ^
```

\d failed:

```
> \d sample                                   
column c.relhasoids does not exist
LINE 2: ...                 c.relhasrules, c.relhastriggers, c.relhasoi...
                                                             ^
```

## Your environment
+ OS: Doesn't matter (Gentoo testing, Debian sid, etc)
+ Version: latest from pip 
```
pgcli --version
Version: 2.1.1
```
- Pg versions:
```
$ pip freeze | grep -i pg
pgcli==2.1.1
pgspecial==1.11.7
psycopg2==2.8.3
psycopg2-binary==2.8.4
```

linux-mate:
I have the same problem. (Arch linux and version exactly the same except my psycopg2==2.8.4)

abawchen:
same here

surprisetalk:
Related to #1058

sashaafm:
Same. Is this expected to be fixed soon? Postgres 12 has been out for some time now. Thanks!

PR #1126, opened 2019-12-03
J bennet/release 2.2.0

j-bennet:
## Description

Prepare for a new release. Most importantly to fix https://github.com/dbcli/pgcli/issues/1121.


lelit:
OOC, why the restriction? I'm using PT 3 since a while, and didn't notice any defect...

j-bennet:
@lelit nobody did extensive testing of pgcli with 3.0, and it has incompatible changes according to @jonathanslenders, so I wanted to be on the safe side. Did you run unit tests / behave tests with 3.0? Does everything pass?

lelit:
No, sorry: I browsed its changelog and didn't spot alerting changes... so I upgraded it.
Didn't have surprises, using it. But if Jonathan said so, the restriction is surely due.

setup.py
```diff
@@ -13,10 +13,10 @@
 description = "CLI for Postgres Database. With auto-completion and syntax highlighting."
 
 install_requirements = [
-    "pgspecial>=1.11.5",
+    "pgspecial>=1.11.8",
     "click >= 4.1",
     "Pygments >= 2.0",  # Pygments has to be Capitalcased. WTF?
-    "prompt_toolkit>=2.0.6,<2.1.0",
+    "prompt_toolkit>=2.0.6,<3.0.0",
```
jonathanslenders: Notice that this won't upgrade to version 3. it's ``<``, not ``<=``. Better use ``<3.1.0`` or ``4.0.0``.

lelit: But that is intentional, given what Irina said, isn't it?

> it has incompatible changes according to @jonathanslenders, so I wanted to be on the safe side

lelit: FWIW, I tried to execute the test suite using latest PT 3.0.2, and didn't notice any difference in the result, but I couldn't spend enough time to investigate on the failures (sorry about that)...
Given that nothing broke badly, neither in the tests, nor in my daily usage, hopefully the _incompatible changes_ are not impacting pgcli usage.

lelit: To be clear, the same tests failed before and after upgrading PT.

j-bennet: Yikes. We better make sure to fix whatever is broken then.

Issue #1133, opened 2019-12-11
Starting issue


Sharofiddin:
## Description
I gave this type of error when I start working, then continue

> traceback (most recent call last):
  File "/usr/lib/python3.6/threading.py", line 916, in _bootstrap_inner                              
    self.run()mpletion: ON  [F3] Multiline: OFF  [F4] Emacs-mode     Refreshing completions...       
  File "/usr/lib/python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "/home/sharof/.local/lib/python3.6/site-packages/pgcli/completion_refresher.py", line 65, in _bg_refresh
    refresher(completer, executor)
  File "/home/sharof/.local/lib/python3.6/site-packages/pgcli/completion_refresher.py", line 110, in refresh_tables
    completer.extend_foreignkeys(executor.foreignkeys())
  File "/home/sharof/.local/lib/python3.6/site-packages/pgcli/pgcompleter.py", line 278, in extend_foreignkeys
    childcolmeta = 

> meta[childschema][childtable][childcol]

## My environment
-Linux Ubuntu 180.04.
-Version: 2.1.1

laixintao:
Hi @Sharofiddin Can you try to upgrade to v2.2.0 see if it still happens?

dbaty:
No news for more than 4 years, issue happened on an old version. I am closing this.

Sharofiddin:
Issue has been solved in new releases thanks

Issue #675, opened 2017-03-21
KeyError in thread completion_refresh


tejasmanohar:
## Description
I get this error occasionally here and there. I ^C and things work again.
```
Exception in thread completion_refresh:
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/usr/local/lib/python2.7/threading.py", line 754, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/usr/local/lib/python2.7/site-packages/pgcli/completion_refresher.py", line 65, in _bg_refresh
    refresher(completer, executor)
  File "/usr/local/lib/python2.7/site-packages/pgcli/completion_refresher.py", line 105, in refresh_tables
    completer.extend_columns(executor.table_columns(), kind='tables')
  File "/usr/local/lib/python2.7/site-packages/pgcli/pgcompleter.py", line 158, in extend_columns
    metadata[schema][relname][colname] = column
KeyError: u'_stmp_3fc53b90_a'
```

## Your environment

- [x] Please provide your OS and version information.
I see it on ubuntu test server and os x machine when connecting to different databases so I think this is irrelevant
-  pgcli -v
Version: 1.3.1
- `pip freeze`
https://gist.github.com/tejasmanohar/ba53787fa736c95e6e867833e08db804

j-bennet:
Hi @tejasmanohar , thanks for the problem report. Is there anything specific to your workflow when this exception happens
 (making a lot of schema changes for example)? Are you working with the local or remote server?

tejasmanohar:
it's often the first query-- not making a lot of schema changes. remote server

kennethkalmer:
I just encountered a very similar issue after adding foreign keys to a partitioned table. Without the foreign keys partitioned tables work like a charm in pgcli. Turns out it can be reproduced quite easily:

```
$ createdb dev
$ pgcli dev
Server: PostgreSQL 11.3
Version: 2.1.1
Chat: https://gitter.im/dbcli/pgcli
Mail: https://groups.google.com/forum/#!forum/pgcli
Home: http://pgcli.com
kenneth@/tmp:dev> create table cities (id serial primary key not null, "name" text);
CREATE TABLE
Time: 0.009s
kenneth@/tmp:dev> CREATE TABLE measurement (
     city_id         bigint references cities (id),
     logdate         date not null,
     peaktemp        int,
     unitsales       int
[...]

PR #1143, opened 2020-01-31
 Fix exception in thread completion_refresh caused by KeyError in extend_foreignkeys call

troyanov:
## Description
When starting pgcli you could see such an Exception
```
Exception in thread completion_refresh:
  File "/usr/local/Cellar/python@3.8/3.8.1/Frameworks/Python.framework/Versions/3.8/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
postgres@localhost:postgres>   File "/usr/local/Cellar/python@3.8/3.8.1/Frameworks/Python.framework/Versions/3.8/lib/python3.8/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/local/Cellar/pgcli/2.2.0_1/libexec/lib/python3.8/site-packages/pgcli/completion_refresher.py", line 65, in _bg_refresh
                                 refresher(completer, executor)
  File "/usr/local/Cellar/pgcli/2.2.0_1/libexec/lib/python3.8/site-packages/pgcli/completion_refresher.py", line 110, in refresh_tables
    completer.extend_foreignkeys(executor.foreignkeys())
  File "/usr/local/Cellar/pgcli/2.2.0_1/libexec/lib/python3.8/site-packages/pgcli/pgcompleter.py", line 278, in extend_foreignkeys
    childcolmeta = meta[childschema][childtable][childcol]s-mode     Refreshing completions...
KeyError: {table_name_here}
```

Error occurs when your schema contains partitioned table with foreign key.
More information can be found here:
https://github.com/dbcli/pgcli/issues/675#issuecomment-503631853

I've added a simple if-statement to check if key exists just not to throw an exeption. 

I think at least those two issues are connected to that error:
https://github.com/dbcli/pgcli/issues/1133
https://github.com/dbcli/pgcli/issues/675

## Checklist

### DIFF
--- a/pgcli/pgcompleter.py
+++ b/pgcli/pgcompleter.py
@@ -275,6 +275,8 @@ class PGCompleter(Completer):
             parentschema, childschema = e([fk.parentschema, fk.childschema])
             parenttable, childtable = e([fk.parenttable, fk.childtable])
             childcol, parcol = e([fk.childcolumn, fk.parentcolumn])
+            if childtable not in meta[childschema]:
+                continue
             childcolmeta = meta[childschema][childtable][childcol]
             parcolmeta = meta[parentschema][parenttable][parcol]
             fk = ForeignKey(
--- a/pgcli/pgcompleter.py
+++ b/pgcli/pgcompleter.py
@@ -275,8 +275,6 @@ class PGCompleter(Completer):
             parentschema, childschema = e([fk.parentschema, fk.childschema])
             parenttable, childtable = e([fk.parenttable, fk.childtable])
             childcol, parcol = e([fk.childcolumn, fk.parentcolumn])
-            if childtable not in meta[childschema]:
-                continue
             childcolmeta = meta[childschema][childtable][childcol]
             parcolmeta = meta[parentschema][parenttable][parcol]
             fk = ForeignKey(
--- a/AUTHORS
+++ b/AUTHORS
@@ -104,6 +104,7 @@ Contributors:
     * Jonas Jelten
     * BrownShibaDog
     * George Thomas(thegeorgeous)
+    * Anton Troyanov
 
 Creator:
 --------
--- a/changelog.rst
+++ b/changelog.rst
@@ -10,6 +10,7 @@ Features:
 Bug fixes:
 
 * Fix warning raised for using `is not` to compare string literal
+* Fix exception in thread completion_refresh caused by KeyError in extend_foreignkeys call
 
 Internal:
 ---------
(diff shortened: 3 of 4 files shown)


pgcli/pgcompleter.py
```diff
@@ -275,6 +275,8 @@ def extend_foreignkeys(self, fk_data):
             parentschema, childschema = e([fk.parentschema, fk.childschema])
             parenttable, childtable = e([fk.parenttable, fk.childtable])
             childcol, parcol = e([fk.childcolumn, fk.parentcolumn])
+            if childtable not in meta[childschema]:
```
j-bennet: Hmm. This will fix the exception, but I wonder why the key would not be present? Is it possible there's an issue when building a dict of `meta[childschema]`? If so, we better fix the root cause.

troyanov: The root issue comes from this place (only when you have partitioned tables)
https://github.com/dbcli/pgcli/blob/master/pgcli/pgexecute.py#L521-L524


For partitioned tables `relkind` equals `p`, so for partitioned tables key is not there because we do not put it there.

Starting from 9.6+ `relkind` has new members (more info here: https://www.postgresql.org/docs/12/catalog-pg-class.html) 

relkind | char |   | r = ordinary table, i = index, S = sequence, t = TOAST table, v = view, m = materialized view, c = composite type, f = foreign table, p = partitioned table, I = partitioned index
-- | -- | -- | --

I've also tried to include new relkind to query metadata and it worked. But I'm not sure if we should treat partitioned table as a table or not.

j-bennet: I think we should include new table types.

troyanov: I'm not sure about (t) and (f) types but I've added changes to introduce (p) type. I've been running pgcli with these modifications for a week and it looks good so far.

j-bennet: @troyanov Can you take a look at https://github.com/dbcli/pgcli/pull/1145? The PR author thinks we should probably add `f` but not `t`, and it sounds sensible to me as well.

troyanov: @j-bennet I've never worked with `f` and `t` table types, not sure if I have correct knowledge. But According to the docs I've read PR author is correct about `t` type that it should not be added.

j-bennet:
I merged https://github.com/dbcli/pgcli/pull/1145 that fixes the same issues, but includes both `f` and `p`. Sorry - but in any case, thanks a lot for the fix!

Issue #1148, opened 2020-03-02
Add support for prompt_toolkit>3


blazewicz:
Currently `pgcli` is pinned to `prompt_toolkit` version 2:

https://github.com/dbcli/pgcli/blob/91263c37b94a1a3b45be13d35ff1ea19c7b8e1b7/setup.py#L12

This causes a compatibility issue with other tools using `prompt_toolkit`, for example with `litecli` which requires `prompt_toolkit>=3.0.3`.

https://github.com/dbcli/litecli/blob/39d7dad9d63f445e288f476f9e4b76cdd7dc8e9b/setup.py#L30

This issue blocks package managers like `macports` to keep up-to-date versions of both `pgcli`, `litecli` and some other tools.

clbarnes:
I would consider installing pgcli using [`pipx`](https://github.com/pipxproject/pipx/), so that it doesn't have to share a virtual environment with anything, and is available everywhere for your user.

laixintao:
> I would consider installing pgcli, along with any other system tools which happen to be written in python, using [`pipx`](https://github.com/pipxproject/pipx/), so that it doesn't have to share a virtual environment with anything, and is available everywhere for your user.

Thanks and yes, that's the recommended way, but the problem is for some distribution developers, their packaging work requires prompt-toolkit>=3 :)

PR #1149, opened 2020-03-03
upgrade prompt-toolit to 3.0

laixintao:
fix: https://github.com/dbcli/pgcli/issues/1148

## Description

Upgrade dependency.

## Checklist


laixintao:
## Breaking Changes Check List

https://github.com/prompt-toolkit/python-prompt-toolkit/blob/master/CHANGELOG#L61
- Python 2 support has been dropped. Minimal Python version is now 3.6,
  although 3.7 is preferred (because of ContextVars).
- Native asyncio, so some async code becomes slightly different.
- The active `Application` became a contextvar. Which means that it should be
  propagated correctly to the code that requires it. However, random other
  threads or coroutines won't be able to know what the current application is.
- The dialog shortcuts API changed. All dialog functions now return an
  `Application`. You still have to call either `run()` or `run_async` on the
  `Application` object.
- The way inputhooks work is changed.
- `patch_stdout` now requires an `Application` as input.

laixintao:
I checked the source code, I don't think those changes affect us.

> ## Breaking Changes Check List
> 
> https://github.com/prompt-toolkit/python-prompt-toolkit/blob/master/CHANGELOG#L61
> 
>     * Python 2 support has been dropped. Minimal Python version is now 3.6,
>       although 3.7 is preferred (because of ContextVars).
> 
>     * Native asyncio, so some async code becomes slightly different.
> 
>     * The active `Application` became a contextvar. Which means that it should be
>       propagated correctly to the code that requires it. However, random other
>       threads or coroutines won't be able to know what the current application is.
> 
>     * The dialog shortcuts API changed. All dialog functions now return an
>       `Application`. You still have to call either `run()` or `run_async` on the
>       `Application` object.
> 
>     * The way inputhooks work is changed.
> 
>     * `patch_stdout` now requires an `Application` as input.

But if we upgrade we have to:

- [ ] Drop Python2.7
- [ ] Drop Python3.5(prompt-toolkit required 3.6 at least)
- [ ] Fix some test case (I can do that :) )

What do you think? @amjith @j-bennet Is that necessary?

j-bennet:
@laixintao to address your questions:

1. python 2.7 reached end-of-life already. Yes, we better terminate the support for it, and do it in a separate PR.
2. python 3.6 reaches EOL in September 2020, but I don't see why we can't terminate support now. Most users would be on 3.6 or 3.7 by now.
3. Certainly have to fix those test cases!

laixintao:
The tests are all passed on my mac but failed in travis:
```
Feature: Special commands # specials.feature:1

  Scenario: run refresh command             # specials.feature:3
    When we refresh completions             # steps/specials.py:11 0.109s
    And we wait for prompt                  # steps/basic_commands.py:57 0.001s
    Then we see completions refresh started # steps/specials.py:19 1.079s

Closed connection: host=localhost user=postgres port=5432 dbname=pgcli_behave_tests_3_6_10.
Created connection: host=localhost user=postgres port=5432 dbname=postgres.
Closed connection: host=localhost user=postgres port=5432 dbname=postgres.
8 features passed, 0 failed, 0 skipped
23 scenarios passed, 0 failed, 0 skipped
109 steps passed, 0 failed, 0 skipped, 0 undefined
Took 0m48.367s
```
That's odd, I am still investigating it...

laixintao:
OK, so I created a vagrant VM to test it, it fails exactly like travis CI. I think it only happens on Linux.
```
[vagrant@lailab pgcli]$ behave tests/features/crud_table.feature
package root: /vagrant/pgcli
fixture dir: /vagrant/pgcli/tests/features/fixture_data
Created connection: host=localhost user=postgres port=5432 dbname=postgres.
Created connection: host=localhost user=postgres port=5432 dbname=pgcli_behave_tests_3_7_6.
reading fixture data: /vagrant/pgcli/tests/features/fixture_data/
--- os.environ changed values: ---
COLUMNS="100"
COVERAGE_PROCESS_START="/vagrant/pgcli/.coveragerc"
EDITOR="ex"
LINES="100"
PAGER="/vagrant/fedora_pgcli/bin/python3 /vagrant/pgcli/tests/features/wrappager.py ---boundary---"
PGDATABASE="pgcli_behave_tests_3_7_6"
PGHOST="localhost"
PGPORT="5432"
PGSERVICEFILE="/vagrant/pgcli/tests/features/fixture_data/mock_pg_service.conf"
PGUSER="postgres"
VISUAL="ex"
XDG_CONFIG_HOME="/tmp/pgcli_home_st3gu4uk"
--------------------
Feature: manipulate tables: # tests/features/crud_table.feature:1
  create, insert, update, select, delete from, drop
  Scenario: create, insert, select from, update, drop table  # tests/features/crud_table.feature:4
    When we connect to test database                         # tests/features/steps/crud_database.py:30 0.102s
    Then we see database connected                           # tests/features/steps/crud_database.py:88 1.265s
    When we create table                                     # tests/features/steps/crud_table.py:12 0.102s
    Then we see table created                                # tests/features/steps/crud_table.py:60 1.226s
    When we insert into table                                # tests/features/steps/crud_table.py:20 0.101s
    Then we see record inserted                              # tests/features/steps/crud_table.py:68 1.251s
    When we update table                                     # tests/features/steps/crud_table.py:28 0.102s
    Then we see record updated                               # tests/features/steps/crud_table.py:76 1.120s
    When we select from table                                # tests/features/steps/crud_table.py:36 0.102s
    Then we see data selected                                # tests/features/steps/crud_table.py:84 1.003s
    When we delete from table                                # tests/features/steps/crud_table.py:44 0.101s
    Then we confirm the destructive warning                  # tests/features/steps/basic_commands.py:134 1.261s
    Then we see record deleted                               # tests/features/steps/crud_table.py:110 0.060s
    When we drop table                                       # tests/features/steps/crud_table.py:52 0.101s
    Then we confirm the destructive warning                  # tests/features/steps/basic_commands.py:134 1.111s
    Then we see table dropped                                # tests/features/steps/crud_table.py:118 0.068s
    When we connect to dbserver                              # tests/features/steps/crud_database.py:39 0.102s
    Then we see database connected                           # tests/features/steps/crud_database.py:88 2.005s
        File "/vagrant/fedora_pgcli/lib64/python3.7/site-packages/behave/model.py", line 1329, in run
          match.run(runner.context)
        File "/vagrant/fedora_pgcli/lib64/python3.7/site-packages/behave/matchers.py", line 98, in run
          self.func(context, *args, **kwargs)
        File "tests/features/steps/crud_database.py", line 93, in step_see_db_connected
          wrappers.expect_exact(context, "You are now connected to database", timeout=2)
        File "/vagrant/pgcli/tests/features/steps/wrappers.py", line 37, in expect_exact
          ).format(expected, actual, context.logfile.getvalue())
      Exception: Expected:
      ---
      'You are now connected to database'
      ---
      Actual:
      ---
      'Time: 0.016s\r\n_behave_tests_3_7_6>  \r\r\n\r\r\n\r\r\n                                                                               \r \r\\connect postgres \x08_behave_tests_3_7_6> \\connect postgres                                    \r \r_behave_tests_3_7_6> \\connect postgres                                    \r \r\r\r\n\r\nUnhandled exception in event loop:\r\n\r\nException None\r\n ENTER to continue...                                                     \r \r'
      ---
      Full log:
 ```

```
[vagrant@lailab pgcli]$ uname -a
Linux lailab 5.3.7-301.fc31.x86_64 #1 SMP Mon Oct 21 19:18:58 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
```

**Comment by laixintao** [comment]
It might be time issue, I test it manually and found that the `\connection postgres` actually takes longer than 1s.

```
pgcli_behave_tests_3_7_6> \connect postgres
You are now connected to database "postgres" as user "postgres"
Time: 1.784s (a second), executed in: 1.771s (a second)
```

pasenor:
> Most users would be on 3.6 or 3.7 by now

[pypi stats page](https://pypistats.org/packages/pgcli) disagrees. I don't know how accurate it is, but it shows comparable daily download numbers for 3.5 and 3.6

laixintao:
I see, maybe there are some Linux distributions that have Python3.5 installed by default. So that leaves us two options:

1. Still develop based on 3.5, I will revert #1154 
2. Python3.5 Users can only use the old version of pgcli

pasenor:
I find keeping up with prompt-toolkit a good idea, and it does mean 3.6+. Maybe we could consider other distribution options, like PyInstaller? Sure, python coders will prefer pip, but they are likely to have newer python anyway. Everyone else just needs an executable.

laixintao:
> I find keeping up with prompt-toolkit a good idea, and it does mean 3.6+. Maybe we could consider other distribution options, like PyInstaller? Sure, python coders will prefer pip, but they are likely to have newer python anyway. Everyone else just needs an executable.

Yes, that's also a good option. I did that in https://github.com/laixintao/iredis (build script: https://github.com/laixintao/iredis/blob/master/.github/workflows/release.yaml#L118 ), I use Github Action for building a binary for every release. You can download from the release page: https://github.com/laixintao/iredis/releases

I use PyOxidizer for packaging, however. PyInstaller basic package the python code into a zip, so it may be slower to start the cli application. PyOxidizer is faster, but the problems we have to have a dir like `lib/` in the file system duo the import system. There is a comparison from PyOxidizer here: https://pyoxidizer.readthedocs.io/en/stable/comparisons.html

PS: Sorry for massive push notification. I am t rying to fix travis-ci, but some behave tests are randomly timeout after upgrading prompt-prompt.

laixintao:
Finally, all of the test cases passed. It seems the latest version of prompt-toolkit is slower so I have to extend the timeout for some tests.

Also, I met the `CPR unsupported` problems, disabling CPR for some tests but `run the cli` can get the tests passed, but I have to idea why.

j-bennet:
@laixintao Is this PR still a WIP, or ready to go?

laixintao:
Hi @j-bennet I think it's ready to go.

To make the tests pass in prompt-toolkit 3.0, I have to:

1. Extend the pexpect's timeout;
2. Disable CPR for some tests;

Not sure if it's ok.

j-bennet:
@jonathanslenders Interesting. From the comment here, it looks like when using `pexpect`, CPR should return not supported already:

https://github.com/prompt-toolkit/python-prompt-toolkit/blob/0b9a64f20717e069a44358f651196584f43e666d/prompt_toolkit/input/vt100.py#L89

Why do we need to set an environment variable?

laixintao:
> @jonathanslenders Interesting. From the comment here, it looks like when using `pexpect`, CPR should return not supported already:
> 
> https://github.com/prompt-toolkit/python-prompt-toolkit/blob/0b9a64f20717e069a44358f651196584f43e666d/prompt_toolkit/input/vt100.py#L89
> 
> Why do we need to set an environment variable?

Hi @j-bennet , I think what the comment wants to say is that when test in pexpect, CPR is not supported, but this function (As far as I can see) didn't do that for us, thus we need to set `PROMPT_TOOLKIT_NO_CPR` to `1`. (see this travis failed test without disabling CPR: https://github.com/dbcli/pgcli/pull/1149/commits/fe4d5b565e01be6237b6df646429d26eaa0c6d92 )

And why some tests need to enable `CPR` to pass? ( https://github.com/dbcli/pgcli/pull/1149/commits/890ae2bfda2990da6649491331270c3242b7c835#diff-903967dacf3fff9120918217578816f9R162 ) I checked those tests, it turns out that the prompt was `expect` twice:

1. https://github.com/dbcli/pgcli/blob/93c14f844ee33e4d47ae07688f0e5ca7219df7de/tests/features/steps/basic_commands.py#L94
2. https://github.com/dbcli/pgcli/blob/93c14f844ee33e4d47ae07688f0e5ca7219df7de/tests/features/basic_commands.feature#L16

So delete one solved this.

I leave the commits un-rebased so you can see the failed ci logs. :)

jonathanslenders:
I'm not sure but think pexpect allocates a pseudo terminal, which means that ``isatty`` return ``True``, so prompt_toolkit would expect CPR to work, unless there's that environment variable.

j-bennet:
@jonathanslenders Thank you for the explanation, that makes sense. Is prompt_toolkit 3.x really that much slower? I don't feel good about the fact that @laixintao had to increase the timeouts from 5 to 15.

laixintao:
Hi @j-bennet. Sorry looks like my fault. The `timeout` only happens [when didn't setting `PROMPT_TOOLKIT_NO_CPR`](https://github.com/dbcli/pgcli/pull/1149/commits/d379b1e736dcc6623ba53fc3d4e6f095fd3e8344). We needn't increase the timeout if we disable CPR in tests. I reverted it back.

Sorry for bothering ;-(

j-bennet:
@laixintao That's good news! Thank you. Let's get this PR out of the door.

Issue #114, opened 2015-01-13
Add automatic testing on windows platform using AppVeyor


amjith:
Check if it's possible to use http://www.appveyor.com/ for testing on Windows.

amjith:
It is high time we do this. I don't want new features to silently break in Windows. 

@dbcli/pgcli-core Any takers? If I don't hear back by Friday, I'll start investigating how to add this.

j-bennet:
I'm interested. Let me look into this.

j-bennet:
I played with this for wharfee and found that pexpect is not available on Windows. In fact, it's available there starting with version 4, but we're pinning it to 3.3 because 4.0 can't run under python 2.6 (see #375).

I think I'll add the config to run unit tests without functional (behave) tests to start with, and then look for ways to replace (or upgrade) pexpect and run functional tests.

darikg:
Would it be possible to do linux tests with pexpect=3.3 and windows tests with pexpect=4? The number of windows python 2.6 users has to be pretty small and it might not be worth worrying about.

j-bennet:
@darikg So only test under 2.7, 3.3 and 3.4 on Windows? That may be an option.

darikg:
Yeah, exactly

amjith:
I like this suggestion. :)

j-bennet:
https://ci.appveyor.com/project/j-bennet/pgcli/build/job/qdqvwl9hpexmrcbp

Hmm... it fails trying to install psycopg2 on Win? What the...

amjith:
Interestingly, that's what I expected. It's not easy to install psycopg on
windows using pip.

I had to download an exe file and use easy install to install it.

darikg:
@j-bennet I played around with your appveyor.yml and got this to install psycopg2 and run the tests. You can see it downloads psycopg2 pre-compiled binaries [from here](http://www.stickpeople.com/projects/python/win-psycopg/). The problem is I had to hardcode the path to a specific python/arch, so I'm not sure go to get that to work with the environmental matrix. 

```
build: false

services:
  - postgresql

environment:
  matrix:
    - PYTHON: "C:\\Python34"
      PYTHON_VERSION: "3.4.1"
      PYTHON_ARCH: "32"

init:
  - "PATH=C:\\Program Files\\PostgreSQL\\9.3\\bin\\;%PYTHON%;%PYTHON%\\Scripts;%PATH%"
  - "SET PGUSER=postgres"
[...]

PR #1154, opened 2020-03-10
Drop Python3.5.

laixintao:
## Description

So we can work on #114 

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.black]
 line-length = 88
-target-version = ['py37']
+target-version = ['py36']
 include = '\.pyi?$'
 exclude = '''
 /(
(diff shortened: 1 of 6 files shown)


laixintao:
Good news, we can using f-string in pgcli's codebase now! 😀

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.black]
 line-length = 88
-target-version = ['py35']
+target-version = ['py37']
```
j-bennet: I think we should keep the target version at 3.6 (lowest we're going to support). We don't want to introduce 3.7 language features that some users may not have.

j-bennet:
Nice work @laixintao ! Thank you! 🍒

PR #1156, opened 2020-03-13
removed py2-related stuff

pasenor:
since Python 2 is now deprecated, this compatibility magic is no longer needed


j-bennet:
I think we don't have much usage of `six`, but can remove that as well.

pasenor:
aha, and encoding declarations (by the way, there are too few of them, I wonder how it used to work).
All removed.

laixintao:
LGTM, merged!

PR #1157, opened 2020-03-18
python3.8: add travis test and setup.py class.

laixintao:
## Description

test in python38.

## Checklist

- [x] I've added this contribution to the `changelog.rst`.
- [x] I've added my name to the `AUTHORS` file (or it's already there).

- [x] I installed pre-commit hooks (`pip install pre-commit && pre-commit install`), and ran `black` on my code.
- [x] Please squash merge this pull request (uncheck if you'd like us to merge as multiple commits)

PR #1158, opened 2020-03-18
upgrade coverage to 5.x

laixintao:
## Description

This fixes the coverage issue from #1157 , it tunes out that the version of `coverage.py` we are using is broken on Python3.8, which produces some phantom lines. see https://github.com/nedbat/coveragepy/issues/714 (which is actually interesting).

So simply upgrade coverage.py solves the problen.

## Checklist

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -4,7 +4,7 @@ tox>=1.9.2
 behave>=1.2.4
 pexpect==3.3
 pre-commit>=1.16.0
-coverage==4.3.4
+coverage==5.0.4
 codecov>=1.5.1
 docutils>=0.13.1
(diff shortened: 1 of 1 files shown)


laixintao:
The line data from @pasenor 's script draws my attention( https://github.com/dbcli/pgcli/pull/1157#issuecomment-600769740 ), I don't the difference is normal, but the script is correct and I even checked the lines manually.

 I figured it out by running coverage locally, this is html report from Python3.6 `pgcli/main.py`:

image

And this from Python3.8:

image

Then google lead me to https://github.com/nedbat/coveragepy/issues/714

j-bennet:
@laixintao Good catch!

PR #1164, opened 2020-03-30
Add --help output to README

clbarnes:
Additionally, refer to environment variables which can be used to configure login options.


laixintao:
Thanks, I think it's a good idea about showing `--help` in readme.md in case of others miss `alias_dsn` feature :)  cc @amjith @j-bennet

Issue #1169, opened 2020-04-17
Prevent Prompt Newline - Feature/Question


stephanoparaskeva:
How do I stop the newline for each prompt, in psql I can have all my commands one above the other
where as in pgcli they're seperated by newlines (even when timing=False)?

<img width="324" alt="Screenshot 2020-04-17 at 18 38 46" src="https://user-images.githubusercontent.com/40718272/79597717-b0767300-80da-11ea-9077-94ceaa34969f.png">

stephanoparaskeva:
Fix if in PR #1171

PR #1171, opened 2020-04-18
bugfix: don't echo a newline when output is empty.

laixintao:
## Description
fix https://github.com/dbcli/pgcli/issues/1169

## Checklist

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -660,7 +660,8 @@ class PGCli(object):
                     except IOError as e:
                         click.secho(str(e), err=True, fg="red")
                 else:
-                    self.echo_via_pager("\n".join(output))
+                    if output:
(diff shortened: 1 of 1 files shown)


laixintao:
before:
image


after:
image

j-bennet:
@laixintao Good catch! Please add it to changelog as well.

amjith:
Very nice! 

I'll update the changelog. Thanks, @laixintao! 

:wine_glass:

laixintao:
> @laixintao Good catch! Please add it to changelog as well.
> 
> P.S. Yes, I think we want this change. It makes the output more compact, which might help if someone wants to copy it from the screen, and it aligns the behavior with other cli tools.

I think the same as well :)

@amjith Thanks for changelog :)

Issue #1087, opened 2019-08-08
Autocomplete on Enter


agzam:
Is there a way to make it, so pressing Enter would just pick up very first suggestion and use it? It's really annoying to have to press `→` and then Enter.

j-bennet:
This feature might end up doing more harm than good. In multiline mode, `enter` to complete the first choice would be annoying and inconvenient. In case of a destructive query (`DROP` or `DELETE`, it would make it too easy to select the wrong suggestion. I don't think we want to do this.

agzam:
Can't we have a flag in the settings that's ignored in multiline mode?

j-bennet:
We try very hard not to add extra config settings. Also, in addition to multiline mode, this would require special handling for destructive queries, and I'm sure there are other use cases I'm not thinking about right now where accidental first choice would be bad. There's not enough benefit to justify the potential danger.

telmotrooper:
Sorry to revive this old thread, but would it be that bad if, with the option manually set by the user, pressing ENTER would just type in the first suggestion but not actually execute it?

That would allow me to work with `pgcli` similarly to how I work with `DBeaver`, right now when I switch between them it feels really confusing.

j-bennet:
@telmotrooper so basically you'd like Enter to work the same as Tab key?

telmotrooper:
Optionally, yes. 🙂️

j-bennet:
@telmotrooper I have a branch with the change you requested. Can you give it a try and see if it works as you expect? Here is the command to install from branch:

`pip install git+https://github.com/dbcli/pgcli.git@j-bennet/enter-accept-first`

j-bennet:
@telmotrooper Did you have the cnagce to try pgcli from the branch?

telmotrooper:
@j-bennet Hey, just tried it out in the branch. That's exactly what I had in mind! Thank you for implementing it.

It feels way more natural to switch between DBeaver and pgcli this way. I'll be glad if you guys decide to implement it as a toggleable feature in the main branch, for now I'll be using the feature branch. 😁️

j-bennet:
@telmotrooper Try reinstalling from branch - I made a change so that it should work in multiline mode as well. It would be great if you can run this branch for a while and report back whether or not you see any unexpected / weird keybinding behaviors.

telmotrooper:
@j-bennet Will do. Thanks!

j-bennet:
@telmotrooper did the change work for you?

[...]

PR #1179, opened 2020-05-08
Autocomplete first suggestion on ENTER

j-bennet:
Implementing a feature requested in https://github.com/dbcli/pgcli/issues/1087#issuecomment-610611560.

The behavior in this branch on "accepting" * remains slightly odd. However, this is the only use case that's not handled perfectly, and that is because I'm not quite sure what "perfectly" should look like here. I'd like want to merge what I have anyway.

### DIFF
--- a/pgcli/key_bindings.py
+++ b/pgcli/key_bindings.py
@@ -106,7 +106,14 @@ def pgcli_bindings(pgcli):
     )
     def _(event):
         _logger.debug("Detected enter key.")
-        event.current_buffer.validate_and_handle()
+
+        buf = event.app.current_buffer
+        doc = buf.document
+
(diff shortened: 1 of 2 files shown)


j-bennet:
Nevermind; closing the PR. The current branch behavior is not great in the very common use-case: typing something and running it without selecting anything from autocomplete drop-down. Consider this:

### on master:
 - type `\l`
 - press ENTER without selecting anything from autocomplete dropdown

The query is executed. 2 steps total.

### on this branch:
 - type `\l`
 - press ENTER. First suggestion is selected.
 - press ENTER again. First suggestion is accepted.
 - press ENTER again. The query finally runs.

4 steps. We're adding 2 steps that the user didn't need at all. I don't like this. Current behavior is fine.

PR #1198, opened 2020-07-27
More explicit error message when DSN alias is not found

tomplex:
## Description

Today I made a typo in a DSN alias while connecting, and the "invalid DSN" error made me look at my pgcli config file, wondering how it changed, before I realized I just misspelled the alias. So I figured it might be nice to make the error for that case more explicit.

I'm happy to add a test or two but I couldn't find any other tests for the `cli` function / error cases to base it off of.

## Checklist

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -1284,7 +1284,15 @@ def cli(
         try:
             cfg = load_config(pgclirc, config_full_path)
             dsn_config = cfg["alias_dsn"][dsn]
-        except:
+        except KeyError:
+            click.secho(
+                f"Could not find a DSN with alias {dsn}. "
+                'Please check the "[alias_dsn]" section in pgclirc.',
+                err=True,
+                fg="red",
+            )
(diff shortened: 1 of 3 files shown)


pgcli/main.py
```diff
@@ -1284,7 +1284,15 @@ def cli(
         try:
             cfg = load_config(pgclirc, config_full_path)
             dsn_config = cfg["alias_dsn"][dsn]
-        except:
+        except KeyError:
+            click.secho(
+                f"Could not find a DSN with alias {dsn}. "
```
tomplex, line 1289: I used an f-string because I see you support Python3.6+, but I'm happy to switch to a different form of string interpolation if there's a preferred method for the package.

j-bennet: We are very happy with f-strings now that we don't have to support python 2. :)

j-bennet:
Well done, thank you @tomplex! 🍓

PR #1216, opened 2020-10-15
Support history_file configuration per dsn, and --history option.

laixintao:
## Description

This feature enables the user to config history file location for specific dsn.

close: #535 

## TODOs

- [ ] `--hisfile` support
- [ ] history file per database
  - [ ] switch history file when switch database
- [ ] alias dsn histfile

## Checklist

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -1194,6 +1194,9 @@ class PGCli(object):
 @click.option(
     "--warn/--no-warn", default=None, help="Warn before running a destructive query."
 )
+@click.option(
+    "--history", default=None, help="Specify history file location."
+)
 @click.argument("dbname", default=lambda: None, envvar="PGDATABASE", nargs=1)
 @click.argument("username", default=lambda: None, envvar="PGUSER", nargs=1)
 def cli(
@@ -1217,6 +1220,7 @@ def cli(
     auto_vertical_output,
     list_dsn,
     warn,
+    history,
 ):
     if version:
         print("Version:", __version__)
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -165,7 +165,7 @@ class PGCli(object):
         auto_vertical_output=False,
         warn=None,
         histfile=None,
-        alias_dsn=None
+        alias_dsn=None,
     ):
 
         self.force_passwd_prompt = force_passwd_prompt
@@ -1203,9 +1203,7 @@ class PGCli(object):
 @click.option(
     "--warn/--no-warn", default=None, help="Warn before running a destructive query."
 )
-@click.option(
-    "--histfile", default=None, help="Specify history file location."
-)
+@click.option("--histfile", default=None, help="Specify history file location.")
 @click.argument("dbname", default=lambda: None, envvar="PGDATABASE", nargs=1)
 @click.argument("username", default=lambda: None, envvar="PGUSER", nargs=1)
 def cli(
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -164,6 +164,8 @@ class PGCli(object):
         prompt_dsn=None,
         auto_vertical_output=False,
         warn=None,
+        histfile=None,
+        alias_dsn=None
     ):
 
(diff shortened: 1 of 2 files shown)


pgcli/main.py
```diff
@@ -1194,6 +1194,9 @@ def echo_via_pager(self, text, color=None):
 @click.option(
     "--warn/--no-warn", default=None, help="Warn before running a destructive query."
 )
+@click.option(
+    "--history", default=None, help="Specify history file location."
```
j-bennet, line 1198: Perhaps `histfile`, so it matches the name used by `psql`?

pgcli/pgclirc
```diff
@@ -187,6 +188,11 @@ output.null = "#808080"
 [alias_dsn]
 # example_dsn = postgresql://[user[:password]@][netloc][:port][/dbname]
 
+# Specify sperate history file for alias_dsn.
+# If not set, will use history_file settings.
+[dsn_history_file]
```
j-bennet, line 193: We can support mapping of history file to not only DSN aliases, but database names as well. That way, if you have `mydb1` hosted on different servers, you could just configure:

```
[db_history_file]
mydb1 = ~/config/pgcli/history_mydb1
```

this could be useful to users that do not use DSN aliases.

There will be ambiguity if we do that: we'd have to figure out if `mydb1` is a DSN alias or a database name. We can look for matching DSN alias first, if user provided `--dsn` when running the cli, and if there's no alias found, assume this is database name.

j-bennet (commented):
Left some comments for you.

pgcli/main.py
laixintao: Excellent! I knew I met `--histfile` somewhere but I forgot it's `psql`, I just haven't use psql for a long time. :)

pgcli/pgclirc
laixintao: Yes, I like the idea that mapping of the history file for the database. I think it covers most of the use cases. Maybe consider `alias_dsn` is kind of overthinking and over-designed, I think we add too mush to configuration, and cloud leads to misunderstanding. I don't think users want to share history between databases. Even if they want to do that, they can use `--histfile` option.

Let's see, we only add a flag `sperate_history_on_database_name = False`, and add `--histfile` option. (Also, if user use `\c[onnect] database_name`, we should switch history file as well.

If the user does want to share history between database, they can:

1. Use `--histfile`, maybe use shell's alias together.
2. Save the most used query to `named query`.

What do you think?

j-bennet: I think that:

* separating history per database does not need a config flag. Once that feature is in, we should do it by default. When you're connected to a specific database, queries from other databases are not very useful / relevant, and would only clutter history.
* user can use `--histfile` to override the default history-file-per-database.
* mapping to DSN vs database name can be useful. I had a use case with `prod` and `beta` database tables named `table1_prod`, `table2_prod` etc. Table structure was the same, but table names were different, so you could not use exactly the same query on those databases. This may be an edge case.

laixintao: I am convinced, I will only add a section named `dsn_history_file` section(maybe we cloud come up with a better name), and left other configs untouched.

So the process of choosing a history file will look like this:

image


Do I understand it correctly?

j-bennet: Yes, this looks exactly right.

gdevanla:
I like this feature and would love this PR to come through. @laixintao are you still working on this. Is there something I can do to help?

laixintao:
Yes still working on it, but I just changed my job and got married last
month so this PR was delayed.

I will get back to this recently, sorry.

laixintao:
I am so sorry, totally forgot this PR. I will get back to this

saucoide:
@laixintao i'd like to help finish this one, do you mind if i try to pick it up?

laixintao:
> @laixintao i'd like to help finish this one, do you mind if i try to pick it up?

sure, thanks! (shame on me 🥺)

PR #1241, opened 2021-02-10
Fix issue where `syntax_style` config value would not have any effect

### LINKED ISSUE #1212: Syntax colors config doesn't work
## Description
Syntax colors config in .~/config/pgcli/config doesn't work
```python
# Syntax Style. Possible values: manni, igor, xcode, vim, autumn, vs, rrt,
# native, perldoc, borland, tango, emacs, friendly, monokai, paraiso-dark,
# colorful, murphy, bw, pastie, paraiso-light, trac, default, fruity
syntax_style = <color scheme>
```
I changed the value of syntax_style but the syntax color in pgcli is always the same as "default" in https://www.pgcli.com/syntax

## Your environment

Ubuntu 20.10
Server: PostgreSQL 12.4 (Ubuntu 12.4-1)
Version: 3.0.0
pgcli @ git+https://github.com/dbcli/pgcli@7626d9a5f27a221d87f31dcb5bb8fc12024f6c58f57829411793929a7a
        installed via git
Pygments==2.7.0

ruohola:
## Description

These config properties got introduced in 41dd24e8 as a means to have
more granular control over the syntax highlighting. The problem is that
these cannot be in the default config file since `get_config()` always
reads both the default config file and the user specified one, and there
is no way to unset these variables in the user specified config file to
restore their default behavior. Even if there would be a way, it
wouldn't be intuitive at all to be required to unset some random
settings under the `[colors]` section just to be able to use the well
documented `syntax_style` setting.

Note that one *can* still set these three lines in their user config
file if they want to utilize them.

Resolves #1212

## Checklist

### DIFF
--- a/pgcli/pgclirc
+++ b/pgcli/pgclirc
@@ -170,9 +170,12 @@ arg-toolbar = 'noinherit bold'
 arg-toolbar.text = 'nobold'
 bottom-toolbar.transaction.valid = 'bg:#222222 #00ff5f bold'
 bottom-toolbar.transaction.failed = 'bg:#222222 #ff005f bold'
-literal.string = '#ba2121'
-literal.number = '#666666'
-keyword = 'bold #008000'
+# These three values can be used to further refine the syntax highlighting.
+# They are commented out by default, since they have priority over the theme set
+# with the `syntax_style` setting and overriding its behavior can be confusing.
+# literal.string = '#ba2121'
+# literal.number = '#666666'
+# keyword = 'bold #008000'
 
 # style classes for colored table output
 output.header = "#00ff5f bold"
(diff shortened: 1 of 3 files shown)


pgcli/pgclirc
```diff
@@ -170,9 +170,6 @@ arg-toolbar = 'noinherit bold'
 arg-toolbar.text = 'nobold'
 bottom-toolbar.transaction.valid = 'bg:#222222 #00ff5f bold'
 bottom-toolbar.transaction.failed = 'bg:#222222 #ff005f bold'
-literal.string = '#ba2121'
```
j-bennet, line 173: Apparently it's not obvious that colors set in this section will override colors set by user's choice of `syntax_style`. Could you please add a comment to this section to make it clear?

j-bennet (approved):
Nice catch @ruohola. I think an extra comment in pgclirc would be useful, and I'll merge this right away. Thank you!

pgcli/pgclirc
ruohola: Fixed! :)

j-bennet:
Great work. Merging!

Issue #1240, opened 2021-02-04
User comments in config file lost after writing to config via `pgcli`


sevens-ef:
## Description
First of all, thanks for a great tool, I use it a lot.

User comments to the pgcli config file are getting lost after the config file gets modified via `pgcli` itself (e.g. adding a named query).  According to https://github.com/dbcli/pgcli/issues/1162#issuecomment-740076796 they should remain.  Additionally, all whitelines from at least the `[named queries]` section onward get lost as well.

I can reproduce this on a clean virtualenv and with a fresh `config` file.

## Reproduce
Log from my reproduction attempts:
* Removed previous config file
* Created new temp virtualenv (using the system Python 3.6.9), installed new pgcli (`pip install pgcli`)
* Started `pgcli`, this created a new config file
* Config file (from `[named queries]` section onward; same goes for the config file snippets below) at this point:
  ```!ini
  # Named queries are queries you can execute by name.
  [named queries]
  
  # DSN to call by -D option
  [alias_dsn]
  # example_dsn = postgresql://[user[:password]@][netloc][:port][/dbname]
  
  # Format for number representation
  # for decimal "d" - 12345678, ",d" - 12,345,678
  # for float "g" - 123456.78, ",g" - 123,456.78
  [data_formats]
  decimal = ""
  float = ""
  ```
* Save new named query from `pgcli`: `\ns foo bar`.  Config file snippet after this:
  ```ini
  [named queries]
  foo = bar
  [alias_dsn]
  [data_formats]
  decimal = ""
  float = ""
  ```
  * *NOTE*: the "# Named queries are queries you can execute by name." comment (and the preceding whiteline) *above* the `[named query]` section header is now gone as well (while the "# style classes for colored table output" comment above `output.header = "#00ff5f bold"` remained in tact.  As did all other comment and whitelines earlier in the file)
* Close `pgcli` (just in case), add user comment above the `foo = bar` named query entry.  Config file snippet after this:
  ```ini
  [named queries]
  # My comment
  foo = bar
  [alias_dsn]
  [data_formats]
  decimal = ""
  float = ""
  ```
* Start `pgcli` again, add new named query: `\ns bar SELECT 1`.  Config file snippet after this:
  ```ini
  [named queries]
  foo = bar
  bar = SELECT 1
  [alias_dsn]
  [data_formats]
  decimal = ""
  float = ""
  ```
[...]

PR #1247, opened 2021-02-22
Fix test_config.py on Windows

ERYoung11:
## Description
I'm developing on Windows with VS Code.  test_config.py failed for the final case as setting a directory to read-only did not prevent writing to that directory on Windows.  I have updated the code to try name the subdir with an illegal character which prevents the creation of rcfile.  I used the double-quote character.  This character is legal on Linux.  This test now passes on Linux (when run as non-root) and Windows.

Wrappager was also marked as changed due to, I'm guessing, some hidden characters.  I have checked it in with this too.  As you can see, github thinks there are no changes to the file.

## Checklist


ERYoung11:
I'm no expert here, but I don't think this should have failed due to timeout as this was a very insignificant change.

j-bennet:
@ERYoung11 Yes, I don't think the test failure was related to your change. Thank you for the PR! Feel free to update changelog / authors file, or I can merge it as-is if you prefer.

ERYoung11:
Go ahead and merge.  If I do something more significant I'll update those sections.  I'm currently looking at Issue #1240.  

Also, is the vagrant file out of date?  I can't find the chef/centos & chef/debian it's asking for.  I could fix that up too.

Is there somewhere that I can chat about these sorts of questions?  Fair warning, I'm sort of a newbie (and old newbie working to learn new tricks).

j-bennet:
> If I do something more significant I'll update those sections. I'm currently looking at Issue #1240.

Sounds good to me. Any and all contributions are appreciated.

> Also, is the vagrant file out of date? I can't find the chef/centos & chef/debian it's asking for. I could fix that up too.

Yes, it is quite outdated. If you're willing to take on updating it, please do.

> Is there somewhere that I can chat about these sorts of questions? Fair warning, I'm sort of a newbie (and old newbie working to learn new tricks).

We prefer to communicate via github issues. This gives us more visibility on when a contributor drops the ball / loses interest in the issue (happens all the time unfortunately), and someone else can pick it up, because all the context is there.

Issue #1257, opened 2021-03-10
comment skipping not working on Windows


ERYoung11:
## Description

For Windows, the code added in #1245 causes a legitimate failure of test_main.py.  It gives the error listed below (truncated).  I have a pull request incoming to deal with this.

The ParseError comes from ConfigObj.py because it cannot process the 'y' that is left over from the comment "because psql is crazy". 

Basically due to EOL differences between Linux and Windows, the pointer movement for the filestream doesn't go far enough.

 raise error
E             File "<string>", line None
E           configobj.ParseError: Invalid line ('y') (matched as neither section nor keyword) at line 1.

configspec = None
content    = ['y', '', '            [myservice]', '            host=a_host', '            user=a_user', '            port=5433', ...]
end        = '\n'
error      = ParseError("Invalid line ('y') (matched as neither section nor keyword) at line 1.")

....
self       = ConfigObj({'myservice': {'host': 'a_host', 'user': 'a_user', 'port': '5433', 'password': 'much_secure', 'dbname': 'a_dbname'}, 'my_other_service': {'host': 'b_host', 'user': 'b_user', 'port': '5435', 'dbname': 'b_dbname'}})

venv\lib\site-packages\configobj.py:1318: ParseError
_

## Your environment

- [X] Please provide your OS and version information.
-
Windows 10 64-bit

- [X] Please provide your CLI version.

Current master branch (after commit: 30212c6fc68ccf19d794530819cda8e6d4edf74c)

- [X] What is the output of ``pip freeze`` command.
- $ pip freeze
appdirs==1.4.4
atomicwrites==1.4.0
attrs==20.3.0
cfgv==3.2.0
cli-helpers==2.1.0
click==7.1.2
colorama==0.4.4
configobj==5.0.6
coverage==5.5
distlib==0.3.1
filelock==3.0.12
identify==1.5.14
iniconfig==1.1.1
mock==4.0.3
mocker==1.1.1
nodeenv==1.5.0
packaging==20.9
pendulum==2.1.2
pgspecial==1.12.1
pluggy==0.13.1
pre-commit==2.10.1
prompt-toolkit==3.0.16
psycopg2==2.8.6
py==1.10.0
Pygments==2.8.0
pyparsing==2.4.7
pytest==6.2.2
pytest-cov==2.11.1
pytest-mock==3.5.1
python-dateutil==2.8.1
pytzdata==2020.1
PyYAML==5.4.1
six==1.15.0
sqlparse==0.4.1
[...]

PR #1258, opened 2021-03-10
skipping initial comments in pgconfig file did not work on Windows.

ERYoung11:
## Description

See Issue #1257.

To fix this issue, we needed to have skip_initial_comment consume EOL in the same way no matter the OS.

codecs.open does not provide an option to consider EOLs for each platform in the same way.

[codecs.open is not favored for python 3. ](https://stackoverflow.com/questions/5250744/difference-between-open-and-codecs-open-in-python)

The built-in open does provide for a way to consider EOLs for each platform in the same way.

Therefore, I removed the import of codecs.open and used the standard open feature.

All tests now run on Windows and Linux and the pgcli command appears to work correctly.

## Checklist

- [ ] I've added this contribution to the `changelog.rst`. (the initial commit is already there so I'm not adding my fix)

- [X] I installed pre-commit hooks (`pip install pre-commit && pre-commit install`), and ran `black` on my code.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -116,6 +116,7 @@ Contributors:
     * Kevin Marsh (kevinmarsh)
     * Eero Ruohola (ruohola)
     * Miroslav Šedivý (eumiro)
+    * Eric R Young (ERYoung11) 
 
 Creator:
 --------
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -21,7 +21,6 @@ import datetime as dt
 import itertools
 import platform
 from time import time, sleep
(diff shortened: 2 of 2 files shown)


ERYoung11:
@pasenor  A fix for one of your recent commits.  Please take a look and make sure it makes sense to you.  Test suite passed on Windows and Linux.

pasenor:
Ouch, that's a good catch, thanks! I think `codecs.open` was just a remnant of the py2-compatibility, definitely time for it to go.

pasenor (approved):

PR #1263, opened 2021-05-02
Fix: ANSI escape codes in first line make the cli choose expanded output incorrectly...

psacawa:
## Description

The `format_output` procedure chooses to output as a table or in the expanded    format on the basis of the length of the first line of the table formatter. However, this first line contains a lot of ANSI    escape codes, so this comparison is wrong, and the expanded mode   is used either way. This patch strips these escape codes before the    comparison.

To reproduce the bug, set `expand_auto = True` in a narrow terminal (number of rows < 100) and try `select 1`. The output is narrow enough to fit in the terminal, but because of the ANSI escape codes, `len(first_line)` exceeds `max_width` in `format_output`, so we get expanded output.

Screencast of the bug fixed: https://asciinema.org/a/vjoWc9OeHxlXvg0joNDLT40rv

## Checklist

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -117,6 +117,7 @@ Contributors:
     * Eero Ruohola (ruohola)
     * Miroslav Šedivý (eumiro)
     * Eric R Young (ERYoung11) 
+    * Paweł Sacawa (psacawa)
 
 Creator:
 --------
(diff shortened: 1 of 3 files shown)


amjith:
Looks like Black formatting is failing. I'll merge it now and fix the Black formatting in master. 

Thanks for the PR. :+1:

PR #1264, opened 2021-05-05
Add support for \clip command

psacawa:
## Description

When profiling Postgres queries with `EXPLAIN ANALYZE`, it is often necessary to copy the query to the system clipboard in order to feed the output to a visualisation tool like (depesz)[explain.depesz.com]. `psql` has no facilities for this, so I usually need to do something like 
```
psql -qAt --no-psqlrc -f cmd.sql | xsel -ib
```
This is annoying, so I see the advantage of a `\clip` command that copies the DB output to the clipboard, which lets you do
```
\clip explain analyze select * from tab;
```

It is different from the native DB commands, so I don't see how to implement it cleanly. It doesn't belong in `pgspecial`. The present implementation is a bit ad hoc. It uses the package `pyperclip`, which implements a cross-platform clipboard client.

I am submitting this PR to solicit comments, with the hope to work toward a good solution. **Please allow me to run workflows**

## Checklist

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -16,6 +16,7 @@ Bug fixes:
 * Fix comments being lost in config when saving a named query. (#1240)
 * Fix IPython magic for ipython-sql >= 0.4.0
 * Fix pager not being used when output format is set to csv. (#1238)
+* Add \clip command to copy to clipboard
 
 3.1.0
 =====
--- a/pgcli/pgexecute.py
+++ b/pgcli/pgexecute.py
@@ -385,6 +385,26 @@ class PGExecute:
                             self.reset_expanded = True
                         sql = sql[:-2].strip()
 
+                    #  stub implementation
+                    #  copying the output is useful for profiling queries with EXPLAIN
+                    if sql.startswith(r"\clip"):
+                        sql = sql.strip(r"\clip")
+                        cur = self.conn.cursor()
+                        cur.execute(sql)
+                        out_lines = cur.fetchall()
+                        output = "\n".join([tup[0] for tup in out_lines])
+
+                        #  cross-platform clipboard copy
+                        _logger.debug("copying to clipboard")
+                        import pyperclip
+
+                        pyperclip.copy(output)
+
+                        #  still print output
+                        #  since it is likely a explain analyze query, we may like to use an
+                        #  existing Pygments lexer for Postgres explain output
+                        print(output)
+
                     # First try to run each query as special
                     _logger.debug("Trying a pgspecial command. sql: %r", sql)
                     try:


amjith:
This does belong in pgspecial package. We do have custom commands that are not DB commands in pgspecial. 

Take a look at the [iocommands](https://github.com/dbcli/pgspecial/blob/master/pgspecial/iocommands.py) file in pgspecial. 

Also have a look at the [develop.rst](https://github.com/dbcli/pgspecial/blob/master/DEVELOP.rst) file, it talks a bit about how to add special commands.

amjith:
Here's a reference implementation from mycli: https://github.com/dbcli/mycli/blob/master/mycli/packages/special/iocommands.py#L165

okbob:
psql has command \g |

<pre>
postgres=# explain select * from o
obce         obce_id_seq  okresy       
postgres=# explain select * from obce limit 10 \g | wl-copy
postgres=# 
</pre>

There are lot of use cases - copy to clipboard is just one.

psacawa:
@okbob, thanks for informing me about this. It is definitely very useful. But for my case it isn't enough. This is because I typically need the JSON format for [pev2](https://explain.dalibo.com/), and the output of `EXPLAIN (ANALYZE, format json) SELECT * from mytable \g | xsel -ib` gives me:
```
QUERY PLAN              
-------------------------------------
 [                                  +
   {                                +
     "Plan": {                      +
       "Node Type": "Seq Scan",     +
       "Parallel Aware": false,     +
       "Relation Name": "mytable",  +
       "Alias": "mytable",          +
       "Startup Cost": 0.00,        +
       "Total Cost": 35.50,         +
       "Plan Rows": 2550,           +
       "Plan Width": 4,             +
       "Actual Startup Time": 0.005,+
       "Actual Total Time": 0.005,  +
       "Actual Rows": 0,            +
       "Actual Loops": 1            +
     },                             +
     "Planning Time": 0.094,        +
     "Triggers": [                  +
     ],                             +
     "Execution Time": 0.029        +
   }                                +
 ]
(1 row)
```
So unless there is some some `\pset`  option I don't know, then a more fine-grained is required.

okbob:
út 18. 5. 2021 v 10:07 odesílatel Paweł Sacawa ***@***.***>
napsal:

> @okbob <https://github.com/okbob>, thanks for informing me about this. It
> is definitely very useful. But for my case it isn't enough. This is because
> I typically need the JSON format for pev2 <https://explain.dalibo.com/>,
> and the output of EXPLAIN (ANALYZE, format json) SELECT * from mytable \g
> | xsel -ib gives me:
>
> QUERY PLAN
> -------------------------------------
>  [                                  +
>    {                                +
>      "Plan": {                      +
>        "Node Type": "Seq Scan",     +
>        "Parallel Aware": false,     +
>        "Relation Name": "mytable",  +
>        "Alias": "mytable",          +
>        "Startup Cost": 0.00,        +
>        "Total Cost": 35.50,         +
>        "Plan Rows": 2550,           +
>        "Plan Width": 4,             +
>        "Actual Startup Time": 0.005,+
>        "Actual Total Time": 0.005,  +
>        "Actual Rows": 0,            +
>        "Actual Loops": 1            +
>      },                             +
>      "Planning Time": 0.094,        +
>      "Triggers": [                  +
>      ],                             +
>      "Execution Time": 0.029        +
>    }                                +
>  ]
> (1 row)
>
> So unless there is some some \pset option I don't know, then a more
> fine-grained is required.
>
It is solved in psql 13, where you can use one shot parameters for \g

explain (format json) select * from pg_class limit 10 \g (format=unaligned
tuples_only) | wl-copy

[
  {
    "Plan": {
      "Node Type": "Limit",
      "Parallel Aware": false,
      "Async Capable": false,
      "Startup Cost": 0.00,
      "Total Cost": 0.43,
      "Plan Rows": 10,
      "Plan Width": 265,
      "Plans": [
        {
          "Node Type": "Seq Scan",
          "Parent Relationship": "Outer",
          "Parallel Aware": false,
          "Async Capable": false,
          "Relation Name": "pg_class",
          "Alias": "pg_class",
          "Startup Cost": 0.00,
          "Total Cost": 16.96,
          "Plan Rows": 396,
          "Plan Width": 265
        }
      ]
    }
  }
]

PR #1278, opened 2021-08-05
correct config path

lucas-seismic:
## Checklist

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -125,7 +125,7 @@ The `pgcli` is written using prompt_toolkit_.
 
 Config
 ------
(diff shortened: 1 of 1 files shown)


j-bennet (changes requested):
Thank you for the PR! This path, `~/.config/pgcli/config`, is not incorrect. It's OS-specific:

https://github.com/dbcli/pgcli/blob/e0a4c18c4a074016beef46886b03e456102e2db1/pgcli/config.py#L11-L17

We should mention that it's OS- and platform-specific in the README, but there's no need to correct the path.

PR #1279, opened 2021-08-07
add explain visualizer

dkuku:
## Description
adds a formatter for explain analyze output
based on https://github.com/dkuku/pyev
## Checklist


dkuku:
tbh I have no idea whats the best way to write tests for it ?

amjith:
The explain plan looks gorgeous. :smile: 

I'm still playing around with it, haven't had a chance to look through the code yet. 

Looks like some of the colors you've used from click aren't available in the older versions of click. So we should consider requiring a modern version of click in setup.py.

amjith:
@dkuku Instead of binding the explain mode to F5 key can we detect an EXPLAIN query and automatically display the results using ExplainVisualizer?

dkuku:
@amjith we can but you need to pass some params to the explain query and when these are missing it may crash due to missing fields or the format (it needs json). I was tinking more of having an easy way to show the explain for your current query. I think its more useful than having a way to switch between vim/emacs mode with a shortcut because you use one or the other?

j-bennet:
Hey @dkuku, sorry it's been forever since you submitted this PR, but I think it's good to merge. Would you be able to rebase it on top of main and resolve the conflicts?

dkuku:
Hi @j-bennet I merged master into this branch.

j-bennet:
Merged. Nice work! 🍫

Issue #1286, opened 2021-09-08
stripping comments


edoardoc:
Hello,

I am tagging queries and I was using pgcli to exec them, things like this:

```
/* TEST123 TAGGING */ select 1
```

it looks to me that pgcli does not send /* */ or -- comments to the backend.
Is this the desired behaviour?

edoardoc:
I'd been asking the same question...
I have been analyzing pretty big queries recently and in postgres there is a postgresql.conf log feature set
`auto_explain`.
With that you can tune pg to automatically log the whole execution plan each time the query is executed (eventually splitting it in sub queries, etc).
While doing this it also logs all the comments  that were included in SQL.
These comments can then be used to match "query part" <----> "execution plan" in this enormous log files. 
And they are also used by some tools doing this automatically, like [https://pganalyze.com/postgres-analyze-query-performance](pganalyze)

j-bennet:
TIL. I believe that `psycopg2` can execute SQL queries with comments in them, it should be easy enough to remove comment stripping in a branch and see what happens. I don't have the bandwidth to do this right now, so I'll leave the issue open as an enhancement request, or feel free to submit a PR yourself.

PR #1296, opened 2021-11-12
no need to strip_comments

edoardoc:
## Description

just a  minor change to make sure SQL comments do not get lost before reaching the backend.
as discussed in https://github.com/dbcli/pgcli/issues/1286

I have run a comparison with psql on a query like

 /* TEST123 TAGGING */ select * from /*test 123 */ bizscheme.accounts limit 10;--jkjkjksdds fsdf sdf sd

and postgresql receives comments in the same way now


j-bennet (approved):
Nice. Thank you for the PR!

Issue #1320, opened 2022-03-01
Result Set Decoration Missing Number of Rows


clairesarraille:
## Description

**CURRENT BEHAVIOR:**
The footer, or result set decoration, at the end of the query output is "SELECT" but does not include the number of rows. Expected behavior for the image below is "SELECT 7" rather than just "SELECT" above "Time: 0.224s"

> seg> select * from test_table;
> +------------------+--------+
> | sessionstartdate | sum    |
> |------------------+--------|
> | 2021-01-15       | 491076 |
> | 2021-01-19       | 448897 |
> | 2021-01-17       | 477568 |
> | 2021-01-20       | 464592 |
> | 2021-01-21       | 442343 |
> | 2021-01-18       | 474817 |
> | 2021-01-16       | 456265 |
> +------------------+--------+
> SELECT
> Time: 0.152s

**EXPECTED BEHAVIOR:**
"SELECT 7" in footer

**NOTE:**
When I connect to my database using "psql" rather than "pgcli" the number of rows is included in the footer.

> seg=# select * from test_table
> seg-# ;
>  sessionstartdate |  sum   
> ------------------+--------
>  2021-01-17       | 477568
>  2021-01-15       | 491076
>  2021-01-19       | 448897
>  2021-01-21       | 442343
>  2021-01-20       | 464592
>  2021-01-16       | 456265
>  2021-01-18       | 474817
> (7 rows)
> 

## Your environment
I'm connecting to Segment db via AWS
`pgcli -h xx-segment-xxx.xxxxxxxxxxxx.xx-xxxx-x.redshift.amazonaws.com -U [USERNAME] -d seg -p [PORT_NUMBER]`

- [ ] Please provide your OS and version information.
```
macOS Monterey
Version 12.2.1
pgcli: stable 3.4.0
```
- [ ] Please provide your CLI version.
`OS Mac Terminal Version 2.12 (443)`
- [ ] What is the output of ``pip freeze`` command.
```
absl-py==0.10.0
aiohttp @ file:///Users/runner/miniforge3/conda-bld/aiohttp_1602487906891/work
appnope @ file:///Users/runner/miniforge3/conda-bld/appnope_1606661752827/work
argon2-cffi @ file:///Users/runner/miniforge3/conda-bld/argon2-cffi_1625821489228/work
arviz @ file:///home/conda/feedstock_root/build_artifacts/arviz_1600909298598/work
astunparse==1.6.3
async-generator==1.10
[...]

PR #1321, opened 2022-03-02
Add rowcount to status returned from redshift.

j-bennet:
## Description

Fix for https://github.com/dbcli/pgcli/issues/1320.

Normally, pgcli simply prints `cursor.statusmessage` that `psycopg2` returns. However, while with other databases `psycopg2` returns `SELECT X`, with Redshift this is just `SELECT`.

This adds missing rowcount in pgcli.

@dvarrazzo This PR does the job, but I wonder if it should be fixed in `psycopg2` instead. What do you think?

## Checklist

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -1,7 +1,10 @@
 TBD
 ===
 
-* [List new changes here].
+Bug fixes:
+----------
+
+* Fix the bug with Redshift not displaying wor count in status ([related issue](https://github.com/dbcli/pgcli/issues/1320)).
 
 3.4.0 (2022/02/21)
 ==================
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -1550,7 +1550,7 @@ def format_output(title, cur, headers, status, settings):
     def format_status(cur, status):
         # redshift does not return rowcount as part of status.
         # See https://github.com/dbcli/pgcli/issues/1320
-        if cur and hasattr(cur, 'rowcount') and cur.rowcount is not None:
+        if cur and hasattr(cur, "rowcount") and cur.rowcount is not None:
             if status and not status.endswith(str(cur.rowcount)):
                 status += " %s" % cur.rowcount
         return status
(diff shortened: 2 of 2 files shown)


dvarrazzo:
Hi @j-bennet, I'm sorry, I missed your ping about this.

That string is returned by the server as part of the communication between backend and frontend. It is not psycopg role to change it. Redshift is not Postgres, it's a different database, with different features; while it's ok to support it (as a side effect of using the same be-fe protocol of psycopg), it wouldn't be right to change its messages.

TBH I don't think you should have changed that string in dbcli either: the status message is a message from the server and I think it should be displayed as it is, or omitted if it doesn't add info, but not altered. The string the OP mentions psql prints (`(7 rows)`) is not the statusmessage: afaics, psql omits to print the status message when it looks like `SELECT...`, and prints `(N rows)` when the result contains rows, for instance:

```
piro=# insert into test values ('foo'), ('bar') returning *;
┌──────┐
│ data │
├──────┤
│ foo  │
│ bar  │
└──────┘
(2 rows)

INSERT 0 2
```

Here, `INSERT 0 2` is the status message; `(2 rows)` is a psql addition obtained from rowcount.

What to do is your project's choice of course. Being psycopg a driver, it is "a mechanism, not a policy", and replacing a message with a different message is totally a policy decision. I would have expected dbcli to do the same, but I'm not familiar with the project, so I don't know whether it prefers to choose to do  "as the user expects" rather than "as the database says" :slightly_smiling_face: 

P.S. if you wish to emulate psql and present a row count only on results returning data, you can use `cursor.pgresult.status` in psycopg 3. I don't think the information is readily available in psycopg2.

```
>>> cur.execute("insert into test values ('foo') returning *")
>>> psycopg.pq.ExecStatus(cur.pgresult.status)
<ExecStatus.TUPLES_OK: 2>

>>> cur.execute("insert into test values ('foo')")
>>> psycopg.pq.ExecStatus(cur.pgresult.status)
Out[14]: <ExecStatus.COMMAND_OK: 1>
```

PR #1324, opened 2022-03-17
Port to psycopg3

j-bennet:
## Description

Port pgcli to psycopg3. Work in progress.

## Checklist


pgcli/pgexecute.py
```diff
@@ -508,8 +521,14 @@ def execute_normal_sql(self, split_sql):
 
         # conn.notices persist between queies, we use pop to clear out the list
```
dvarrazzo, line 522: Comment out of date

j-bennet:
@dvarrazzo Can you advise about `register_json_typecasters` here?

https://github.com/dbcli/pgcli/blob/366c01bbb97fee96a8169629854af7c83d626c99/pgcli/pgexecute.py#L105

Do we still need this, and what would be the equivalent in psycopg if yes?

dvarrazzo:
> @dvarrazzo Can you advise about register_json_typecasters here?

@j-bennet I seem to remember that you only use that function to actually disable json loading, by calling `register_json_typecasters(conn, lambda x: x)`, so that people would see the json data instead of the repr of the python objects. If that's the case, I suggest you to register the text loader for it as you do in `register_typecasters()` with other types:

```python
def register_typecasters(connection):
    ...
    connection.adapters.register_loader("json", psycopg.types.string.TextLoader)
    connection.adapters.register_loader("jsonb", psycopg.types.string.TextLoader)
```

and to drop `register_json_typecaster()` altogether.

A function similar to the psycopg2 one is available as `psycopg.types.json.set_json_loads` (see [Json adaptation](https://www.psycopg.org/psycopg3/docs/basic/adapt.html#json-adaptation) for details), but I don't think it covers your use case. That function is meant to customise `loads()`; you need to disable it.

The exception checking in the old code is not necessary. It was necessary in the past to look for the oid of json/jsonb types, when they were still Postgres extensions. Now that they are in core, their oids are fixed and well known, so psycopg doesn't need to run queries which might fail. Even if you run pgcli with an old Postgres version, the function calls in `register_typecaster()` wouldn't fail.

.github/workflows/ci.yml
```diff
@@ -11,7 +11,8 @@ jobs:
 
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: ["3.10"]
+        # python-version: ["3.7", "3.8", "3.9", "3.10"]
```
amjith, line 15: Is this intentional? Are dropping support for the older versions for Python?

PR #1328, opened 2022-03-31
Changed destructive_warning to take a list of destructive commands

rigoneri:
## Description

* Changed the `destructive_warning` config to be a list of commands that are considered 
  destructive. This would allow you to be warned on `create`, `grant`, or `insert` queries.
* Updated the destructive warning message to display the dsn alias if available.

## Checklist

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -1,3 +1,13 @@
+Upcoming
+========
+
+Features:
+---------
+
+* Changed the `destructive_warning` config to be a list of commands that are considered 
+  destructive. This would allow you to be warned on `create`, `grant`, or `insert` queries.
+* Destructive warnings will now include the alias dsn connection string name if provided (-D option).
+
 3.5.0 (2022/09/15):
 ===================
 
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -64,6 +64,7 @@ from .config import (
 from .key_bindings import pgcli_bindings
 from .packages.formatter.sqlformatter import register_new_formatter
 from .packages.prompt_utils import confirm_destructive_query
+from .packages.parseutils import parse_destructive_warning
 from .__init__ import __version__
 
 click.disable_unicode_literals_warning = True
@@ -224,11 +225,10 @@ class PGCli:
         self.syntax_style = c["main"]["syntax_style"]
         self.cli_style = c["colors"]
         self.wider_completion_menu = c["main"].as_bool("wider_completion_menu")
-        self.destructive_warning = warn or c["main"]["destructive_warning"]
-        # also handle boolean format of destructive warning
-        self.destructive_warning = {"true": "all", "false": "off"}.get(
-            self.destructive_warning.lower(), self.destructive_warning
+        self.destructive_warning = parse_destructive_warning(
+            warn or c["main"].as_list("destructive_warning")
         )
+
         self.less_chatty = bool(less_chatty) or c["main"].as_bool("less_chatty")
         self.null_string = c["main"].get("null_string", "<null>")
         self.prompt_format = (
@@ -424,8 +424,11 @@ class PGCli:
             return [(None, None, None, str(e), "", False, True)]
 
         if (
-            self.destructive_warning != "off"
-            and confirm_destructive_query(query, self.destructive_warning) is False
+            self.destructive_warning
+            and confirm_destructive_query(
+                query, self.destructive_warning, self.dsn_alias
+            )
+            is False
         ):
             message = "Wise choice. Command execution stopped."
             return [(None, None, None, message)]
@@ -693,15 +696,16 @@ class PGCli:
         query = MetaQuery(query=text, successful=False)
 
         try:
-            if self.destructive_warning != "off":
+            if self.destructive_warning:
                 destroy = confirm = confirm_destructive_query(
-                    text, self.destructive_warning
+                    text, self.destructive_warning, self.dsn_alias
                 )
                 if destroy is False:
                     click.secho("Wise choice!")
                     raise KeyboardInterrupt
                 elif destroy:
                     click.secho("Your call!")
+
             output, query = self._evaluate_command(text)
         except KeyboardInterrupt:
             # Restart connection to the database
@@ -1266,7 +1270,6 @@ class PGCli:
 @click.option(
     "--warn",
     default=None,
-    type=click.Choice(["all", "moderate", "off"]),
     help="Warn before running a destructive query.",
 )
 @click.option(
--- a/pgcli/packages/parseutils/__init__.py
+++ b/pgcli/packages/parseutils/__init__.py
@@ -40,6 +40,9 @@ def is_destructive(queries, keywords):
 
 def parse_destructive_warning(warning_level):
     """Converts a deprecated destructive warning option to a list of command keywords."""
+    if not warning_level:
+        return []
+
     if not isinstance(warning_level, list):
         if "," in warning_level:
             return warning_level.split(",")
--- a/pgcli/packages/parseutils/__init__.py
+++ b/pgcli/packages/parseutils/__init__.py
@@ -1,6 +1,17 @@
 import sqlparse
 
 
+BASE_KEYWORDS = [
+    "drop",
+    "shutdown",
+    "delete",
+    "truncate",
+    "alter",
+    "unconditional_update",
+]
+ALL_KEYWORDS = BASE_KEYWORDS + ["update"]
+
+
 def query_starts_with(formatted_sql, prefixes):
     """Check if the query starts with any item from *prefixes*."""
     prefixes = [prefix.lower() for prefix in prefixes]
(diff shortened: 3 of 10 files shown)


.pre-commit-config.yaml
```diff
@@ -1,6 +1,6 @@
 repos:
 -   repo: https://github.com/psf/black
-    rev: 21.5b0
+    rev: 22.3.0
```
rigoneri, line 3: Black 22.3.0 fixed an issue with `>8.1.0` versions of click that comes with black https://github.com/psf/black/issues/2964

changelog.rst
```diff
@@ -1,3 +1,15 @@
+3.5.0 (2022/03/31)
```
j-bennet, line 1:
```suggestion
Upcoming
```
We don't know version number or date of next release yet. Just put "Upcoming" here for now, to be updated by the maintainer when releasing.

pgcli/main.py
```diff
@@ -181,6 +181,7 @@ def __init__(
         prompt_dsn=None,
         auto_vertical_output=False,
         warn=None,
+        env_name=None,
```
j-bennet, line 184: It seems excessive to add a new config option just to warn that "You're about to run a destructive command in PROD". You could serve the same purpose by simply printing out the connection string. What do you think?

pgcli/pgclirc
```diff
@@ -31,6 +31,12 @@ multi_line_mode = psql
 # "off" - skip all warnings
 destructive_warning = all
 
+# The list of commands when checking for a destructive query.
+# Note: "update" is based on the destructive_warning level.
+# Here are few more commands you may want to add to this list:
+# insert, create, grant
+destructive_commands = drop, shutdown, delete, truncate, alter, update
```
j-bennet, line 38: You probably could have refactored the old `destructive_warning` setting, without adding a new one. Instead of warning level, it would now explicitly contain a list of destructive commands.

To support existing users of pgcli, you could do one of the following:

1. rewrite their configs to replace:

| old                     | new                              |
|-------------------------|----------------------------------|
| destructive_warning=all | destructive_warning=drop, shutdown, delete, truncate, alter, update |
| destructive_warning=moderate | destructive_warning=drop, shutdown, delete, truncate, alter |
| destructive_warning=off | destructive_warning=off          |

2. or, detect old values (`all`, `moderate` or `off`) and replace accordingly in the code.

TBH, I don't know how safe rewriting is with configobj. I remember we had an issue of merged configs losing comments, so there's that. You'd have to do some digging.

requirements-dev.txt
```diff
@@ -11,3 +11,5 @@ click==6.7
 twine==1.11.0
 wheel==0.33.6
 prompt_toolkit==3.0.5
+sshtunnel==0.4.0
+psycopg==3.0.11
```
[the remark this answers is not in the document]
j-bennet: Does psycopg2 needs to be pinned to an exact version?

j-bennet (changes requested):
Thank you for the PR! I'm not very fond of 2 separate settings for destructive warnings, and I'd rather not add `env_name` just to display it in destructive warning. Made some suggestions - see what you think.

pgcli/main.py
rigoneri: We use ssh-tunnels to connect to all of our environments which makes the connection string usually pointing to localhost at some port. Because of that it's hard to tell which environment we are in and creates a risk of applying a destructive command to an environment we didn't intent to. I know every organization does things differently, but I've seen this problem in different organizations and it would be pretty simple solution to prevent it. Would instead of `env_name` calling it a `alias` or `connection_alias` be better?

pgcli/pgclirc
rigoneri: Reusing it makes sense. I'll get that changed.

requirements-dev.txt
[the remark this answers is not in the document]
rigoneri: I don't think it does. I can change it to point to >= 3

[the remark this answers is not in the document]
dvarrazzo: >= 3.0.11 is better, as 3.0.10 and 3.0.11 have released fixes for bugs raised by pgcli.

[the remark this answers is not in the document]
j-bennet: Wait, this shouldn't be needed at all, because the [porting PR](https://github.com/dbcli/pgcli/pull/1324) is still unmerged. If we have `psycopg` referenced in tests somewhere, this is an error. Main still has `psycopg2`.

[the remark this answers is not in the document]
rigoneri: Sorry for the confusion. I'll remove it. It was because of `pgspecial`. It wouldn't install `pgcli` as editable because of `pscopg2` so I had to manually install it `pgspecial` which brought the dependency for `psycopg`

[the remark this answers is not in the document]
j-bennet: Yes, please remove both `psycopg` and `sshtunnel` from `requirements-dev.txt`. When you install `pgcli`, it installs `psycopg2`. As for `sshtunnel`, instead of adding it to dev reqs, we need to update the contributing instructions to tell developers to install things the same way it's done in CI:

https://github.com/dbcli/pgcli/blob/366c01bbb97fee96a8169629854af7c83d626c99/.github/workflows/ci.yml#L40-L43

Also, the version of `pgspecial` that's currently in `main` branch is already ported to `psycopg`. So you can't use it with `main` branch of pgcli. You need the last version of `pgspecial that still uses `psycopg2`, that's 1.13.1.

pgcli/main.py
j-bennet: Are you using dsn aliases?

https://github.com/dbcli/pgcli/blob/366c01bbb97fee96a8169629854af7c83d626c99/pgcli/pgclirc#L200

That might be what you're looking for. You can even display them in prompt:

https://github.com/dbcli/pgcli/blob/366c01bbb97fee96a8169629854af7c83d626c99/pgcli/pgclirc#L141

That way, there's always a reminder of what you're connected to.

rigoneri: Omg! That was exactly what I was looking for :) Thank you! For the record, I had no idea that was an option. Why is it called DSN? I stayed away from it because I didn't know. I would suggest naming something easier for people to find and understand. 
I would still think it would be useful to have the destructive warning say which "alias" it is, do you think that's something I could include as part of this PR? if the alias `-D` option is passed?

requirements-dev.txt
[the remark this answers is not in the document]
rigoneri: I've removed the changes form the `requirements-dev.txt`. But I did have to add `sshtunnel` as a `deps` in `tox.ini` for Tox to run.

pgcli/main.py
rigoneri: Updated the PR let me know what you think. Thank you!

j-bennet: > Omg! That was exactly what I was looking for :) Thank you! For the record, I had no idea that was an option. Why is it called DSN? I stayed away from it because I didn't know. I would suggest naming something easier for people to find and understand.

Oh man, where to start. DSN is an old and pretty widely used term in the database world, it's been around probably since ODBC and JDBC drivers. Could call it "connection string". Maybe the original PR author thought that `connection_string_alias` is too long for a name. The `alias_dsn` section in config could use better comments. Feel free to add some.

> I would still think it would be useful to have the destructive warning say which "alias" it is, do you think that's something I could include as part of this PR? if the alias -D option is passed?

Yes, I believe that would be useful.

pgcli/packages/parseutils/__init__.py
```diff
@@ -1,34 +1,37 @@
...
-
-
-def query_is_simple_update(formatted_sql):
-    """Check if the query starts with UPDATE."""
-    tokens = formatted_sql.split()
-    return bool(tokens) and tokens[0] == "update"
-
-
-def is_destructive(queries, warning_level="all"):
+def is_destructive(queries, keywords):
```
j-bennet, line 14: Looks like we're losing a valuable behavior here. User can no longer differentiate between filtered update and unconditional update.

I see a couple ways to resolve this. We could have a special value in the list of keywords (for example, `unconditional_update`, that would warn on updates but only unconditional ones. That would have to be clearly commented in `pgclirc`.

Another way to solve this, if the user-provided list of destructive keywords is not empty, but does NOT contain keyword `update`, we could still force the warning on unconditional updates.

What do you think?

pgcli/pgclirc
```diff
@@ -22,14 +22,11 @@ multi_line = False
...
+# Destructive warning will alert you before executing a sql statement
 # that may cause harm to the database such as "drop table", "drop database",
-# "shutdown", "delete", or "update".
-# Possible values:
-# "all" - warn on data definition statements, server actions such as SHUTDOWN, DELETE or UPDATE
-# "moderate" - skip warning on UPDATE statements, except for unconditional updates
-# "off" - skip all warnings
-destructive_warning = all
+# "shutdown", "delete", or "update". You can pass a list of destructive 
+# commands or "off" if you want to skip all warnings.
```
j-bennet, line 28: Let's not tell people to use `off`, this is an "old style" value now. They can leave the setting empty. I.e `destructive_warning =`.

pgcli/pgclirc
```diff
@@ -135,6 +132,7 @@ less_chatty = False
 # \H - Hostname of the server
 # \d - Database name
 # \p - Database port
+# \e - Environment name
```
j-bennet, line 135: Not going to add environment name anymore, right?

j-bennet (commented):
I think we should preserve special treatment for unconditional updates. Added a few other comments.

pgcli/main.py
rigoneri: Ah, that makes sense. Thank you for explaining. I'll add some comments.

pgcli/pgclirc
rigoneri: Sounds good. The reason I thought initially about keeping `off` was for the `--warn` option that overrides the config. But I guess one could always pass `--warn off` since it's backwards compatible.

pgcli/pgclirc
rigoneri: Missed that, I'll remove it.

pgcli/packages/parseutils/__init__.py
rigoneri: Good point. I think `unconditional_update` makes sense.

tests/parseutils/test_parseutils.py
```diff
@@ -263,18 +268,42 @@ def test_is_open_quote__open(sql):
...
-        ("drop abc", "all", True),
-        ("alter abc", "all", True),
-        ("delete abc", "all", True),
-        ("truncate abc", "all", True),
+        ("true", ALL_KEYWORDS),
+        ("false", []),
+        ("all", ALL_KEYWORDS),
+        ("moderate", BASE_KEYWORDS),
+        ("off", []),
+        ("", []),
```
j-bennet, line 300: What about None?

j-bennet (commented):
This is looking good. I think that `parse_destructive_warning` should also handle None, just in case.

tests/parseutils/test_parseutils.py
rigoneri: Updated to handle None

j-bennet (approved):
Looks good. Thank you for the PR!

Issue #1326, opened 2022-03-22
stderr swallowed when using visidata as pager


kbd:
## Description

When using visidata as pager, stderr is swallowed.

There are a few tickets about similar issues (#1200, #1275) but after reading those tickets I don't know that there is a solution that works.

Interestingly, `litecli` works properly! Here are my configs and some examples:

### pgcli config:

```ini
[main]
pager = kw /usr/local/bin/vd -f csv
table_format = csv
less_chatty = True
prompt = '> '
```

### litecli config:

```ini
[main]
multi_line = False
pager = kw /usr/local/bin/vd
table_format = tsv
less_chatty = True
```

`kw` is "kitty window" that opens the output in a split, as you can see in the screenshots:

Visidata works as pager:

![Screen Shot 2022-03-21 at 4 17 18 PM](https://user-images.githubusercontent.com/728060/159375406-4a86e73a-81c7-4317-bb79-0abcade6ce6d.png)

But stderr swallowed:

![Screen Shot 2022-03-21 at 4 17 28 PM](https://user-images.githubusercontent.com/728060/159375575-d4c6fc7b-91b8-4287-bab7-b5a76fb54a00.png)

Visidata works as pager in litecli:

![Screen Shot 2022-03-21 at 4 51 13 PM](https://user-images.githubusercontent.com/728060/159376036-e4b1d5ab-da88-49c3-82af-6643f07f060f.png)

But stderr is not swallowed:

![Screen Shot 2022-03-21 at 4 51 55 PM](https://user-images.githubusercontent.com/728060/159376043-fb3fa656-cb86-449e-9cf9-27d6daa5bcfe.png)

For some reason using Visidata only works as a pager in pgcli with `csv` format and not `tsv` (like I have set with litecli). If set to tsv, the pager seems to be ignored.

Running on Mac Big Sur 11.6.4 and pgcli version Version: 3.4.1.

amjith:
@lelit I think you ran into a similar issue earlier. Were you able to fix the issue?

amjith:
I'm able to reproduce the issue. I'll take a deeper look into why this is happening. Sorry about the issue.

amjith:
Here's the pending PR: https://github.com/dbcli/pgcli/pull/1341

You can install this locally with this command: 

```
pip install -U https://github.com/dbcli/pgcli/archive/refs/heads/fix-csv-status.zip
```

Can you give that a try? 


[...]

PR #1173, opened 2020-04-18
Add support for using pspg as the pager.

pmav99:
addresses #1102

## Checklist

- [x] I've added this contribution to the `changelog.rst`.
- [x] I've added my name to the `AUTHORS` file (or it's already there).

- [x] I installed pre-commit hooks (`pip install pre-commit && pre-commit install`), and ran `black` on my code.
- [x] Please squash merge this pull request (uncheck if you'd like us to merge as multiple commits)

PR #1341, opened 2022-05-23
Show status in csv format.

amjith:
## Description
Show error status in CSV output format.

## Checklist

- [X] I've added this contribution to the `changelog.rst`.


j-bennet:
@amjith I thought we aren't displaying status in CSV [intentionally](https://github.com/dbcli/pgcli/pull/1173), because it was breaking `pspg` pager. Can you explain why we need this now?

changelog.rst
```diff
@@ -4,7 +4,9 @@
 Bug fixes:
 ----------
 
-* Fix the bug with Redshift not displaying word count in status ([related issue](https://github.com/dbcli/pgcli/issues/1320)).
+* Fix the bug with Redshift not displaying wor count in status ([related issue](https://github.com/dbcli/pgcli/issues/1320)).
```
lelit, line 7: s/wor/word/?

amjith:
I believe [pspg](https://github.com/okbob/pspg) implemented a special flag to handle pgcli. `--pgcli-fix`

lelit:
> So what happens now with `pspg`

FWIW, I'm happy with minimal configuration suggested at the end of the [pgcli](https://github.com/okbob/pspg#pgcli) section:
```
pager = /usr/bin/pspg --rr=2 --quit-if-one-screen --ignore-case
```


> I believe pspg implemented a special flag to handle pgcli: `--pgcli-fix`

Yes, but AFAICT that targets problems when using `CSV` output. Using the recommended `table_format = psql` it's not needed.

j-bennet:
@amjith I noticed that in this comment:

https://github.com/dbcli/pgcli/issues/1326#issuecomment-1162472608

the author says that even with this cnange, `pgcli` still doesn't behave the right way when displaying stderr, while litecli does. Do you know what is different?

Issue #1342, opened 2022-05-24
The \e feature to jump to an editor does not work with gvim. (graphical vim)


bossstein:
## Description
The \e feature to jump to an editor does not work with gvim. When I use it with other editors the text you have written so far goes into the editor and you can edit then after a save and quit you get that same text back in pgcli. With gvim the editor always opens empty and after save and quit nothing is returned to pgcli.

It selects gvim because that is what I have selected under $VISUAL environment variable. I would be even more happy to use vim in terminal, which does work under the $EDITOR environment variable but it only defaults to this if $VISUAL is empty. There is no option to chose an alternate editor as far as I can see.

As such I think there are some sub-tasks here:

1. Fix gvim compatibility.
2. Add config option to use the non-visual editor option.
3. And/Or add config option to set alternate editor.
4. The docs on the website are inconsistent. Someplace says only $EDITOR is used which is incorrect.

I am happy to take a punt at this issue myself but I may not have time this week in which case it will be three weeks at the earliest.

## Your environment
Ubuntu 20.04.4 LTS
pgcli version 3.4.1
vim version 8.1.3741

[pip-freeze.txt](https://github.com/dbcli/pgcli/files/8764402/pip-freeze.txt)

bossstein:
I found that my issue was that gvim forks when it's run from the terminal. There is an option to run --nofork . This isn't ideal but I can, as a work around, set my VISUAL variable to gvim --nofork. 

I found that the functionality here uses the pgspecial library that in turn uses the click library. The click library supports specifying an editor. I have therefore made a tiny pull request to the pgspecial library to expose that feature. 

Once the pgspecial library supports this I have the change ready to add a config option for a non default editor to pgcli.

PR #1350, opened 2022-06-18
 Add alternate_editor config option to specify an editor.

bossstein:
## Description
Solution to issue #1342 . This change adds an `alternate_editor` config option. Previous functionality was to check the $VISUAL environment variable and use this and then if this is blank it would use $EDITOR. These environment variables are used by other software and as such a user may not want to change them, however they may want to not use a visual editor in pgcli or they may want to use a different editor to the default. By adding this alternate editor option a user can choose a different editor and even specify arguments if they want to use different editor behaviour while in pgcli.

## Checklist

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -122,6 +122,7 @@ Contributors:
     * Daniele Varrazzo
     * Daniel Kukula (dkuku)
     * Kian-Meng Ang (kianmeng)
+    * Max Steinbach (bossstein)
 
 Creator:
 --------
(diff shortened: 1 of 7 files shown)


bossstein:
@j-bennet can you check this pr when it's convenient.

j-bennet (commented):
Looks fine, except I'd leave the config option empty, instead of `default`.

j-bennet:
@bossstein Do you plan to continue working on this PR?

PR #1352, opened 2022-06-27
Add support for overriding DSN connection information

A-23187:
## Description
When connecting using a DSN alias, you can specify additional options (including  `-h`, `-p`, `-u`, `-W` and `-d`) to override the connection information represented by the DSN you used. For example, `pgcli -D dsn -p 54321` means to connect using the alias `dsn`, while the port is overridden as `54321`.  It is very useful for me because I have multiple PostgreSQL services deployed on a single server that listen on different ports, I just need to configure only one DSN alias and specify different ports via the `-p` option to connect to these services.

## Checklist


pgcli/main.py
```diff
@@ -1385,7 +1387,21 @@ def cli(
...
+            "port": port,
+            "username_opt": username_opt,
+            "dbname_opt": dbname_opt,
+        }
+        remap = {"username_opt": "user", "dbname_opt": "dbname"}
+        kwargs = {
+            remap.get(k, k): v
+            for k, v in kwargs.items()
+            if ctx.get_parameter_source(k) is click.core.ParameterSource.COMMANDLINE
+        }
```
laixintao, line 1401: can you add testcases for this part? thanks

laixintao (commented):
I prefer that you set multiple DSN, this feature is kind of hard to explain.

But if @j-bennet or @amjith is ok then I am ok with it.

Can you document this behavior and add some tests?

pgcli/main.py
A-23187: Specifically, on a test machine, we have many instances of the PostgreSQL service running on random ports (when creating an instance, our administration system randomly selects an available port for that instance), so I can't easily connect to each instance by setting multiple DSN. I hope that when using a DSN whose IP address is set to our test machine to connect, I can specify the port obtained from our administration system with the `-p` option. But currently pgcli ignore my `-p` option when I use a DSN to connect, even if this DSN does not explicitly specify a port (if no port is specified, pgcli will use the default port 5432 to connect)

A-23187:
Specifically, on a test machine, we have many instances of the PostgreSQL service running on random ports (when creating an instance, our administration system randomly selects an available port for that instance), so I can't easily connect to each instance by setting multiple DSN. I hope that when using a DSN whose IP address is set to our test machine to connect, I can specify the port obtained from our administration system with the -p option. But currently pgcli ignore my -p option when I use a DSN to connect, even if this DSN does not explicitly specify a port (if no port is specified, pgcli will use the default port 5432 to connect)

j-bennet:
> our administration system randomly selects an available port for that instance), so I can't easily connect to each instance by setting multiple DSN

@laixintao @A-23187 I agree, this sounds like a valid use case. Are you able to add a test for this, @A-23187 ?

pgcli/main.py
A-23187: Well, I add a script to test this part, I mock `PGCli.connect_uri` to verify that whether using `-D dsn` along with `-h` `-p` `-u` `-d` `-W` can override the connection information of `dsn`. But I'm not familiar with pytest, so let me know if there are any problems in my test script

A-23187:
> > our administration system randomly selects an available port for that instance), so I can't easily connect to each instance by setting multiple DSN
> 
> @laixintao @A-23187 I agree, this sounds like a valid use case. Are you able to add a test for this, @A-23187 ?

Sure, I added a script to test this.

PR #1354, opened 2022-07-06
Autocommit

vanviegen:
## Description

- Toggle autocommit mode using F6 or a config setting.
- When in a transaction, rollback erring queries without aborting the transaction. ('On error rollback'.)

Closes #410. This work is loosely based on PR #917 by Matthieu Guilbert, but attempts to keep things simpler. I'm no PostgreSQL expert, so reviewers are encouraged to look for edge cases that make this simpler approach fall apart.

## Checklist
- [x] I've added this contribution to the `changelog.rst`.
- [x] I've added my name to the `AUTHORS` file (or it's already there).
- [x] I installed pre-commit hooks (`pip install pre-commit && pre-commit install`), and ran `black` on my code.
      --> I'm getting `(no files to check) Skipped`.
- [x] Please squash merge this pull request (uncheck if you'd like us to merge as multiple commits)

PR #1355, opened 2022-07-07
More compact bottom bar

vanviegen:
## Description

The multiline and transaction parts of the bottom bar could be rather verbose, causing (import) info not to fit when using a somewhat narrow terminal. This PR removes some verbosity, while hopefully maintaining clarity.

This PR is based on PR #1354 as it touches some of the same code.

## Checklist

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -122,6 +122,7 @@ Contributors:
     * Daniele Varrazzo
     * Daniel Kukula (dkuku)
     * Kian-Meng Ang (kianmeng)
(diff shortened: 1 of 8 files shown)


j-bennet:
Hi @vanviegen ,

I see your PR's tests are failing on `test_virtual_database`. We have some special handling for `pgbouncer` database, and I suspect transactions wouldn't be supported in that, so you might need to use that check for `self.is_virtual_database()`. Do you need help with this PR?

PR #1375, opened 2022-10-26
magic: use PGCli.connect_uri instead of connect

aglasgall:
This makes %pgcli work even if you use non-password (e.g. TLS cert) authentication

## Description

This change changes `pgcli.magic` to use `PGCli.connect_uri` instead of `PGCli.connect`, to make it work with PG connection urls that use TLS certs for authentication instead of username/password.

I looked into passing the underlying pg connection from sqlalchemy into pgcli instead (it's the `connection` method on `conn.session` to get a dbapi-connection-alike, or `conn.session.connection.dbapi_connection` to get the real underlying connection), but pgcli is using psycopg3 and sqlalchemy is still using psycopg2. So that won't be as easy without writing some sort of compatibility shim (sqlalchemy supports psycopg3 in 2.0, which is still in beta, so this problem will eventually go away)

## Checklist

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -124,6 +124,7 @@ Contributors:
     * Kian-Meng Ang (kianmeng)
     * Liu Zhao (astroshot)
     * Rigo Neri (rigoneri)
+    * Anna Glasgall (annathyst)
 
 Creator:
 --------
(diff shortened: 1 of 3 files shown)


j-bennet:
@aglasgall Nice work - thank you for the PR.

PR #1379, opened 2022-11-06
Do not automatically restart connection on destructive warning

andyscho:
## Description
Before this change, cancelling a query with a destructive warning would automatically restart the connection. This can be unintuitive behavior for users that might expect only their destructive warning statement to be cancelled. For instance, a user in the middle of a transaction could have their entire transaction aborted because they cancelled a destructive warning statement, leading to previous non-destructive inserts being lost. 

With the new default for this config option (not automatically restarting the whole connection), the past statements within the transaction will not be lost, and the user may continue executing more statements within the same transaction.

This change adds a config option for whether or not a destructive warning will restart the connection, and defaults it to False. It only affects statements issued directly in the cli, and not destructive warnings from sql statements files.

## Checklist

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -10,6 +10,8 @@ Features:
 * pgcli.magic will now work with connection URLs that use TLS client certificates for authentication
 * Have config option to retry queries on operational errors like connections being lost.
   Also prevents getting stuck in a retry loop.
+* Config option to not restart connection when cancelling a `destructive_warning` query. By default,
+  it will now not restart.
 
 3.5.0 (2022/09/15):
 ===================
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -231,6 +231,9 @@ class PGCli:
         self.destructive_warning = parse_destructive_warning(
             warn or c["main"].as_list("destructive_warning")
         )
+        self.destructive_warning_restarts_connection = c["main"].as_bool(
+            "destructive_warning_restarts_connection"
+        )
 
         self.less_chatty = bool(less_chatty) or c["main"].as_bool("less_chatty")
         self.null_string = c["main"].get("null_string", "<null>")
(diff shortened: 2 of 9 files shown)


j-bennet:
@andyscho Thank you for the PR! I have a few questions related to the problem you're trying to solve:

1. What are the downsides of the connection being recreated? How does it affect the user experience?
2. Are there any downsides to not re-creating the connection? What are they?
3. How did you test this use case? Is it feasible to add a unit and / or an integration test for that?

j-bennet:
@andyscho You'll have to resolve merge conflicts on this one.

andyscho:
@j-bennet Thanks for the response!

> 1. What are the downsides of the connection being recreated? How does it affect the user experience?

One key noticeable downside is the dropping of current transactions. For example, before this PR, we have the following behavior:

```
postgres@/tmp:postgres> begin
BEGIN
Time: 0.001s
postgres@/tmp:postgres> insert into test (username, value) VALUES ('username1', 50);
INSERT 0 1
Time: 0.002s
postgres@/tmp:postgres> alter table test add column field2 VARCHAR(50)
You're about to run a destructive command.
Do you want to proceed? (y/n): n
Wise choice!
cancelled query
postgres@/tmp:postgres> select * from test;
+----------+-------+
| username | value |
|----------+-------|
+----------+-------+
SELECT 0
Time: 0.015s
postgres@/tmp:postgres> rollback
there is no transaction in progress
None

ROLLBACK
```


This is unintuitive behavior as it results in previous SQL statements also being cancelled in addition to the current one that is executing. In my opinion, normal SIGINT keyboard interrupts should not break out of transactions automatically. This is the case in the psql CLI, and seems like a good decision to me; users should specify behavior related to transactions. For instance, just wanting to cancel the current query I’m currently typing doesn’t seem like grounds to have all my previous statements within a transaction discarded.

> 2. Are there any downsides to not re-creating the connection? What are they?

As far as I can tell, the original behavior [entered the pgcli codebase](https://github.com/dbcli/pgcli/pull/92/files) due to some limitations in the original handling of keyboard interrupts. [Psycopg3 seems to be able to handle keyboard interrupts differently now (at least for > 3.0.14, the current minimum version required by pgcli)](https://github.com/psycopg/psycopg/issues/231), so it should no longer be necessary to restart the connection when handling a keyboard interrupt.

As far as I can tell, there aren’t any other downsides. This new behavior seems to better match existing psql behavior expectations. I left it as a config option for anyone that does like the old behavior or has some other need for it, but I think a default of not restarting is reasonable for most users.

> 3. How did you test this use case? Is it feasible to add a unit and / or an integration test for that?

I have added a behave test scenario demonstrating that a transaction is not automatically rolled back when rejecting a destructive warning.

I also added a test to make sure a query is actually cancelled when pressing Ctrl+c, just to make sure that behavior is maintained.

j-bennet:
This looks perfect! Thank you for adding tests as well. If it was me, I'd remove the `destructive_warning_restarts_connection` config option, because your new default behavior is less disruptive than the old behavior, and I can't see why anybody would want the old one back. But I don't feel strongly about it, so going to merge as-is.

Thank you! 🍓

PR #1380, opened 2022-11-07
Add config option to not automatically retry connections with operational errors

andyscho:
## Description
This change:
1. Adds a config option (defaulting to auto-retrying) for whether or not to retry a query after it has failed with an OperationalError.
2. Will only retry at most once on a connection that failed with an OperationalError. This will still be pretty usable and non-invasive in scenarios where we should gracefully reconnect and run a query, as in the scenario outlined in the PR introducing auto-retrying (#1009).

It is not necessarily good to always reconnect and retry queries that fail with an OperationalError. For instance, it might be desirable for a database admin to be able to kill a long-running resource-hogging query that a user left running unattended. Before this change, if the misbehaving query was killed with pg_terminate_backend, it would just be automatically restarted. Even worse, before this change, the query could find itself in a reconnect-and-retry loop since every time it's killed with pg_terminate_backend, it will just start up again automatically. Since the database admin might not necessarily even be aware of this as a possibility, it might be difficult to debug.

## Checklist

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -125,6 +125,7 @@ Contributors:
     * Liu Zhao (astroshot)
     * Rigo Neri (rigoneri)
     * Anna Glasgall (annathyst)
+    * Andy Schoenberger (andyscho)
 
 Creator:
 --------
--- a/changelog.rst
+++ b/changelog.rst
@@ -8,6 +8,8 @@ Features:
   destructive. This would allow you to be warned on `create`, `grant`, or `insert` queries.
 * Destructive warnings will now include the alias dsn connection string name if provided (-D option).
 * pgcli.magic will now work with connection URLs that use TLS client certificates for authentication
+* Have config option to retry queries on operational errors like connections being lost.
+  Also prevents getting stuck in a retry loop.
 
 3.5.0 (2022/09/15):
 ===================
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -63,7 +63,7 @@ from .config import (
 )
 from .key_bindings import pgcli_bindings
 from .packages.formatter.sqlformatter import register_new_formatter
-from .packages.prompt_utils import confirm_destructive_query
+from .packages.prompt_utils import confirm, confirm_destructive_query
 from .packages.parseutils import parse_destructive_warning
 from .__init__ import __version__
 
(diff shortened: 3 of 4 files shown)


j-bennet:
@andyscho Is there a way to determine if the OperationalError happened because the query was forcibly terminated with `pg_erminate_backend`? a more specific error code? an inner excepton?

I'm wondering if we can be a little smarter here, and not just flip between "always reconnect" and "never reconnect". It feels like automatically reconnecting would be a good choice in most cases, and only rarely user might want to suppress that.

andyscho:
@j-bennet Unfortunately, I don't believe psycopg passes along more specific information than an OperationalError with an error message string. Although there is an Error type for e.g. [AdminShutdown](https://github.com/psycopg/psycopg/blob/f12f5f37f3e623b7d4fbe7f0482970bccdcd1cc3/psycopg/psycopg/errors.py#L1326), the actual code that seems to raise when I run `pg_terminate_backend` [just calls OperationalError on a string error message](https://github.com/psycopg/psycopg/blob/f12f5f37f3e623b7d4fbe7f0482970bccdcd1cc3/psycopg/psycopg/pq/pq_ctypes.py#L525).

One option I did consider was isolating the errors that are probably safe to always retry on, but I think this might also be difficult to isolate, and I would expect it to also vary a bit by use case.

I agree that in many situations, auto-reconnecting and retrying is probably a good choice - this pull request leaves it as the default config option. However, not always automatically retrying is a good choice for some environments where a very slow query can cause major problems for other users. 

Only letting users automatically reconnect when receiving any type of OperationalError as a blanket method seems like a removal of functionality; I feel like I should be able to only try querying once if I want to, especially since I'll still be prompted to decide if I want to try again anyways. Psycopg raises an OperationalError in a variety of other use cases; I do not think it's likely that `pg_terminate_backend` is the only one that users would prefer not to always automatically retry on.

j-bennet:
@andyscho 

> Only letting users automatically retry when receiving any type of OperationalError as a blanket method seems like a removal of functionality; I feel like I should be able to have an option to specifically only try querying once if I want to, especially since I'll still be prompted to decide if I want to try again anyways

Sounds fair. Thank you for the PR. Merging.

PR #1382, opened 2022-11-11
Allow defining a json file with preferred aliases

rob-b:
## Description

At $WORK we have a lot of tables with names like `foo_noun_verb` or `foo_noun_related-noun_verb` and so while the default aliasing is very helpful for shortening unwieldy names we do end up with lots of aliases like `LEFT JOIN fnv on fnv2.id = fnv.fnv2_id`

This change will allow defining a json file of preferred aliases

```
> cat ~/.config/pgcli/aliases.json
{
    "foo_user": "user",
    "foo_user_group": "user_group"
}
```

so the alias suggestion for `SELECT * FROM foo_user` will be `SELECT * FROM foo_user AS user` instead of the default `SELECT * FROM foo_user AS fu`

## Checklist

### DIFF
--- a/pgcli/pgclirc
+++ b/pgcli/pgclirc
@@ -9,6 +9,10 @@ smart_completion = True
 # visible.)
 wider_completion_menu = False
 
+# Do not create new connections for refreshing completions; Equivalent to
+# always running with the --single-connection flag.
+always_use_single_connection = False
+
 # Multi-line mode allows breaking up the sql statements into multiple lines. If
 # this is set to True, then the end of the statements must have a semi-colon.
 # If this is set to False then sql statements can't be split into multiple
@@ -31,9 +35,13 @@ destructive_warning = drop, shutdown, delete, truncate, alter, update, unconditi
 
 # Destructive warning can restart the connection if this is enabled and the
 # user declines. This means that any current uncommitted transaction can be
-# aborted if the user doesn't want to proceed with a destructive_warning 
+# aborted if the user doesn't want to proceed with a destructive_warning
 # statement.
 destructive_warning_restarts_connection = False
+
+# When this option is on (and if `destructive_warning` is not empty),
+# destructive statements are not executed when outside of a transaction.
+destructive_statements_require_transaction = False
 
 # Enables expand mode, which is similar to `\x` in psql.
 expand = False
@@ -99,9 +107,10 @@ qualify_columns = if_more_than_one_table
 # When no schema is entered, only suggest objects in search_path
 search_path_filter = False
 
-# Default pager.
-# By default 'PAGER' environment variable is used
-# pager = less -SRXF
+# Default pager. See https://www.pgcli.com/pager for more information on settings.
+# By default 'PAGER' environment variable is used. If the pager is less, and the 'LESS'
+# environment variable is not set, then LESS='-SRXF' will be automatically set.
+# pager = less
 
 # Timing of sql statements and table rendering.
 timing = True
--- a/pgcli/pgclirc
+++ b/pgcli/pgclirc
@@ -54,7 +54,7 @@ generate_aliases = False
 #     "some_table_name": "desired_alias",
 #     "some_other_table_name": "another_alias"
 # }
-alias_map_file = ""
+alias_map_file =
 
 # log_file location.
 # In Unix/Linux: ~/.config/pgcli/log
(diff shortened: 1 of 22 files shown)


pgcli/pgcompleter.py
```diff
@@ -100,6 +102,15 @@ def __init__(self, smart_completion=True, pgspecial=None, settings=None):
         self.call_arg_oneliner_max = settings.get("call_arg_oneliner_max", 2)
         self.search_path_filter = settings.get("search_path_filter")
         self.generate_aliases = settings.get("generate_aliases")
+
+        # when should this file be loaded? IO in constructors is not my preference but slow startup is
+        # probably better than slow first query
+        alias_map_file = settings.get("alias_map_file")
+        if alias_map_file is not None:
+            with open(alias_map_file) as fo:
+                self.alias_map = json.load(fo)
```
rob-b, line 111: I wanted to flag this part as I wasn't sure of what approaches here fir the project goals

1. If the file doesn't exist what should I do? Allow the crash or just handle it silently and set the map to `None`?
2. is loading the file in the constructor a good idea? I tend not to do IO in constructors normally but since theoretically the alias file could be quite large I thought slow startup was better than a slow first query. Maintainers might think the opposite

j-bennet: 1. If someone specified a file in the config, and it doesn't exist, it should be an error. We can assume that the user wanted the file to be picked up, and so the user will be interested in fixing the path.
2. How large is "quite large"? Did you do any benchmarks with your file? If you think it's worth the effort, you could run a background thread to load these aliases, like the [completion refresher](https://github.com/dbcli/pgcli/blob/main/pgcli/completion_refresher.py) does.

rob-b: I pushed a new version that wraps loading/decoding into a helper function and raises an informative error if the map_file cannot be read


My personal alias file is tiny - 5 aliases and the whole file is ~220 bytes - and so for myself at least loading is fast

```
In [1]: from pgcli.pgcompleter import load_alias_map_file

In [2]: from pgcli import config

In [3]: import timeit

In [4]: map_file = config.get_config()["main"]["alias_map_file"]

In [5]: t = timeit.Timer(lambda: load_alias_map_file(map_file))

In [6]: t.timeit(10_000)
Out[6]: 0.15082729199999534
```

I tried upping the alias file to 1000 aliases and it's still  pretty fast

```
In [9]: t = timeit.Timer(lambda: load_alias_map_file(map_file))

In [10]: t.timeit(10_000)
Out[10]: 1.3473110840000118
```

> How large is "quite large"? 

For my own usage I can't see the file getting bigger than 200 lines (we have ~ 120 tables in prod) which is 8kb and quick to load

```
In [8]: t.timeit(10_000)
Out[8]: 0.3809067500000083
```

my caution was purely down to experience of users always doing something surprising and using features in unexpected ways.

I don't think that loading the aliases would need to be performed in a background thread

pgcli/pgclirc
```diff
@@ -42,6 +42,14 @@ auto_retry_closed_connection = True
 # If set to True, table suggestions will include a table alias
 generate_aliases = False
 
+# Path to a json file that specifies specific table aliases to use when generate_aliases is set to True
+# the format for this file should be:
+# {
+#     "some_table_name": "desired_alias",
+#     "some_other_table_name": "another_alias"
+# }
+alias_map_file = ""
```
j-bennet, line 51:
```suggestion
alias_map_file =
```

pgcli/pgcompleter.py
```diff
@@ -61,18 +62,38 @@ def Candidate(
 normalize_ref = lambda ref: ref if ref[0] == '"' else '"' + ref.lower() + '"'
 
 
-def generate_alias(tbl):
+def generate_alias(tbl, alias_map=None):
```
j-bennet, line 65: Looks like we have no tests for this functions. I'd add one, especially now that the behavior is getting more complicated.

rob-b:
@j-bennet is there a way I can get codecov to re-run? It's reporting based on the first version of this PR and I'd be interested to see if the additional tests I've written have moved the number

j-bennet:
@rob-b I actually don't know how to bump codecov to recalculate, but I wouldn't worry about coverage report, it's not that important. Can you add your name to AUTHORS file? Otherwise I think the PR is good to merge.

j-bennet:
@rob-b Sorry this took forever, it got lost from my radar. Thank you for the PR!

PR #1383, opened 2022-11-11
separate formatted status lines from normal output and send to stderr

teepark:
## Description
pgcli only makes use of standard out for output, including sending output to a pager. But there is no good way to mix formatted query output with trailing status lines ("SELECT N" &c) and error messages.

One choice is you can omit these status and error messages entirely with programmatic output (as was done before #1341) and take away users' ability to see these important messages (see #1326 - though the title is confusing because the author was under the impression stderr was already being used). Or you include them and confuse helpful pager programs like visidata that expect to only receive CSV:
invalid CSV being handled by visidata

But this is why stderr exists, and it gives us the best of both worlds: users can still see this output on the terminal, but it won't be sent into the pager program.

## Checklist

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -126,6 +126,7 @@ Contributors:
     * Rigo Neri (rigoneri)
     * Anna Glasgall (annathyst)
     * Andy Schoenberger (andyscho)
+    * teepark
 
 Creator:
 --------
--- a/changelog.rst
+++ b/changelog.rst
@@ -10,6 +10,7 @@ Features:
 * pgcli.magic will now work with connection URLs that use TLS client certificates for authentication
 * Have config option to retry queries on operational errors like connections being lost.
   Also prevents getting stuck in a retry loop.
+* Send trailing status and error lines to stderr to avoid breaking pagers expecting CSV or TSV.
 
 3.5.0 (2022/09/15):
 ===================
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -709,7 +709,7 @@ class PGCli:
                 elif destroy:
                     click.secho("Your call!")
 
-            output, query = self._evaluate_command(text)
+            output, status, query = self._evaluate_command(text)
         except KeyboardInterrupt:
             # Restart connection to the database
             self.pgexecute.connect()
@@ -742,6 +742,8 @@ class PGCli:
                 else:
                     if output:
                         self.echo_via_pager("\n".join(output))
+                    if status:
+                        self.echo_to_stderr("\n".join(status))
             except KeyboardInterrupt:
                 pass
 
@@ -949,7 +951,7 @@ class PGCli:
         """Used to run a command entered by the user during CLI operation
         (Puts the E in REPL)
 
-        returns (results, MetaQuery)
+        returns (results, status, MetaQuery)
         """
         logger = self.logger
         logger.debug("sql: %r", text)
@@ -962,6 +964,7 @@ class PGCli:
         db_changed = False
         path_changed = False
         output = []
+        status_output = []
         total = 0
         execution = 0
 
@@ -1008,11 +1011,12 @@ class PGCli:
                 max_field_width=self.max_field_width,
             )
             execution = time() - start
-            formatted = format_output(
+            formatted, formatted_status = format_output(
                 title, cur, headers, status, settings, self.explain_mode
             )
 
             output.extend(formatted)
+            status_output.extend(formatted_status)
             total = time() - start
 
             # Keep track of whether any of the queries are mutating or changing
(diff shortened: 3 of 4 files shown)


j-bennet:
@teepark I see two problems with this PR.

1. Yes, `pgcli` should be making use of stderr, and it's a bug that it doesn't. At the same time, `stdout` should be used for normal output, and `stderr` for error output. `SELECT 1` status is normal output, and when something breaks, the error should be sent into `stdout`. With this PR, OK status will be sent to `stderr`, and it doesn't feel like the right thing to do.

I think the right thing would be:

1. `pgcli` should track whether query (queries) succeeded or not (we already have that here)
2. normal status lines should be sent to `stdout`, except in case of CSV format we should drop them
3. error status lines should be sent to `stderr`

2. This doesn't handle multiple queries well:

``
test> select 1; select 2
+----------+
| ?column? |
|----------|
| 1        |
+----------+
+----------+
| ?column? |
|----------|
| 2        |
+----------+
SELECT 1
SELECT 1
Time: 0.019s
```

What are your thoughts?

**Comment by teepark** [comment]
Yep, I can definitely address both of these points.

I disagree with your first conclusion, but not particularly strongly. My opinion is based on two things:

1. In my experience "errors" is too narrow of a definition for normal uses of `stderr` (despite the name, I know 🙃). I see it used for all kinds of *helpful human-oriented output*, errors or otherwise. So situations like "downstream tools expect CSV/TSV/JSON/XML/whatever but I want to put a diagnostic message in front of the user" are solid use-cases. And all other things being equal, it's probably better to provide the `SELECT N` output than not. Litecli provides precedent too:

    success output going to stderr in litecli

2. It's more broad than just this change, but I think it's a misstep to make additional output decisions based on the chosen format. Maybe someone has already built a parser for the `psql` format before they switched to using pgcli, or the `ascii` one because they were previously using mysql. They'll be getting the human-consumption set of decisions just for not using CSV specifically. The larger point is that `format == 'csv'` is a poor proxy for "user wants machine-oriented output". I've already been bitten by this because I'd prefer to use visidata with TSV coming from pgcli, but the magical *always use the pager* behavior is only activated for CSV - with any other format the pager is only activated by very long or wide output.

But all that said, I'm also content to just implement your logic. It still solves the problem I've encountered and I doubt anyone would notice the differences in our proposals. So just let me know if I've convinced you 😁.

Your second requirement is interesting because we need to preserve order even between stdout-bound and stderr-bound lines. My first instinct is that the current two lists becomes a list of 2-tuples `(line:str, is_stdout:bool)`, and line-by-line we send the output to the correct destination. But that's just my personal style to use direct data structures like tuples, some would probably rather have an `OutputLine` class with a `line` string and `destination` enum. I want to match the project coding style though, so I'm happy to take direction here.

**Comment by j-bennet** [comment]
I looked at `litecli` code to figure out what's different. I was surprised. Just like `pgcli`, it uses stdout, not stderr to output status:

https://github.com/dbcli/litecli/blob/536e29dcb53062184da72c4e93541192ad26ca7d/litecli/main.py#L681

There is a difference though. `Litecli` iterates over multiple query results, and for each query, it calls `[echo_via_pager(rows)](https://github.com/dbcli/litecli/blob/536e29dcb53062184da72c4e93541192ad26ca7d/litecli/main.py#L674)`, and then `[echo(status)](https://github.com/dbcli/litecli/blob/536e29dcb53062184da72c4e93541192ad26ca7d/litecli/main.py#L681)`. Something like this (pseudocode):

```
for cursor in query_results:
   output, status = format_output(output_args)
   click.echo_via_pager(output)
   click.echo(status)
```

While `pgcli` collects output from all queries [into one large list](https://github.com/dbcli/pgcli/blob/fa054a5546d33ed34586513f71f371f81ef6ecfd/pgcli/main.py#L1015), and calls `echo_via_pager` on the whole thing. Something like this:

```
complete_output = []
for cursor in query_results:
    output, status = format_output(output_args) 
    complete_output.extend(output)
    complete_output.append(status)
click.echo_via_pager(complete_output)
```

So, with `litecli`, `visidata` pager is invoked several times, each time with a single table. It has no problem handling one table at a time. And `secho` on a single status line doesn't trigger the pager at all.

With `pgcli`, `visidata` pager is invoked once, with concatenated table - status line - table - status line. It can't handle that very well.

And `stderr` is not even part of the equation.

The question is, does `litecli` do the right thing? I believe so. I think `pgcli` should do what `litecli` does. It might turn into a pretty big refactor though.

j-bennet:
@teepark

If you feel up to it, I would suggest you rework this PR to handle output similarly to `litecli`, i.e. separate calls to `echo_via_pager` per table/query. No need to complicate things by switching between stdout and stderr. I expect it might break a lot of tests, including integration tests, but who knows, I might be wrong and it won't be too bad.

teepark:
Hi, sorry for my radio silence. I took a stab at this before Thanksgiving but ran into some problems. I do intend to get back to this though, so please don't close due to inactivity :)

teepark:
actually I did away with this code so I'll add a separate PR for the new approach.

Issue #1372, opened 2022-10-16
Wrong quotation in \ev


Jakubzik:
## Description
`\ev schema.my_view` produces `CREATE OR REPLACE VIEW "schema.my_view"`. It *should* produce `... VIEW "schema"."my_view"`, I think (with quotes enclosing "schema" and "my_view" separately).

The effect or this had me confused for quite a while, as the `CREATE OR REPLACE ...` statement (with the wrong quotation marks) produces no exception, but of course it also does not work as expected (but creates a *new* view instead).

## Your environment

- Arch Linux
- pgcli 3.5.0
- pip freeze: anytree==2.8.0
appdirs==1.4.4
apsw==3.38.5.post1
argcomplete==2.0.0
async-timeout==4.0.2
autocommand==2.2.1
backports.csv==1.0.7
beautifulsoup4==4.11.1
Brotli==1.0.9
brotlicffi==1.0.9.2
btrfsutil==6.0
CacheControl==0.12.11
cchardet==2.1.7
certifi==2022.9.24
cffi==1.15.1
chardet==5.0.0
cli-helpers==2.2.1
click==8.1.3
colorama==0.4.5
commonmark==0.9.1
configobj==5.1.0.dev0
contextlib2==21.6.0
cryptography==38.0.1
css-parser==1.0.7
cssselect==1.1.0
cupshelpers==1.0
Cython==0.29.32
distlib==0.3.6
distro==1.7.0
dnspython==2.2.1
docutils==0.19
entrypoints==0.4
feedparser==6.0.10
filelock==3.8.0
future==0.18.2
gpg==1.18.0
html2text==2020.1.16
html5-parser==0.4.10
html5lib==1.1
humanize==3.1.0
idna==3.4
ifaddr==0.2.0
imaplib2==3.6
importlib-metadata==5.0.0
inflate64==0.1.3
inflect==6.0.0
jaraco.classes==3.2.3
jaraco.context==4.1.2
jaraco.functools==3.5.2
jaraco.text==3.9.1
jeepney==0.8.0
Jinja2==3.1.2
keyring==23.9.3
lensfun==0.3.3
lockfile==0.12.2
louis==3.23.0
lxml==4.9.1
Markdown==3.4.1
MarkupSafe==2.1.1
mechanize==0.4.8
meson==0.63.3
more-itertools==8.14.0
msgpack==1.0.4
multivolumefile==0.2.3
netifaces==0.11.0
netsnmp-python==1.0a1
[...]

PR #1384, opened 2022-11-20
fix psycopg.sql.Identifier in \ev handling

andyscho:
## Description
This fixes the issue outlined in #1372 by changing the [psycopg.sql.Identifier](https://www.psycopg.org/psycopg3/docs/api/sql.html#psycopg.sql.Identifier). I would guess this problem originally entered the codebase relatively recently during the port to psycopg3.

Before this PR, `CREATE OR REPLACE VIEW` will not replace the view, and will instead create a new view with the name `<schema>.<view>`, leaving the original view unchanged.

Adds extremely small test assertion to make sure the view is quoted correctly.

## Checklist

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -12,6 +12,7 @@ Features:
   Also prevents getting stuck in a retry loop.
 * Config option to not restart connection when cancelling a `destructive_warning` query. By default,
   it will now not restart.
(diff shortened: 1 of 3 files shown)


j-bennet:
Nice work @andyscho .

PR #1386, opened 2022-11-26
Add config option to always run with a single connection

andyscho:
## Description
Add a config option for always running with a single connection. This is equivalent to always running with the `--single-connection` flag.

This is a very minor change, but this flag seems like it's the type of thing that would be useful to have persist as a regular config setting; for people that use it, it can be annoying to always have to specify it.

Leaving it defaulting to False to have the same behavior as before this change (not specifying the flag will not run in single-connection mode). When the config setting is False, users can still enable it in a non-persistent way by running with the flag in the same way as before.

## Checklist

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -13,6 +13,7 @@ Features:
 * Config option to not restart connection when cancelling a `destructive_warning` query. By default,
   it will now not restart.
 * Fix \ev not producing a correctly quoted "schema"."view"
+* Config option to always run with a single connection.
 
 3.5.0 (2022/09/15):
 ===================
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -264,6 +264,9 @@ class PGCli:
         # Initialize completer
         smart_completion = c["main"].as_bool("smart_completion")
         keyword_casing = c["main"]["keyword_casing"]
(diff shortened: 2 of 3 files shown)


andyscho:
@j-bennet does this seem like a reasonable config addition?

j-bennet:
@andyscho I'm ok with this option. Can you provide some use cases where it's necessary?

andyscho:
> Can you provide some use cases where it's necessary?

The main use case I had in mind when I made this pull request was a situation where the number of simultaneous connections a user can have to a database are limited. In this situation, it's probably helpful to basically always run it in `single_connection` mode (assuming there aren't any other issues with running in `single_connection` mode). Since it would be helpful to be able to enforce that the setting is always enabled, I think it seems reasonable to include it as a persistent config setting.

The `single_connection` option appears to have been introduced in #557, with the intention of the completer being able to function with one-time passwords when connecting to a database. I'm not particularly familiar with this specific use case, but it seems like it could be useful to have ways to easily turn it on and off as a command line flag, as well as a persistent always-on config setting.

It seems okay to me to leave always-on mode OFF by default, as it seems possible there are other use cases where this would not really be an issue.

PR #1395, opened 2023-02-24
Add comment to pgclirc pager section on default LESS

andyscho:
## Description
Add brief comment detailing the behavior in [pgcli/main.py](https://github.com/dbcli/pgcli/blob/141873f86d299a7c8469392367b820960d01707a/pgcli/main.py#L160).

I also changed the example `pager` setting since it seemed misleading that the `less` options in the `pager` setting would be able to be used as if the `LESS` environment variable is not being set automatically.

## Checklist

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -13,6 +13,7 @@ Features:
 * Config option to not restart connection when cancelling a `destructive_warning` query. By default,
   it will now not restart.
 * Config option to always run with a single connection.
+* Add comment explaining default LESS environment variable behavior and change example pager setting.
 
 Bug fixes:
 ----------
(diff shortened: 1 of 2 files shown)


j-bennet:
Thanks for the PR @andyscho ! While you're at it, you can also add a link to https://www.pgcli.com/pager for more info.

andyscho:
Adding the link to the pager page seems reasonable. [I made a PR to also add some info to it here.](https://github.com/dbcli/pgcli.com/pull/54)

I also updated this PR and its description since I think it's good to also change the example pager setting in the config.

I'm also not really opposed to changing the current behavior in `main.py` to something like "only set the `LESS` environment variable if the `pager` setting in the config is not set", but I don't have a strong opinion. It also seemed like it would be a bit sad to force users updating to a later version of pgcli to change their config to retain the `less` behavior they previously had in pgcli.

j-bennet:
I think it's fine as is, @andyscho. I'll go ahead and merge it. Thank you!

PR #1397, opened 2023-03-03
feature: Add config option to require a transaction for destructive statements

dbaty:
When this option is on, any statement that is deemed destructive (through the use of the existing `destructive_warning` config option) will not be executed unless a transaction has been started. The default value is off (thus not changing the current behaviour).

Demo: [asciicast](https://asciinema.org/a/564064)

Rationale: I think it's a good practice to execute "destructing" queries within a transaction (even more so when executing multiple queries, obviously!). It provides a last chance to rollback if things go wrong. Especially when it (unfortunately) becomes a reflex to just type "yes, I do want to execute this _destructive_ query, duh!". ;)

Notes:
- tests: I don't know what/which tests to add. I could not see appropriate unit tests to add or adapt, nor any related behaviour test. I'll gladly write tests if anyone can point me in the right direction.
- docs: If merged, [this page](https://github.com/dbcli/pgcli.com/blob/main/content/pages/config.md) will need to be updated to include the new config option.

## Checklist
- [ ] ~Please squash merge this pull request.~ There is a "lint" commit that could be merged separately from the "feature" commit.

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -64,6 +64,7 @@ from .config import (
 from .key_bindings import pgcli_bindings
 from .packages.formatter.sqlformatter import register_new_formatter
 from .packages.prompt_utils import confirm, confirm_destructive_query
+from .packages.parseutils import is_destructive
 from .packages.parseutils import parse_destructive_warning
 from .__init__ import __version__
 
@@ -233,6 +234,9 @@ class PGCli:
         )
         self.destructive_warning_restarts_connection = c["main"].as_bool(
             "destructive_warning_restarts_connection"
+        )
+        self.destructive_statements_require_transaction = c["main"].as_bool(
+            "destructive_statements_require_transaction"
         )
 
         self.less_chatty = bool(less_chatty) or c["main"].as_bool("less_chatty")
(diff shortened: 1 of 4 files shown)


pgcli/main.py
```diff
@@ -432,15 +436,21 @@
...
-            and confirm_destructive_query(
+        if self.destructive_warning:
+            if not self.pgexecute.valid_transaction() and is_destructive(
+                query, self.destructive_warning
+            ):
+                click.secho(
+                    "Destructive statements must be run within a transaction. "
+                    "Command execution stopped."
+                )
+                return [(None, None, None, message)]
```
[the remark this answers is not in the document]
dbaty: This failure has been fixed in a following push-forced commit.

pgcli/main.py
```diff
@@ -64,15 +64,16 @@
...
 from .packages.prompt_utils import confirm, confirm_destructive_query
+from .packages.parseutils import is_destructive
 from .packages.parseutils import parse_destructive_warning
 from .__init__ import __version__
 
 click.disable_unicode_literals_warning = True
 
 try:
-    from urlparse import urlparse, unquote, parse_qs
+    from urlparse import urlparse
```
amjith, line 74: Can we remove this completely? I believe this import with try/catch was designed to accommodate Python 2.7 which we no longer support.

amjith:
Thank you!

pgcli/main.py
dbaty: @amjith : I removed this try/except (and a few other leftovers of Python 2 support) in #1399.

PR #1400, opened 2023-03-29
Refuse to quit cli if a transaction is ongoing.

### LINKED ISSUE #1071: Warning of existing transaction when quitting pgcli (asking for COMMIT or ROLLBACK)
It would be really helpful to warn the user who leaves his session (quitting pgcli) that in case of an existing transaction, he will need to **explicitly commit or rollback** in order not to lose the pending changes, hence blocking the exit

This feature might be related to #410

dbaty:
## Description

If a transaction is ongoing, quitting the cli (with "ctrl-d" or "quit") is not possible anymore. An error message is displayed. The user is forced to commit or rollback the transaction before quitting.

Fixes #1071.

## Notes

1. I did not add a new configuration option. **This commit thus changes the current behaviour of pgcli** (for good, I believe). This is debatable, I could add a configuration option (off by default) if preferable.
2. I added a "behave" test. This is my first one, feel free to comment if it's not the right way to do it.
3. Demo: [asciicast](https://asciinema.org/a/571678)

## Checklist

- ~[ ] I've added my name to the `AUTHORS` file (or it's already there).~ (I am already in this file.)

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -4,6 +4,7 @@ Upcoming
 Features:
 ---------
 
+* Ask for confirmation when quitting cli while a transaction is ongoing.
 * New `destructive_statements_require_transaction` config option to refuse to execute a
   destructive SQL statement if outside a transaction. This option is off by default.
 * Changed the `destructive_warning` config to be a list of commands that are considered
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -800,6 +800,34 @@ class PGCli:
                 logger.debug("Search path: %r", self.completer.search_path)
         return query
 
+    def _check_ongoing_transaction_and_allow_quitting(self):
+        """Return whether we can really quit, possibly by asking the
+        user to confirm so if there is an ongoing transaction.
+        """
+        if not self.pgexecute.valid_transaction():
+            return True
+        while 1:
+            try:
+                choice = click.prompt(
+                    "A transaction is ongoing. Choose `c` to COMMIT, `r` to ROLLBACK, `a` to abort exit.",
+                    default="a",
+                )
+            except click.Abort:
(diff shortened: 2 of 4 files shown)


j-bennet:
@dbaty Very nice work, you went above and beyond, what with the `behave` test, and the demo.

I think I would things a little differently. If the user tries to quit and a transaction is in progress, I would output a message as follows:

```
A transaction is ongoing. Choose c to COMMIT or r to ROLLBACK Default: [r] >
```

If the user presses `ENTER`, transaction is rolled back, and `pgcli` exits.
If the user types `r` and presses `ENTER`, same.
If the user types `c` and presses `ENTER`, transaction is rolled back, and `pgcli` exits.
If the user types anything else besides `r` or `c`, same message is repeated.

It's slightly more complicated to implement, but nicer user experience, because if the user is ok with losing their changes, they can just press `ENTER` once more to quit, without typing anything.

What do you think?

dbaty:
That makes sense. I'll try to implement that next week.

j-bennet:
Hi @dbaty, this looks great, and works as expected. Thanks a lot, great feature! Merging. 🍫

matt0089:
Hi everyone 👋.  Thank you for this great tool, not sure how I'd get work done without it tbh.

I have an issue where pgcli thinks I've started a transaction periodically, when I haven't.  Then when I try to exit, it never lets me, so I have to do a pkill in another window.

It usually happens after I've cancelled a query (with ctrl-c).

Is this a known issue?  I've searched Github issues and haven't found any mentions of this behavior.  If not, I may open one (with more details).  I'll see if I can come up with a fix also.

I'm using pgcli v4.1.0

dbaty:
Hi. Could you please open an issue that details how to reproduce it? Thanks!

PR #1410, opened 2023-07-02
confirm_destructive_query: Use confirm rather than prompt

sharonyogev:
## Description
Until now, the destructive query prompt, which is a great feature, uses click `prompt` with `type=bool` to confirm the query, those resulting weird behavior such:
```
postgres@localhost:tests> UPDATE test SET id=0;
You're about to run a destructive command.
Do you want to proceed? (y/n): t
Your call!
UPDATE 0
Time: 0.014s
postgres@localhost:tests>
```
This is because it capture `t`/`true` as yes and `f`/`false` as no, for example

I think we should use click `confirm` which is more suitable for this action
We can use it with default to "No" or without a default and force the user to enter
```
postgres@localhost:tests> UPDATE test SET id=0;
You're about to run a destructive command.
Do you want to proceed? [y/N]: t
Error: invalid input
You're about to run a destructive command.
Do you want to proceed? [y/N]:
Wise choice!
cancelled query
postgres@localhost:tests>
```

## Checklist

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -129,6 +129,7 @@ Contributors:
     * Damien Baty (dbaty)
     * blag
     * Rob Berry (rob-b)
+    * Sharon Yogev (sharonyogev)
 
 Creator:
 --------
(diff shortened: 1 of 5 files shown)


sharonyogev:
Thanks @j-bennet fixed the tests

sharonyogev:
@j-bennet Hi, the failure on 3.12 does not look like it's related to my change (Timeout exceeded), any idea? Thanks

j-bennet:
@sharonyogev Integration tests can be flaky; I re-ran and now it's fine.

sharonyogev:
@j-bennet Yay!! Thanks 🙏

PR #1441, opened 2023-11-15
Fix short host (\h) display in prompt when using an IP address

### LINKED ISSUE #964: \h shows first octet of IP address
Pgcli shows the first octet of an IP address ("127" for example) when giving an IP address as host name. The `\h` variable ("Short hostname of the server (up to first '.')" is part of the default prompt. Since IP addresses don't have a short form, this should probably be the full IP address.

Psql shows the same behaviour (in case you aim for consistency with `psql` even in incorrect cases)
```
pgcli postgresql://postgres:password@127.0.0.1
Server: PostgreSQL 11.0
Version: 2.0.0
[...]
postgres@127:5432/postgres dbcli>
```

dbaty:
When connecting to an IPv4 address (`pgcli -h 127.0.0.1`), trying to
use the "short host" in the prompt (with `\h`) would only display the
first octet (127). Now it shows the full IP.

Fixes #964.

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -1,3 +1,12 @@
+Upcoming
+========
+
+Bug fixes:
+----------
+
+* Fix display of "short host" in prompt (with `\h`) for IPv4 addresses ([issue 964](https://github.com/dbcli/pgcli/issues/964)).
+
+
 ==================
 4.0.1 (2023-10-30)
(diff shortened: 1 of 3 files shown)


j-bennet (approved):
Nice.

PR #1456, opened 2024-04-06
F1 shows man page for current query

serpent7776:
## Description
Make F1 key launch man page for current query. E.g. I have `CREATE TABLE foo` as my query and I press `F1`. The man page `CREATE_TABLE` is launched. This requires man pages to be installed locally, I'm not sure they're always installed. I have it on my manjaro, but I don't see man pages on ubuntu, only the html docs.

This is currently just a scratch implementation, let me know if you're interested in having such functionality.

Also, I'm not a python dev, any suggestions welcomed.

## Checklist


j-bennet:
Nice. :) If this is useful to you, it's useful to other people.

serpent7776:
I implemented a simple heuristic that should work for most cases.

Two cases that are not supported are:

- `CREATE TABLE AS`
- `SELECT INTO`

These cases are too complex to come up with a heuristic. An actual SQL parser would need to be used to support these.

Now I have an issue with showing SQL query after man page quits. The user entered query should be re-desplayed, but I don't know how to do that. I tried `click.clear()`, `event.app.invalidate()` and `get_app().current_buffer.reset()`, but none of them work.
Maybe man page should be displayed in separate window? I don't know how to do that though.

j-bennet:
@jonathanslenders can you advise?

Issue #1470, opened 2024-06-16
Document how to replace pg_isready with pgcli


offbyone:
I'm gonna try figure this out for myself, but it would be super cool to have this info; I want to ditch requiring contributors to my project to install the postgres command line, since I'm already installing pgcli in development.

offbyone:
`pgcli --list &>/dev/null` seems to do the trick; would it be worth adding a `pgcli --ping` command line option to do this?

PR #1471, opened 2024-06-16
Add `--ping` as a CLI option

offbyone:
## Description

This builds in the functionality asked for in #1470; it allows pgcli to replace `pg_isready` from the postgresql command line toolchain

Fixes: #1470 

## Checklist

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -1643,7 +1643,7 @@ def cli(
             ...
 
         if results:
-            pgcli.echo("PONG")
+            click.echo("PONG")
             sys.exit(0)
         else:
             click.secho(
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -1639,19 +1639,15 @@ def cli(
         try:
             results = list(pgcli.pgexecute.run("SELECT 1"))
         except Exception:
-            # yup, really, anything here
-            ...
-
-        if results:
-            click.echo("PONG")
-            sys.exit(0)
-        else:
             click.secho(
                 "Could not connect to the database. Please check that the database is running.",
                 err=True,
                 fg="red",
             )
             sys.exit(1)
+        else:
+            click.echo("PONG")
+            sys.exit(0)
 
     pgcli.logger.debug(
         "Launch Params: \n" "\tdatabase: %r" "\tuser: %r" "\thost: %r" "\tport: %r",
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -1635,9 +1635,8 @@ def cli(
         sys.exit(0)
 
     if ping_database:
-        results = None
         try:
-            results = list(pgcli.pgexecute.run("SELECT 1"))
+            list(pgcli.pgexecute.run("SELECT 1"))
         except Exception:
             click.secho(
                 "Could not connect to the database. Please check that the database is running.",
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -1464,6 +1464,13 @@ class PGCli:
     help="list available databases, then exit.",
 )
 @click.option(
+    "--ping",
+    "ping_database",
+    is_flag=True,
+    default=False,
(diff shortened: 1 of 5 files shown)


j-bennet:
@offbyone This looks nice! I'd add an integration test scenario for this as well, see `--list` for n example:

https://github.com/dbcli/pgcli/blob/46fe654cd16a3af837639dc2c8be60d1b7a038c7/tests/features/basic_commands.feature#L50-L52

offbyone:
I added two feature tests, but one of them is failing and I don't quite know why; the exit code 0 is coming back with exit code 120, which ... I do not understand. If you know why, I'd be thrilled to update the spec.

pgcli/main.py
```diff
@@ -1626,6 +1634,25 @@ def cli(
...
+        if results:
+            click.echo("PONG")
+            sys.exit(0)
+        else:
+            click.secho(
+                "Could not connect to the database. Please check that the database is running.",
+                err=True,
+                fg="red",
+            )
+            sys.exit(1)
```
dbaty, line 1654: [style] The comment in the `except` block sounds a bit weird to me. I think that an `else` block would be clearer (and in fact shorter):

```suggestion
        except Exception:
            click.secho(
                "Could not connect to the database. Please check that the database is running.",
                err=True,
                fg="red",
            )
            sys.exit(1)
        else:
            click.echo("PONG")
            sys.exit(0)
```

pgcli/main.py
```diff
@@ -1626,6 +1634,25 @@ def cli(
 
         sys.exit(0)
 
+    if ping_database:
+        results = None
+        try:
+            results = list(pgcli.pgexecute.run("SELECT 1"))
+        except Exception:
```
dbaty, line 1641: I wonder what exception we could get, here. If pgcli cannot connect to the server (or use the database), we'll fail when calling `PGCLI.connect()` a few blocks before, not here.

offbyone:
I am 100% fine with any of those changes; I'm not super familiar with the codebase, and it took me until earlier today to even figure out how to test it effectively. Feel free! I left "Allow edits by maintainers" enabled so you are more than welcome to push any changes that make sense to you!

dbaty:
> I added two feature tests, but one of them is failing and I don't quite know why; the exit code 0 is coming back with exit code 120, which ... I do not understand. If you know why, I'd be thrilled to update the spec.

I pulled your branch, ran `behave` and did not get any error. Could you clarify which command you run and the error you get?

offbyone:
I don't, any more; I found the cause of my failure to write the tests correctly and pushed an update to it.

pgcli/main.py
```diff
@@ -1626,6 +1634,21 @@ def cli(
 
         sys.exit(0)
 
+    if ping_database:
+        results = None
+        try:
+            results = list(pgcli.pgexecute.run("SELECT 1"))
```
j-bennet, line 1640:
```suggestion
        try:
            list(pgcli.pgexecute.run("SELECT 1"))
```

pgcli/main.py
j-bennet: You're right, @dbaty, it's extremely unlikely, but not impossible. I've seen this once in my life, and not with postgresql; database server was up, but the transaction log was full. I'm ok with keeping the try/catch.

j-bennet:
Looks good @offbyone; merging. 🍒 Thanks for your help, @dbaty!

offbyone:
Thanks to all of you for the cleanups and finalizing it!

pgcli/main.py
dbaty: > You're right, @dbaty, it's extremely unlikely, but not impossible. I've seen this once in my life, and not with postgresql; database server was up, but the transaction log was full. 

@j-bennet : then it would be useful to log the exact error we got, instead of the misleading "could not connect to the database".

Issue #247, opened 2015-05-28
Run commands on startup


lee-kagiso:
I would like to run some commands on start up, like \x.

Is this possible?

amjith:
Right now there is no way to have a startup command. 

Is `\x` the only thing you have in mind or do you have other ideas for this request? 

The reason I ask is because I can made the expanded output to be a config option in `~/.pgclirc` and you'll be able to achieve this feature. If you have more things in mind, I'll have to think of a different solution.

lee-kagiso:
I would also like to be able to change the value for nulls:

```\pset null '(null)'````.

In my current psqlrc I also echo out a cheatsheet of \ commands, but I can live without those...

https://github.com/leebrooks0/dotfiles/blob/master/psqlrc

amjith:
The `null` thing is not going to work. `\pset` is not supported by pgcli. Right now all null values are displayed as `<null>` by default.

The idea is that `pgcli` should ship with sane defaults that requires no configuration. But there does exist a config file which tweaks certain settings such as colorscheme and keybindings. 

I can sympathize with the cheat sheet commands. I'll see what I can do.

lee-kagiso:
<null> is perfect, anything other than blank is good.

amjith:
Also I didn't know about `\x auto` that seems like an awesome feature of psql that I might steal. :)

lee-kagiso:
One other thing worth mentioning, can you also add to the config file the
ability to use multiline by default?

amjith:
Aha! This is already part of the config file. Open up `~/.pgclirc` and you'll see an option to enable or disable multi-line mode.

normk61:
Is it possible to use the .pgclirc config file to set my default search path, like I can in .psqlrc?

spinus:
I would join to the question how to set up defaul search_path. Is it possible now?

Gollum999:
I too would like the ability to execute arbitrary commands on startup.  My use case is a little different; I want to be able to `CREATE TEMP VIEW` on startup for some tables that I frequently join together.  I don't think "named queries" are what I'm looking for because AFAICT you can't use named queries as part of another query, e.g. I wouldn't be able to join a named query against some other table.  (Correct me if I'm wrong.)

`psql` just interprets everything in `.psqlrc` literally; I'm not super familiar with the details of the config file format, but would it be possible to add a section like "literal startup commands" that just sends every line straight to `_evaluate_command`?

assembler:
[...]

PR #1482, opened 2024-10-16
Set session timezone to local timezone on startup

maximsmol:
## Description
Sets the session time zone to the system time zone after connecting.

Adds [`tzlocal`](https://github.com/regebro/tzlocal/tree/master) as a dependency for a cross-platform way of getting the current time and handling the `TZ` environment variable. I've considered including code for this directly but we need to handle a lot of different Windows quirks. This package adds 126KB to the on-disk installation size.

## Checklist

- [X] I've added this contribution to the `changelog.rst`.

- [X] I installed pre-commit hooks (`pip install pre-commit && pre-commit install`), and ran `black` on my code.


dbaty:
Thanks for your contribution! :) I am not familiar with the `set time zone` command (though I have read [its doc](https://www.postgresql.org/docs/current/sql-set.html)). I feel like it could lead to unexpected results (and a difference in behaviour with what we have now in pgcli, and also a difference between pgcli and psql).

For example, suppose that the server is on timezone "tz0", user1 on tz1, user2 on tz2, user3 on tz3. Currently, all users calling `now()` will get the same timestamp with time zone tz0 (the server timezone). With your proposal, they could get 3 different things:
- user1 uses pgli and gets a timestamp on timezone tz1
- user2 uses pgli and gets a timestamp on timezone tz2
- user3 uses **psql** (**not** pgcli) and gets a timestamp on timezone tz0 (server timezone).

We end up with 3 different values. They represent the same time, only expressed under 3 different timezones, but still these are different values. Also, what concerns me is that what you get in pgcli is different from what you get on psql. That in itself seems problematic to me. At least, problematic enough that this feature should not be activated by default.

Could you please comment on that and clarify the other changes in behaviour (if any)?

I believe that this kind of "customization" should rather take place in "startup commands". The idea of such a feature has been expressed some time ago (see #247). What do you think?

@j-bennet : what's your take on this?

maximsmol:
Thanks for responding so quickly.

# Concerns

To summarize, there are two concerns:
1. different users may see the same value in time represented by visually different values,
2. behavior is different from psql.

In response:
1. Note that the values are in fact *equal* according to Postgres itself, despite having different string representations and time zone offsets. This is also the case in other popular languages (see below for experiments).
2. Being different from psql is, for me, the entire point of using pgcli.

---

The only other potential issue I can come up with is for people using `timestamp without timezone` since they would actually resolve to different moments in time as they are always in the "current" time zone. [Those are already frowned upon for a lot of other reasons, however.](https://wiki.postgresql.org/wiki/Don't_Do_This#Don.27t_use_timestamp_.28without_time_zone.29)

# Startup Commands

I think "startup commands" would be nice *in addition* to this PR. It's not a good *replacement* because to fully replicate this behavior you would need to set the commands dynamically each time, in case the time zone changed (laptops in particular tend to move time zones a lot).

Here is what we currently do to get psql (which does support running commands before a REPL) to use the local time zone:
```bash
#!/usr/bin/env bash

tz_data_raw=$(strings /etc/localtime| tail -1)
IFS=',' read -r -a tz_data <<< "$tz_data_raw"

URL="postgresql://$SQL_ADMIN_USER:$SQL_ADMIN_PW@$SQL_DB_HOST:$SQL_DB_PORT/$SQL_DB_NAME$SQL_URL_EXTRA"
psql "$URL" -c "set time zone ${tz_data[0]}" "$@"
```

As you can see, it's not a great solution.

---

As a compromise, maybe we can make this opt-in with a config parameter? I am generally against random configuration parameters but if there is an impasse I'd rather do that than have to make each member of my team install a fork.

# Thoughts

- Showing time in the current user's local time is a ubiquitous default in interactive programs. Even calendar programs do this despite much larger potential issues (e.g. they are the only place where timezone-naive datetime can be preferable).
- Converting from UTC in places with daylight savings rules, large offsets is really annoying, especially for queries that span months or years.
- Converting from non-UTC time zones for queries that include international data is even worse.
- UTC timestamps sometimes cause confusion if users do not consciously make the conversion.
- This is one of the few parts where the experience is currently *worse* than in psql where it is at least technically possible to automatically set the current time zone. This could alternatively be solved with startup commands, though I doubt people would use the ugly script above and so would likely have to manually adjust the time zone command periodically.

# Experiment: Are Datetimes In Different Time Zones Equal?

Postgres:
```
> select '2024-10-17 00:00:00-07'::timestamptz = '2024-10-17 01:00:00-06'::timestamptz
?column?
True
SELECT 1
Time: 0.040s
```

JavaScript:
```js
> const x = new Date("2024-10-17 00:00:00-07")
> const y = new Date("2024-10-17 01:00:00-06")
> x - y === 0
true
> x >= y && x <= y
true
> x === new Date("2024-10-17 00:00:00-07")
// this is to show that `===` is referential equality here for some reason
false
```

Python:
```py
>>> x
datetime.datetime(2024, 10, 16, 0, 0, tzinfo=datetime.timezone(datetime.timedelta(days=-1, seconds=61200), 'PDT'))
>>> x.astimezone(timezone(timedelta(0)))
datetime.datetime(2024, 10, 16, 7, 0, tzinfo=datetime.timezone.utc)
>>> x == x.astimezone(timezone(timedelta(0)))
True
```

pgcli/main.py
```diff
@@ -1624,6 +1625,8 @@ def cli(
     else:
         pgcli.connect(database, host, user, port)
 
+    pgcli.pgexecute.set_timezone(tzlocal.get_localzone_name())
```
j-bennet, line 1628: Does this ever raise an exception or return a None?
What happens if I set `TZ` to a non-existent timezone name?

j-bennet:
@maximsmol Well, you have a use case for the feature, so it must be useful! However, I would like to add a couple of things:
* opt-off via a config setting
* a message visible to the user about timezone being changed. Bonus points if it also tells you what the server timezone is, and how to opt out from the feature.

maximsmol:
@j-bennet Sorry for the delay, I added the config variable and updated the error messages to refer to it and specify the server's default time zone

---

Corresponding pgcli.com PR is here: https://github.com/dbcli/pgcli.com/pull/68
(note that I included the unrelated `keyring` setting which is currently missing on the website)

Not sure how releases are timed with website updates, guessing this one will need to stay open until the main PR makes it into a release

j-bennet:
Thank you @maximsmol, looks fine to me. Merging.

j-bennet:
4.3.0 is released.

PR #1503, opened 2025-04-08
Update Linux installation instructions

### LINKED ISSUE #1502: Do not recommend `sudo pip`, at least on Linux
Recent Debian versions refuse to install packages as root with pip. Recent Fedora versions display a big warning message (although they proceed by default). This is a good idea, because installing with root uses the system Python, and installing things with that can cause unintended side-effects.

(I'm not sure I would recommend this either on macOS, but I have less knowledge about that.)

I would recommend steering people towards distribution packages. Debian stable has a somewhat old version (well, the version that was out when Debian stable was released), but Fedora tracks closely pgcli.

If people want to install a newer version, I'd recommend [pipx](https://github.com/pypa/pipx), which is packaged for common distros, or uvx.

I'm happy to send a PR editing the README if you let me know what you would prefer.

alexpdp7:
Closes #1502

## Description

Installing packages using `sudo pip` is actively discouraged. (This can break some systems. Some distributions actually disable this.) Additionally, I have found that many distributions carry `pgcli`.

* Mention first using distribution packages.
* Recommend the use of `pipx` and `uvx`.

See rendered rst here <https://github.com/alexpdp7/pgcli/tree/main?tab=readme-ov-file#linux>.

## Checklist

- [ ] I've added this contribution to the `changelog.rst`. **I don't think this does not belong to the changelog.**
- [ ] I've added my name to the `AUTHORS` file (or it's already there). **I don't think this is a significant contribution. Happy to add myself if it's a policy.**
- [ ] I installed pre-commit hooks (`pip install pre-commit && pre-commit install`), and ran `black` on my code. **No code changes.**
- [x] Please squash merge this pull request (uncheck if you'd like us to merge as multiple commits) **I don't care, it's a single commit.**

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -209,43 +209,27 @@ If pip is not installed check if easy_install is available on the system.
 Linux:
 ======
 
-In depth getting started guide for ``pip`` - https://pip.pypa.io/en/latest/installation/
+Many distributions have ``pgcli`` packages.
+Refer to https://repology.org/project/pgcli/versions or your distribution to check the available versions.
 
-Check if pip is already available in your system.
+Alternatively, you can use tools such as `pipx`_ or `uvx`_ to install the latest published package to an isolated virtual environment.
 
-::
+.. _pipx: https://pipx.pypa.io/
+.. _uvx: https://docs.astral.sh/uv/guides/tools/
 
-    $ which pip
-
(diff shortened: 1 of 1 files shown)


j-bennet (approved):
Ok! Thanks for keeping us up to date.

PR #1527, opened 2025-11-19
Update Dockerfile to use Python 3.9

StefanMallia:
## Description
Small change to the Dockerfile to use Python 3.9 as required by pyproject.toml. Currently the docker build fails on "Package 'pgcli' requires a different Python: 3.8.20 not in '>=3.9'.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.8
+FROM python:3.9
 
 COPY . /app
 RUN cd /app && pip install -e .


dbaty:
Python 3.9 has reached its end of life last October ([see here](https://devguide.python.org/versions/)). We should rather use the latest version: 3.14.

dbaty:
Using the latest version of Python should not be notably more complex than your fix, is it? :)

PR #1536, opened 2025-11-28
Adding port so keyring is not overwritten

bechampion:
## Description
Keys in keychain are overwritten when same user/host is used (this happens a lot when you proxy over ssh where the host is always localhost and username pretty much is always the same)
### Before change
image

### After change
image

## Checklist


j-bennet:
@DiegoDAF I think it's ok to not try for backwards compatibility in this case.

DiegoDAF:
yep, it's not necessary to keep the compatibility.

j-bennet:
Thanks @bechampion! 🍒

PR #1541, opened 2025-12-04
Implement daily log rotation with improved log management

DiegoDAF:
## Summary

This PR implements automatic daily log rotation and improves the default log location for pgcli.

### 1. Daily Log Rotation
- Implements automatic log rotation at midnight using `TimedRotatingFileHandler`
- Keeps 30 days of log history automatically
- Log files are named with the format: `pgcli.YYYY-MM-DD.log`
- Old logs are automatically cleaned up after 30 days

### 2. Improved Default Log Location
- New default location: `/var/log/pgcli/pgcli.log` (system-wide logging)
- Automatic fallback to `~/.config/pgcli/log` if user lacks system permissions
- Creates directories automatically if they don't exist

### 3. Updated Documentation
- Updated `pgclirc` configuration file with new log behavior
- Added Internal section to `changelog.rst`

## Technical Details

**Before:**
- Single log file that grew indefinitely
- Default location: `~/.config/pgcli/log`
- Manual cleanup required

**After:**
- Automatic daily rotation at midnight
- 30-day retention policy
- Better default location with fallback
- Log format: `pgcli.YYYY-MM-DD.log`

## Files Modified

- `pgcli/main.py`: Added `logging.handlers` import and updated `initialize_logging()` function
- `pgcli/pgclirc`: Updated log_file documentation
- `changelog.rst`: Added entry for log rotation feature

## Testing

Tested manually:
- Log rotation works correctly at midnight
- Automatic directory creation works
- Fallback to user directory works when lacking system permissions
- Old logs are properly cleaned up after 30 days
- Log file naming follows the new format

## Backward Compatibility

- Fully backward compatible
- Existing log files are not affected
- Users with custom `log_file` settings in config will continue to work as before
- Only affects default behavior when no custom log location is specified

## Benefits

1. **Better disk management**: Automatic cleanup prevents disk space issues
2. **Easier troubleshooting**: Date-based log files make it easier to find relevant logs
3. **System-friendly**: Uses standard `/var/log` location when possible
4. **No breaking changes**: Respects user configuration and provides sensible fallback

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -561,15 +561,21 @@ class PGCli:
             # Use TimedRotatingFileHandler for daily log rotation
             # Rotates at midnight, keeps 30 days of logs
             expanded_log_file = os.path.expanduser(log_file)
+
+            # Remove .log extension if present to avoid pgcli.log.log pattern
+            base_log_file = expanded_log_file
+            if base_log_file.endswith('.log'):
+                base_log_file = base_log_file[:-4]
+
             handler = logging.handlers.TimedRotatingFileHandler(
-                expanded_log_file,
+                base_log_file,
                 when='midnight',
                 interval=1,
                 backupCount=30,
                 encoding='utf-8'
             )
-            # Format: pgcli.log.2025-12-02
-            handler.suffix = "%Y-%m-%d"
+            # Format: pgcli.YYYY-MM-DD.log
+            handler.suffix = ".%Y-%m-%d.log"
 
         level_map = {
             "CRITICAL": logging.CRITICAL,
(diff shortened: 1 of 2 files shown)


j-bennet:
@DiegoDAF Well, I was happy to see a new feature in pgcli. :) But @dbaty has a good point. It's the code we have to maintain. And we already find it very difficult to juggle contributing to `pgcli` in our "free time" (yeah right... free time? who has that?) with our day jobs and family responsibilities.

Still, I think that supplying a custom logging configuration is a reasonable ask. But I would do it in a declarative way, instead of the imperative way - less code to maintain.

1. User would configure a location of `logging_config` in `pgclirc`.
2. If that file is configured and present, we would initialize the logger [from dict](https://docs.python.org/3/library/logging.config.html#logging.config.dictConfig) stored in that file.
3. Otherwise, proceed with our regular `initialize_logging`.

WDYT?

P.S. I agree with @dbaty on `/var/log/pgcli`.

PR #1544, opened 2025-12-05
Add support for -y/--yes flag to bypass confirmation prompts

DiegoDAF:
## Summary

This PR adds support for the `-y`/`--yes` option to pgcli, allowing users to bypass destructive command confirmation prompts. This is particularly useful for automated scripts and CI/CD pipelines.

## Features

- **Skip confirmations**: `pgcli -y` or `pgcli --yes`
- **Automated-friendly**: Perfect for scripts and CI/CD pipelines
- **Safe by default**: Flag must be explicitly set
- **Respects transactions**: Transaction requirements still apply even with `-y`
- **Clean output**: Suppresses "Your call!" message when auto-confirming

## Use Cases

```bash
# CI/CD pipeline
pgcli -y -c "DROP TABLE old_data;"

# Automated maintenance script
pgcli --yes -f cleanup.sql

# Interactive use (still requires confirmation)
pgcli  # without -y, prompts as usual
```

## Implementation Details

- Added `force_destructive` parameter to PGCli class
- Modified destructive command checks in two locations:
  - `execute_command()` method for programmatic execution
  - Interactive command handler in `run_cli()`
- When `-y` is set, `confirm_destructive_query()` is bypassed
- Transaction requirements are NOT bypassed (safety measure)
- "Your call!" message is suppressed when using `-y` (no user interaction occurred)

## Testing

Comprehensive unit tests included:
- Initialization with different flag values
- Confirmation bypass when flag is True
- Normal confirmation flow when flag is False/not set
- Mock verification that confirmation is/isn't called as expected

## Safety Considerations

- Default behavior unchanged (confirmations still required)
- Transaction requirements remain enforced
- Users must explicitly opt-in to skip confirmations
- Helps prevent accidental destructive operations in interactive use

---

Part of the feature list in discussion #1603: this is **item 9** (`-y/--yes`).


pgcli/main.py
```diff
@@ -484,7 +486,10 @@ def execute_from_file(self, pattern, **_):
             ):
                 message = "Destructive statements must be run within a transaction. Command execution stopped."
                 return [(None, None, None, message)]
-            destroy = confirm_destructive_query(query, self.destructive_warning, self.dsn_alias)
+            if self.force_destructive:
```
j-bennet: Rather than check `self.force_destructive` in two places, I'd change the signature of `confirm_destructive_query` to include it as a parameter.

DiegoDAF: Done, `confirm_destructive_query` takes the flag now and owns the whole decision:

```python
def confirm_destructive_query(queries, keywords, alias, force=False):
```

It returns `None` when the query is not destructive, `True` when `force` is set, `None` when there is no tty, and otherwise prompts as before. Both call sites in `main.py` just pass `self.force_destructive` through.

One `self.force_destructive` check is left in `main.py`, but it is a different concern: it suppresses the `Your call!` message, which would otherwise be printed on every destructive statement in a script that never asked anything. Happy to move that too if you would rather have it inside the helper.

Ready for another look whenever you have time.

j-bennet:
Ok @DiegoDAF , this is in.

PR #1562, opened 2026-03-18
fix: respect DSN password over PGPASSWORD env var

BillionClaw:
When connecting via DSN URI containing a password (e.g., `postgresql://user:pass@host/db`), pgcli incorrectly prioritized the PGPASSWORD environment variable over the password embedded in the DSN. This caused authentication failures when PGPASSWORD was set to a different value.

Changed the connection logic to check if the DSN contains a password before falling back to PGPASSWORD. If a password is present in the DSN, it takes precedence.

Fixes #1555

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -632,7 +632,13 @@ class PGCli:
         # If password prompt is not forced but no password is provided, try
         # getting it from environment variable.
         if not self.force_passwd_prompt and not passwd:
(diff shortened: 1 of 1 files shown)


lgx0319:
Friendly bump -- this PR has been open for a while. Happy to make any changes if needed, or close it if no longer relevant.

j-bennet:
Thanks @BillionClaw !

ochaton:
Thank you a lot ❤️

PR #1570, opened 2026-04-03
Make auto-suggest optional

bimlas:
Auto-suggest could distract the user. It can be disabled by setting `history_file = /dev/null`, but in this case, none of the historical queries are saved. Instead, add an option to disable auto-suggest while keeping the history.


j-bennet:
Thank you @bimlas !

PR #1559, opened 2026-03-02
Fix trailing SQL comments breaking multiline submit and execution

DiegoDAF:
## Summary

Fixes two related bugs when a SQL comment follows the semicolon, e.g.:

```sql
vacuum freeze verbose tpd.file_delivery; -- 82% towards emergency, 971 MB
```

**Bug 1 — `pgbuffer._is_complete()`:** `sql.endswith(";")` returned `False` when a trailing comment followed the semicolon. In multiline mode, pressing Enter never submitted the query — the UI appeared frozen.

**Bug 2 — `pgexecute.run()`:** `sql.rstrip(";")` couldn't strip the semicolon because the string ended with the comment text, not `;`. The malformed SQL (with embedded `;`) was sent to PostgreSQL via psycopg's extended query protocol, which doesn't support multiple statements.

## Fix

Both locations now use `sqlparse.format(sql, strip_comments=True)` before checking for / removing the trailing semicolon.

## Files Changed

- `pgcli/pgbuffer.py` — `_is_complete()` strips comments before `endswith(";")`
- `pgcli/pgexecute.py` — `run()` strips comments before `rstrip(";")`
- `changelog.rst` — Added bug fix entry to Upcoming
- `tests/test_trailing_comments.py` — 12 new tests covering edge cases

## Test plan

- [x] All 12 new tests pass (`pytest tests/test_trailing_comments.py -v`)
- [x] Full test suite passes (2788 passed, 128 skipped)
- [x] Manually verified: `vacuum freeze verbose tpd.file_delivery; -- 82% towards emergency` executes correctly
- [ ] Verify multiline mode submits correctly with trailing comments

PR #1576, opened 2026-04-18
fix: handle trailing SQL comments in multiline submit

bobo-xxx:
Fixes two related bugs when a SQL comment follows the semicolon:

**Bug 1 — pgbuffer._is_complete():** sql.endswith(';') returned False when a trailing comment followed the semicolon. In multiline mode, pressing Enter never submitted the query.

**Bug 2 — pgexecute.run():** sql.rstrip(';') couldn't strip the semicolon because the string ended with comment text. The malformed SQL was sent to PostgreSQL via psycopg's extended query protocol.

Both locations now use sqlparse.format(sql, strip_comments=True) before checking for / removing the trailing semicolon.

Fixes dbcli/pgcli#1559

### DIFF
--- a/pgcli/pgbuffer.py
+++ b/pgcli/pgbuffer.py
@@ -4,6 +4,7 @@ from prompt_toolkit.enums import DEFAULT_BUFFER
 from prompt_toolkit.filters import Condition
 from prompt_toolkit.application import get_app
 from .packages.parseutils.utils import is_open_quote
+import sqlparse
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


j-bennet:
Hey @bobo-xxx , Copilot has some good points there. Also, I think you'll need to rebase.

dbaty:
There are no tests. The description says that it fixes a pull request and not the underlying issue. The author did not follow the contributing guidelines. The author opened another LLM-generated pull request that was obviously wrong (see https://github.com/dbcli/pgcli/pull/1578#issuecomment-4439047042).

They will probably never return and fix this. Let's not waste time and close this.

PR #1580, opened 2026-04-23
Hide timezone notice when local and server timezones are the same

### LINKED ISSUE #1579: Skip time zone warning when local and server are the same
This warning is unnecessary when starting `pgcli`:

```
Using local time zone Europe/London (server uses Europe/London)
Use `set time zone <TZ>` to override, or set `use_local_timezone = False` in the config
```

Especially since the first line is in green, drawing attention to itself over everything else there.

It is possible to get rid of it by setting `use_local_timezone = False`, but I don't want to do that: if I ever happen to connect to a database which uses a different time zone, then I want to keep using local times.

Would it be possibly simply to skip displaying this when the local and server time zones are the same as each other, so there's nothing to warn about? Thanks.

dbaty:
## Description

This takes space and is not very useful.

Now the notice is only printed when the local timezone is different from the server timezone, for example:

    Using local time zone Europe/Paris (server uses Etc/UTC)
    Use `set time zone <TZ>` to override, or set `use_local_timezone = False` in the config

Fixes #1579.

## Demo

When local and server timezones are different:

```
$ pgcli
Using local time zone Europe/Paris (server uses Etc/UTC)
Use `set time zone <TZ>` to override, or set `use_local_timezone = False` in the config
Server: PostgreSQL 17.5 (Debian 17.5-1.pgdg110+1)
[...]
```

When timezones are the same:

```
$ pgcli
Server: PostgreSQL 17.5 (Debian 17.5-1.pgdg110+1)
[...]
```

## Checklist
- [ ] Please squash merge this pull request (uncheck if you'd like us to merge as multiple commits)

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -11,6 +11,7 @@ Features:
 Bug fixes:
 ----------
 * Add `VERSION` to built-in function completion so `SELECT VERSION();` is suggested.
+* Hide timezone notice at startup when local and server timezones are the same.
 
 4.4.0 (2025-12-24)
 ==================
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -1646,7 +1646,7 @@ def cli(
 
             if local_tz is None:
                 echo_error("No local time zone configuration found\n")
-            else:
+            elif local_tz != server_tz:
                 click.secho(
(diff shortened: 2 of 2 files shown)


j-bennet:
Nice!

PR #1599, opened 2026-06-01
Suggest columns after GROUP BY

Kynde:
## What

Suggest columns after `GROUP BY`, the same way pgcli already does for `ORDER BY`.

## Why

The completion dispatch in `suggest_based_on_last_token` offers column suggestions for `SELECT`, `WHERE`, `HAVING`, `ORDER BY` and `DISTINCT`, but `GROUP BY` was missing from that keyword tuple. As a result, a query like:

```sql
SELECT mfa_method, count(*) FROM companies GROUP BY <tab>
```

fell through to the generic keyword handling and ultimately re-derived its suggestions from the `FROM` clause — offering tables/schemas instead of the table's columns. `ORDER BY`, `HAVING` and `WHERE` in the same position all correctly suggest columns.

This appears to have been the case in every release I checked (3.5.0 through current `main`); it's a long-standing omission rather than a recent regression.

## Change

- Add `"group by"` to the keyword tuple in `pgcli/packages/sqlcompletion.py` so it routes through `_suggest_expression`, exactly like `ORDER BY`.
- Extend the existing `test_distinct_and_order_by_suggestions_with_aliases` and `..._with_alias_given` parametrized tests with `GROUP BY` cases (both bare `GROUP BY ` and qualified `GROUP BY x.`).
- Add a changelog entry.

## Testing

- `pytest tests/` → 2699 passed, 1 xfailed, 1 xpassed (pre-existing).
- `ruff format --check` and `ruff check` clean on the changed files.


j-bennet:
Thanks! 🍓 Merging.

PR #1614, opened 2026-08-15
Detect an unconditional UPDATE with sqlparse instead of splitting on whitespace

VXNCXNX:
## Description

With `destructive_warning` covering `unconditional_update`, an UPDATE with no WHERE clause skips the confirmation prompt if any string literal in it contains the standalone word "where".

## Repro

Run on this checkout, not reasoned about:

```
$ python -c "
from pgcli.packages.parseutils import is_destructive
K = ['unconditional_update']
print(is_destructive(\"update accounts set balance = 0\", K))
print(is_destructive(\"update accounts set note = 'no where clause here'\", K))"
True
False
```

The second one is an unconditional UPDATE. It returns False, so the prompt never appears.

## The fix

`query_is_unconditional_update` did `formatted_sql.split()` and checked `"where" not in tokens`. Splitting on whitespace turns the contents of a string literal into tokens, so `'no where clause here'` yields a bare `where`.

It now uses sqlparse's parse tree, which the module already imports: confirm the first meaningful token is the DML keyword UPDATE, then look for a `sqlparse.sql.Where` among the statement's top-level tokens. Since sqlparse handles case and comments itself, the call site passes the raw query rather than the lowercased comment-stripped one. `query_starts_with` is untouched, it serves the separate keyword-prefix path and is not affected.

A WHERE inside a subquery is nested under the Parenthesis, not a top-level token, so `update t set c = (select x from y where z = 1)` correctly still warns.

## Verification

Checked against these, all as expected:

```
update t set c = 0                                  True
update t set n = 'no where clause here'             True
update t set c = 'nowhere'                          True
UPDATE t SET c = 1                                  True
/* where */ update t set c = 1                      True
update t set c = 1 -- where id = 1                  True
update t set c = (select x from y where z = 1)      True
update t set c = 0 where id = 1                     False
UPDATE  t  SET  c = 1  WhErE  id = 1                False
select * from t                                     False
```

Added a parametrized case list to `tests/parseutils/test_parseutils.py` next to the existing `test_is_destructive`. It fails against the old implementation on the string-literal and subquery cases. `pytest tests/parseutils/` passes, 115 tests.

One thing I did not change: `with x as (...) update t set c = 1` returns False, because the first token is the CTE rather than the DML keyword. That is the same as before this patch, so it is a separate gap rather than a regression, and I left it alone.

## Checklist

*Apologies for not using the template on the original description; I have restructured it above and added the changelog and AUTHORS entries that were genuinely missing. Disclosure: written with AI assistance (Claude Code); I ran the reproduction and the tests myself.*


DiegoDAF:
Nice catch, and thank you for fixing it!

I can confirm the bug and the fix from my fork. Against current main, with
`destructive_warning = drop, unconditional_update`:

```
update t set a = 'x where y', b = 2   ->  no warning   (should warn)
```

The bare `where` inside the string literal makes the whitespace split believe
there is a WHERE clause, so an unconditional UPDATE runs with no confirmation.
That is the bad direction for a safety check to fail in.

I cherry-picked your commit and re-ran the same cases:

| statement | warns |
|---|---|
| `update t set x = 1` | yes |
| `update t set x = 1 where id = 2` | no |
| `update t set a = 'x where y', b = 2` | yes (was the bug) |
| `update t set x = 1 -- where` | yes |
| `update t set x = (select 1 from y where z = 1)` | yes |
| `select * from t` | no |

The subquery case is a nice bonus of using the parse tree: a WHERE that only
exists inside a subquery is no longer mistaken for a top-level one, so the
statement is still flagged as unconditional.

Full suite green on my side too. Hope this helps it along.

VXNCXNX:
Rebased onto current `main`. The only conflict was `AUTHORS`, where another entry landed on the same line; both are kept.

On the new base: `pytest tests/parseutils` -> 115 passed, 1 xpassed. Reverting `parseutils/__init__.py` to `main` fails the two new cases as assertion failures, the string-literal one and the subquery one, so the coverage is real:

```
FAILED test_is_destructive_unconditional_update_string_literal[update accounts set note = 'no where clause here'-True]
FAILED test_is_destructive_unconditional_update_string_literal[update t set c = (select x from y where z = 1)-True]
```

VXNCXNX:
Rebased again onto current `main`. Conflicts were both in list files, `AUTHORS` and the `changelog.rst` bug-fix block; all entries kept.

Since `main` has now widened the pin to allow `sqlparse` 0.6.x and this change reads the sqlparse parse tree, I ran the suite against both ends of the range:

```
sqlparse 0.5.5 -> 115 passed, 1 xpassed
sqlparse 0.6.0 -> 115 passed, 1 xpassed
```

`token_first(skip_cm=True)`, `tokens.DML` and `sql.Where` behave the same on both.

VXNCXNX:
Rebased onto `924e7d4`. The branch was 13 commits behind, and the red check was from that stale base.

The failure was the `edit sql in file with external editor` behave scenario, which times out after 2s waiting for `Entering Ex mode.` from the external editor. Nothing to do with this change, which only touches `parseutils`. `main`'s own CI has been green since then.

On the new base, locally:

```
pytest tests/parseutils/     115 passed
pytest tests/ (unit)        2656 passed, 131 skipped, 1 xfailed, 1 xpassed
```

One caveat worth stating rather than leaving you to notice it: the push moved both workflows into `action_required`, so there is now no CI signal at all until someone approves the run. That is arguably worse to look at than the old red, though the old red was measuring a base from four weeks ago. Approving the run should give a clean result.

Four files, and the behaviour change is confined to `is_destructive`.

j-bennet:
Thank you @VXNCXNX. 🍒

PR #1616, opened 2026-08-18
Honor PSQL_EDITOR when opening the external editor

### LINKED ISSUE #1398: [Feature Request] Respect `PSQL_EDITOR` environment variable
The `psql` command supports using a `PSQL_EDITOR` environment variable in place of `EDITOR`. This is useful if you want to use a specific editor for SQL that differs from your normal terminal editor. It would be a nice addition if pgcli supported this environment variable as well.

ChrisJr404:
## Description

Closes #1398. psql lets you point `\e` at a dedicated editor via `PSQL_EDITOR` so it doesn't have to be whatever `EDITOR`/`VISUAL` you use everywhere else, and pgcli didn't respect it.

pgcli hands editing off to `pgspecial.open_external_editor`, which calls `click.edit()` and lets click resolve the editor from `VISUAL`/`EDITOR`. That function already takes an `editor=` override, we just weren't passing anything. I added a small `get_editor()` helper that mirrors psql's lookup order (`PSQL_EDITOR`, then `EDITOR`, then `VISUAL`) and threaded its result through both editor call sites (`\e`/`\ev`/`\ef` and `\ne`). When none of those are set it returns `None`, which keeps click's existing platform-default behaviour, so nobody who wasn't setting `PSQL_EDITOR` sees a change.

Test-wise I added `test_get_editor_precedence` asserting `PSQL_EDITOR` wins over `EDITOR`/`VISUAL`, that it falls back to `EDITOR` then `VISUAL`, and returns `None` when the environment is empty. Ran `python -m pytest tests/test_main.py -k "editor or named_query"` (2 passed) and the full `tests/test_main.py` (60 passed, 10 db skips). `ruff check` and `ruff format --diff` are both clean.

## Checklist
- [x] I verified that my changes work as expected.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -152,6 +152,7 @@ Contributors:
     * Shayan Golshani (shgol)
     * Tommi Kyntölä (kynde)
     * Diego
+    * Chris (ChrisJr404)
 
 Creator:
 --------
--- a/changelog.rst
+++ b/changelog.rst
@@ -10,6 +10,9 @@ Bug fixes:
 
 Features:
 ---------
+* Honor the ``PSQL_EDITOR`` environment variable when opening the external
+  editor (``\\e``, ``\\ev``, ``\\ef``, ``\\ne``), matching psql's precedence of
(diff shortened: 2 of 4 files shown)


dbaty (approved):
Looks good, thank you! :)

Issue #1489, opened 2025-01-13
alias dsn is not detected


ShahriarKh:
## Description
I edited `pgclirc` file and added a new alias_dsn. However, when I run `pgcli --list-dsn` it doesn't return anything. 
this is from my config file:
```
[alias_dsn]
pg = postgresql://user:pass@localhost:1234
```

Using `pgcli -D <alias>` doesn't work either:
```bash
Could not find a DSN with alias pg. Please check the "[alias_dsn]" section in pgclirc.
```

To see if config file is being read, I added a random thing inside it and it threw an error (so it is okay):
```
[alias_dsn]
pg = postgresql://user:pass@localhost:1234

bbb
```
```bash
Traceback (most recent call last):
...
...
configobj.ParseError: Invalid line ('bbb') (matched as neither section nor keyword) at line 237.
```

I also tried restarting terminal, using another shell, and restarting system; but none of them worked.

## Your environment

- Windows 11 22631.4602
- pgcli 4.1.0
- pipfreeze output:
```
aioboto3==12.4.0
aiobotocore==2.12.3
aiohttp==3.9.5
aioitertools==0.11.0
aiosignal==1.3.1
ansicon==1.89.0
anyio==4.3.0
appdirs==1.4.4
argon2-cffi==23.1.0
argon2-cffi-bindings==21.2.0
arrow==1.3.0
asttokens==2.4.1
async-lru==2.0.4
attrs==23.2.0
Babel==2.14.0
beautifulsoup4==4.12.3
bleach==6.1.0
blessed==1.20.0
blessings==1.7
boto3==1.34.69
botocore==1.34.69
bottle==0.12.25
bottle-websocket==0.2.9
certifi==2024.2.2
cffi==1.16.0
chardet==5.2.0
charset-normalizer==3.3.2
cli_helpers==2.3.1
click==8.1.7
colorama==0.4.6
comm==0.2.2
configobj==5.0.9
contourpy==1.2.1
cycler==0.12.1
debugpy==1.8.1
decorator==5.1.1
defusedxml==0.7.1
diff_cover==9.2.0
Eel==0.16.0
executing==2.0.1
Faker==30.3.0
faker-commerce==1.0.4
fastjsonschema==2.19.1
fonttools==4.51.0
fqdn==1.5.1
frozenlist==1.4.1
future==1.0.0
gevent==24.2.1
gevent-websocket==0.10.1
greenlet==3.0.3
h11==0.14.0
httpcore==1.0.5
[...]

PR #1617, opened 2026-08-18
Fix alias_dsn detection for --list-dsn and -D

ChrisJr404:
## Description

`--list-dsn` and `-D`/`--dsn` read the config with a bare `load_config(pgclirc, config_full_path)` that skips the packaged default `pgclirc`, unlike everything else in pgcli which goes through `get_config`. That's the root cause of #1489.

Two ways it bites:

- `--list-dsn` runs before `PGCli()` is constructed, so on a fresh install the config file hasn't been written yet. `load_config` then reads a non-existent file, `cfg["alias_dsn"]` raises `KeyError`, and the broad `except Exception` swallows it and prints `Invalid DSNs found in the config file` with exit 1. No template gets written either, so there's nothing for the user to edit. To someone who did add an alias it just looks like the section is being ignored.
- `-D <alias>` has the same bare read plus a `cfg["main"]` access further down (timezone handling). If the user's config only contains an `[alias_dsn]` block, `[main]` isn't there and it blows up after resolving the alias.

Switching both call sites to `get_config(pgclirc)` fixes it: `get_config` writes the default template when the file is missing and merges the packaged default, so the expected sections (`[main]`, `[alias_dsn]`) are always present and aliases resolve the same way they do everywhere else.

I couldn't reproduce the original report on a config that already has a populated `[alias_dsn]` (that path works), but the fresh/minimal-config cases above are reproducible and match the reported symptoms.

Tests in `tests/test_alias_dsn.py`: fresh config now lists nothing and exits 0 instead of erroring; populated `[alias_dsn]` is listed; `-D <alias>` resolves to the right connection; `-D <unknown>` still errors. The fresh-config and `-D`-resolve tests fail on `main` and pass with this change.

## Checklist


tests/test_alias_dsn.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def write_config(tmp_path, body):
+    cfg = tmp_path / "config"
+    cfg.write_text(body)
+    return cfg
+
+
+@pytest.fixture
+def isolate_config(monkeypatch, tmp_path):
```
j-bennet, line 14: Nice!

j-bennet:
@ChrisJr404 Nice work! The fix looks beautifully minimal. But I would suggest a different approach.

When the user wants to `--list-dsn`, it's supposed to be a lightweight command that only reads the config. If we encounter no config file, why do the extra work to write one? It's already clear that there's nothing to list. I'd add special handling for `KeyError` (or whatever the error raised) instead.

With the `-D <alias>`, the case is not as clear cut. But if the file exists, we shouldn't need to write it. The user might prefer a minimal file with just an `[alias_dsn]` section, and I think we should tolerate having no `[main]` section.

What do you think?

ChrisJr404:
That makes sense, and I agree the current approach does more work than it should. I'll rework it so:

- `--list-dsn` treats a missing config as simply "nothing to list" (handling the `KeyError`/missing-section case) instead of writing a fresh config just to read it back.
- `-D <alias>` reads from the existing file when present and does not rewrite it; it only needs to resolve the one `[alias_dsn]` entry.

That keeps both paths read-only in the common case. I'll push an update along those lines.

ChrisJr404:
Reworked both paths to stay read-only, as suggested:

- `--list-dsn` no longer writes a config just to read it back. If the config file does not exist, or has no `[alias_dsn]` section, it reports nothing and exits cleanly (using `cfg.get("alias_dsn", {})`).
- `-D`/`--dsn` now resolves the alias from the config `PGCli()` already loaded at startup instead of reading the file a second time.

Added an assertion to the fresh-config test that no config file is written as a side effect, and updated the changelog. The style-check failure was a stray `ruff format` diff in the test file, which is fixed too. `ruff check`, `ruff format`, and the alias/main tests all pass locally.

j-bennet:
Thanks @ChrisJr404 looks great! 🍒 Merging.

PR #1609, opened 2026-06-29
Add \ne to edit a named query in the external editor

DiegoDAF:
## Summary

Adds `\ne <name>`, a special command to edit a named query in the external editor (`$EDITOR`).

It loads the named query's SQL into the editor and, on save, writes the edited result back to the `[named queries]` section of the config (creating the entry if it does not exist). This complements `\ns` (save) by making it practical to edit longer named queries without retyping them, mirroring how `\e` / `\ev` / `\ef` already open the editor.

Addresses the editor part of #1430.

## Behavior
- `\ne foo` on an existing query: opens it in `$EDITOR`; saving updates it.
- `\ne foo` on a new name: opens an empty editor; saving creates it.
- Empty result or no changes: nothing is saved (reported in the status).

## Tests
- `tests/test_main.py::test_edit_named_query` covers create / update / no-change / empty / editor-error / usage (editor mocked, no DB needed).

## Changelog
- Added an entry under a new `Upcoming` section.

PR #1619, opened 2026-08-18
Raise the external-editor behave timeouts to reduce CI flakiness

DiegoDAF:
## Description

The `edit sql in file with external editor` scenario in
`tests/features/iocommands.feature` errors intermittently on CI. It is a timing
issue, not a real failure: `expect_exact` waits only 2 seconds for the ex-mode
banner (and 1 second per token when checking the resulting prompt), which is
not always enough on a loaded runner.

When it expires, behave reports the scenario as an **error** and fail-fast then
cancels the sibling matrix jobs, so a single slow wait looks like a systematic
breakage across Python versions.

Recent examples: it hit #1543 on 3.10, #1544 and #1609 on other versions (both
went green on a later run with no code change), and my own fork's CI on commits
that touch nothing near the editor code. @j-bennet asked about exactly this in
#1543.

## Change

Raise the `expect_exact` timeouts in `tests/features/steps/iocommands.py` to 10
seconds (from 2s, 1s and 5s).

This does not slow down passing runs: pexpect returns as soon as the expected
text appears, so the timeout only bounds how long a *failing* wait blocks
before behave gives up.

## Validation

`behave features/iocommands.feature` green locally (2 scenarios, 12 steps), and
the full behave suite green as well (15 features, 61 scenarios, 0 failed). The
scenario has been stable in my fork since the change.

## Checklist

---

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -1,6 +1,15 @@
 Upcoming (TBD)
 ==============
 
+Internal:
+---------
+* Make the external-editor behave scenario less flaky: the ``expect_exact``
+  timeouts in ``tests/features/steps/iocommands.py`` were as low as 1-2
+  seconds, which intermittently expired on loaded CI runners and reported
+  ``Scenario: edit sql in file with external editor`` as an error. Raised to 10
+  seconds; passing runs are unaffected because pexpect returns as soon as the
+  expected text appears.
+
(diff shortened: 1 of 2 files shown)


j-bennet:
Looks safe enough, let's try it.

PR #1621, opened 2026-08-18
Keep the connection string when using -l/--list or --ping

DiegoDAF:
## Description

A connection string is silently discarded when `-l`/`--list` or `--ping` is
used, and pgcli then tries a local socket connection as the OS user:

```
$ pgcli "postgresql://someuser@somehost:5432/somedb?sslmode=verify-ca"
someuser@somehost:somedb>            # connects fine

$ pgcli "postgresql://someuser@somehost:5432/somedb?sslmode=verify-ca" -l
connection failed: connection to server on socket "/var/run/postgresql/.s.PGSQL.5432"
failed: FATAL:  role "<osuser>" does not exist
```

The cause is in `cli()`:

```python
# because option --ping, --list or -l are not supposed to have a db name
if list_databases or ping_database:
    database = "postgres"
```

The intent is right for a plain db name, but the positional argument may also
be a connection string, and that carries the entire connection: host, user,
port, sslmode, sslrootcert. Replacing it with `"postgres"` throws all of that
away, so the branch dispatcher below falls through to
`pgcli.connect("postgres", host, user, port)` with empty host and user.

The same applies to a `key=value` conninfo string, which is the only way to
pass parameters like `sslmode` when using the `-h`/`-U` form.

## Change

Discard the positional argument only when it is a plain database name. A
connection string is passed through untouched.

One extra case: if the connection string names no database, `"postgres"` is
merged in for the listing. libpq would otherwise default the database to the
user name, which is rarely an existing database, so
`pgcli "host=h user=u sslmode=verify-ca" -l` would fail for a different reason.

## Validation

Five new tests in `tests/test_main.py`: URI preserved with `-l`, `key=value`
conninfo preserved with `-l`, URI preserved with `--ping`, the missing-dbname
fallback (checking the rest of the conninfo survives), and a plain db name
still being discarded. Four of the five fail on current main.

Verified end to end against a local server for all three paths (URI, conninfo
with and without a dbname, and no connection string at all). Full suite green
locally (2735 passed).

## Checklist

---


pgcli/main.py
```diff
@@ -1682,9 +1682,23 @@ def cli(
...
     if list_databases or ping_database:
-        database = "postgres"
+        if not is_conn_string:
+            database = "postgres"
+        else:
+            try:
+                if not conninfo_to_dict(database).get("dbname"):
+                    database = make_conninfo(database, dbname="postgres")
+            except Exception:
+                pass  # invalid conninfo: let the connection attempt report it
```
j-bennet: Perhaps we can simplify the condition here. We only want to override the database name with `postgres` when the user didn't provide one, right? How about this:

```python
        if not database:
            database = "postgres"
        elif is_conn_string and not conninfo_to_dict(database).get("dbname"):
            database = make_conninfo(database, dbname="postgres")
```

DiegoDAF: Agreed, and applied, with one interesting confirmation: your version is actually closer to psql than mine was. I checked against real psql: `psql -l nonexistentdb` fails with "database does not exist", so psql connects to the named database and lists from there rather than discarding it. The test now documents that.

Two small deltas from your snippet:

- Kept the `try/except` around `conninfo_to_dict`, so a malformed connection string is reported by the connection attempt instead of a traceback.
- `elif is_conn_string and not conninfo_to_dict(...)` stayed as a nested `if` inside the `elif is_conn_string`, purely for the except to wrap only the parse.

Changelog updated to match the new behavior.

j-bennet:
Thanks @DiegoDAF , merging.

PR #1622, opened 2026-08-19
Add --timeout option and a connect_timeout config default

DiegoDAF:
## Description

pgcli never sets a connection timeout, so it inherits libpq's default of `0`,
which means "wait until the operating system gives up on the TCP connection".
That can take minutes with no feedback.

Measured against `192.0.2.1` (TEST-NET, silently drops SYNs):

```
$ time pgcli "postgresql://u@192.0.2.1:5432/db" -c "select 1"
   ... still waiting after 30s, killed manually
```

That is a poor default for an interactive client: an unreachable host, a
bastion that is down or a wrong port all look like a hang.

## Change

Adds a `--timeout` command line option and a `connect_timeout` config value
(default 30 seconds). Precedence, highest first:

1. `--timeout`
2. `connect_timeout` in the connection string
3. `$PGCONNECT_TIMEOUT`
4. the `connect_timeout` config value

Points 2 and 3 keep libpq's own ordering, and when only the environment
variable is set, nothing is injected so libpq reads it itself. `--timeout 0`
is meaningful (wait forever, the previous behaviour) and is not treated as
unset, so nobody is locked out of the old default.

The resolved value is passed as a connection parameter rather than merged into
the dsn, so the user's connection string reaches `PGExecute` exactly as
written. `PGExecute` keeps `connect_timeout` alongside the dsn the same way it
already keeps `hostaddr`.

## Validation

Seven new tests in `tests/test_main.py` covering every precedence combination:
config default, config value, connection string over config, connection string
over env, env left to libpq, `--timeout` over everything, and `--timeout 0`.

Two existing tests (`test_pg_service_file`, `test_application_name_db_uri`)
assert the exact `PGExecute` call and now include the resolved timeout.

Measured end to end against the same unreachable address:

| setting | time to fail |
|---|---|
| config default 30 | 30s |
| `connect_timeout=5` in the connection string | 5s |
| `PGCONNECT_TIMEOUT=4` | 5s |
| `--timeout 2` (with 20 in the string and 9 in the env) | 2s |

Full suite green locally (2737 passed).

## Checklist

---

Part of the feature list in discussion #1603: this is **item 23**.

### DIFF
--- a/pgcli/main.py
+++ b/pgcli/main.py
@@ -157,6 +157,35 @@ def get_editor():
     return os.environ.get("PSQL_EDITOR") or os.environ.get("EDITOR") or os.environ.get("VISUAL") or None
 
 
+def get_connect_timeout(explicit, dsn, kwargs, default):
+    """Pick the connection timeout to apply, in seconds.
+
+    Precedence, highest first:
+
+    1. ``explicit``, i.e. ``--timeout`` on the command line
+    2. ``connect_timeout`` in the connection string, or in ``kwargs``
+    3. ``$PGCONNECT_TIMEOUT``
+    4. ``default``, the ``connect_timeout`` config value
+
+    Returns ``None`` when the user already stated a timeout by one of the
+    means we must not override, in which case the caller leaves the
+    connection parameters alone and libpq reads it from where it already is.
+
+    A default matters because libpq's own is 0, which waits until the
+    operating system gives up on the TCP connection, so an unreachable host
+    hangs for minutes.
+    """
+    if explicit is not None:
+        return explicit
+    if "connect_timeout" in kwargs:
+        return None
+    if dsn and "connect_timeout" in conninfo_to_dict(dsn):
+        return None
+    if os.environ.get("PGCONNECT_TIMEOUT"):
+        return None
+    return default
+
+
 class PGCli:
     default_prompt = "\\u@\\h:\\d> "
     max_len_prompt = 30
@@ -266,7 +295,7 @@ class PGCli:
         self.on_error = c["main"]["on_error"].upper()
         # Connection timeout, in seconds. See connect() for the precedence.
         self.connect_timeout = connect_timeout
-        self.default_connect_timeout = c["main"].get("connect_timeout", "30")
+        self.default_connect_timeout = c["main"].as_int("connect_timeout")
         self.decimal_format = c["data_formats"]["decimal"]
         self.float_format = c["data_formats"]["float"]
         self.column_date_formats = c["column_date_formats"]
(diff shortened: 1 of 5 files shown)


pgcli/main.py
```diff
@@ -676,6 +680,26 @@ def connect(self, database="", host="", user="", port="", passwd="", dsn="", **k
...
+        # libpq's own default is 0, which waits until the operating system gives
+        # up on the TCP connection, so an unreachable host hangs for minutes.
+        # The resolved value is passed as a connection parameter rather than
+        # merged into the dsn, leaving the user's connection string untouched.
+        timeout = self.connect_timeout
+        if timeout is None:
+            in_dsn = "connect_timeout" in conninfo_to_dict(dsn) if dsn else False
+            if not in_dsn and "connect_timeout" not in kwargs and not os.environ.get("PGCONNECT_TIMEOUT"):
+                try:
+                    timeout = int(self.default_connect_timeout)
```
dbaty: I'd rather see this conversion when setting `self.default_connect_timeout` in `__init__()`. That way, all uses of the attribute can be sure that it's an integer (and not garbage). We could use this:

```pyton
self.default_connect_timeout = c["main"].as_int("default_connect_timeout")
```

That will (1) convert to an integer; (2) automatically default to 30 (from `pgclirc`) if the user removed the line from their configuration file; and (3) raise an error if the user configured a non-integer value, which I find useful: I prefer when the software says I have done something stupid, instead of silently ignoring it, and making me search and finally find out that, yes, I did something stupid. ;)

pgcli/pgexecute.py
```diff
@@ -216,7 +216,7 @@ def connect(
 
         if new_params["dsn"]:
             # When using DSN, only keep dsn, password, and hostaddr (for SSH tunnels)
-            new_params = {k: v for k, v in new_params.items() if k in ("dsn", "password", "hostaddr")}
+            new_params = {k: v for k, v in new_params.items() if k in ("dsn", "password", "hostaddr", "connect_timeout")}
```
dbaty, line 221: The comment above is now out-of-date.

tests/test_main.py
```diff
@@ -701,3 +703,62 @@ def test_get_editor_precedence():
     # Nothing set -> None, so click uses its platform default.
     with mock.patch.dict(os.environ, {}, clear=True):
         assert get_editor() is None
+
+
+def _effective_connect_timeout(tmpdir, cli_timeout=None, dsn_timeout=None, env=None, cfgval=None):
```
dbaty, line 709: I guess it works... but a `get_connect_timeout(dsn, default, kwargs)` helper method would probably be easier to unit-test. What do you think?

dbaty (commented):
Nice addition, thanks!

DiegoDAF:
Thanks for the review! All three applied.

1. `as_int("connect_timeout")` in `__init__`, so the attribute is always an int and a typo in the config is reported instead of silently ignored. Added a test for that.
2. Extracted `get_connect_timeout(explicit, dsn, kwargs, default)`. The precedence table is now a parametrized unit test calling it directly, and two end-to-end tests stay to prove it is wired into `connect()`.
3. Comment updated.

j-bennet:
LGTM. Let's merge. Thanks @DiegoDAF and @dbaty !

PR #1625, opened 2026-08-28
Remove stale xfail on schema-qualified INSERT parseutils test

pacocartones:
## Description

`test_simple_insert_single_table_schema_qualified` in
`tests/parseutils/test_parseutils.py` was marked `@pytest.mark.xfail` back when an
older `sqlparse` mislabeled schema-qualified `INSERT` statements. That has long
been fixed. The test now passes across the whole supported range
(`sqlparse >=0.3.0,<0.7` — I checked 0.3.0 and 0.6.0), so the marker only ever
produced an XPASS.

Since `xfail_strict` isn't set, that XPASS was silent, which means the assertion
wasn't actually guarding `extract_tables` anymore. This just drops the decorator
so the test does its job again. No production code changes.

Verified locally: the file passes repeatedly on both sqlparse 0.3.0 and 0.6.0,
and `ruff format`/`ruff check` are clean.

## Checklist

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -9,6 +9,11 @@ Internal:
   ``Scenario: edit sql in file with external editor`` as an error. Raised to 10
   seconds; passing runs are unaffected because pexpect returns as soon as the
   expected text appears.
+* Remove a stale ``@pytest.mark.xfail`` on
+  ``test_simple_insert_single_table_schema_qualified``. It was marked for an old
+  ``sqlparse`` that mislabeled schema-qualified ``INSERT``; the test now passes
+  across the supported ``sqlparse`` range (0.3.0 to 0.6.x), so the marker only
+  hid a passing test (an unreported XPASS, since ``xfail_strict`` is not set).
 
 Bug fixes:
 ----------
--- a/changelog.rst
+++ b/changelog.rst
@@ -14,6 +14,13 @@ Internal:
   ``sqlparse`` that mislabeled schema-qualified ``INSERT``; the test now passes
   across the supported ``sqlparse`` range (0.3.0 to 0.6.x), so the marker only
   hid a passing test (an unreported XPASS, since ``xfail_strict`` is not set).
+* Document the still-needed ``@pytest.mark.xfail`` on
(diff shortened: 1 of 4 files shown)


changelog.rst
```diff
@@ -9,6 +9,11 @@ Internal:
   ``Scenario: edit sql in file with external editor`` as an error. Raised to 10
   seconds; passing runs are unaffected because pexpect returns as soon as the
   expected text appears.
+* Remove a stale ``@pytest.mark.xfail`` on
+  ``test_simple_insert_single_table_schema_qualified``. It was marked for an old
+  ``sqlparse`` that mislabeled schema-qualified ``INSERT``; the test now passes
+  across the supported ``sqlparse`` range (0.3.0 to 0.6.x), so the marker only
+  hid a passing test (an unreported XPASS, since ``xfail_strict`` is not set).
```
dbaty, line 16: Please remove this. The changelog is for users of pgcli, who are unlikely to feel concerned and interested by this change.

Also, I am not sure that this warrants an addition in `AUTHORS`. Unless you really want your name there, of course...

dbaty:
Also, there is another `xfail` on `tests.test_sqlcompletion.test_sub_select_multiple_col_name_completion()`. It's been added by 4e862015f8df1ce62084a8d04d2c13a9e2815fe3 in 2015 (without explanation, unfortunately). Care to check if it's still needed? :)

pacocartones:
Good catch — thanks for pointing at 4e862015. I checked empirically: it's still needed. Dropping the marker makes `test_sub_select_multiple_col_name_completion` fail on sqlparse 0.6.0 — parsing the incomplete `SELECT a, FROM abc` still treats the token before the trailing comma (`a`) as a table, so the Column suggestion comes back with `table_refs` for both `a` and `abc` instead of just `abc` (and `last_keyword` differs too). So rather than remove it, I've added a `reason=` to the marker documenting the parser limitation — that closes the 2015 "no explanation" gap while keeping the expected-failure recorded. Pushed to this branch; ready for another look when you have a moment.

PR #1631, opened 2026-09-08
Preserve literal passwords in PostgreSQL service files

### LINKED ISSUE #1512: pgcli couldn't parse password that contained character '#'

jackwalkerlabs:
Service-file passwords containing hashes, percent signs, commas, or quote characters now reach PostgreSQL unchanged, matching libpq and psql. The parser preserves leading comments and original error line numbers, and unknown service names retain the existing error-handling behavior.

Fixes #1512. This is a breaking change for values previously quoted only to escape the old parser: those quotes must be removed. The changelog now states that explicitly. The parser comment and AUTHORS entry follow the maintainer's requested wording changes. Main configuration parsing is handled separately in #1635.

Validation of current head `7726fceee3b25b4bcd65a5ec1f44fe7779d9829c`: independent Factory source/UX review approved; isolated offline Python 3.11.2 verification passed the configuration/prompt and selected startup tests, eight service-file test cases, 16 configuration compatibility checks, and Ruff lint/format. Full live-PostgreSQL integration and the supported-Python matrix were not rerun for this revision.

Factory run: `FC-20260913-EEB554`. The original PR and branch were retained, with a descendant commit and current upstream changes merged in.

AI disclosure: The original implementation used Codex. This revision was implemented, independently reviewed, and verified by hosted FactoryChief using Codex under the account owner's authorization; no human code review is claimed.

## Checklist
- [x] Added changelog entry.
- [x] Added submitting account to AUTHORS.
- [x] Installed pre-commit hooks for the original contribution; current revision passed Ruff lint and formatting directly.
- [x] Verified behavior and ran the checks listed above.
- [x] Please squash merge.


pgcli/main.py
```diff
@@ -2213,14 +2213,13 @@ def parse_service_info(service):
         return None, service_file
     with open(service_file, newline="") as f:
         skipped_lines = skip_initial_comment(f)
-        try:
-            service_file_config = ConfigObj(f)
-        except ParseError as err:
-            err.line_number += skipped_lines
-            raise err
+        # libpq treats values literally: hashes, commas, quotes and percent
+        # signs are part of the value, not ConfigObj comments or syntax.
```
dbaty, line 2217: It makes sense in the diff, but it looks odd to reference `ConfigObj` in a comment here while the module makes not reference to it at all. I'd remove that part.


```suggestion
        # signs are part of the value.
```

AUTHORS
```diff
@@ -155,6 +155,8 @@ Contributors:
     * Chris (ChrisJr404)
     * Pieter Ouwerkerk (pouwerkerk)
 
+    * jackwalkerlabs (Codex-assisted)
```
dbaty, line 158: You can remove the empty line above this line. Also, the "Codex-assisted" comment is not warranted. I appreciate that you mentioned it in the pull request description. But here in the changelog it's... you: I suppose that you do have an existence outside Codex. And who knows, maybe you could contribute without Codex, someday! ;)

(I suppose that Codex wrote that comment. That makes me a bit sad, actually.)

changelog.rst
```diff
@@ -22,6 +22,9 @@ Features:
 
 Bug fixes:
 ----------
+* Preserve hashes, percent signs, commas and quotes in service-file passwords,
+  matching libpq. Quotes around service-file values are now literal; remove
+  quotes previously added solely to escape ConfigObj syntax.
```
dbaty, line 27: I would make it clearer that it's a breaking change. Also, I would not reference `ConfigObj`, which is an implementation detail. Suggestion:


```suggestion
* [breaking change] Password in service files (usually ``~/.pg_service.conf``) are now read as they appear, as `libpq` and `psql` do. Previously, hashes, percent signs and other values had to be quoted. They must not be quoted anymore.
```

@j-bennet: it's a (small) breaking change, I think it would warrant a new minor version (from 4.6.0 to 4.7.0). Do you agree?

pgcli/main.py
```diff
@@ -2213,14 +2213,13 @@ def parse_service_info(service):
...
+        service_file_config = ConfigParser(interpolation=None, delimiters=("=",), comment_prefixes=("#",))
+        service_file_config.read_file(itertools.chain(itertools.repeat("\n", skipped_lines), f), source=service_file)
     if service not in service_file_config:
         return None, service_file
-    service_conf = service_file_config.get(service)
+    service_conf = service_file_config[service]
```
dbaty, line 2222: This changes the behaviour: previously, `parse_service_info()` returned `(None, service_file)` upon an unknown service name. The caller would detect it and print an error message. Now this function raises a `KeyError`.

The previous behaviour should be kept.

pgcli/main.py
```diff
@@ -2213,14 +2213,13 @@ def parse_service_info(service):
...
         skipped_lines = skip_initial_comment(f)
-        try:
-            service_file_config = ConfigObj(f)
-        except ParseError as err:
-            err.line_number += skipped_lines
-            raise err
+        # libpq treats values literally: hashes, commas, quotes and percent
+        # signs are part of the value, not ConfigObj comments or syntax.
+        service_file_config = ConfigParser(interpolation=None, delimiters=("=",), comment_prefixes=("#",))
+        service_file_config.read_file(itertools.chain(itertools.repeat("\n", skipped_lines), f), source=service_file)
```
dbaty, line 2219: What's the use of `itertools.repeat("\n", skipped_lines)`? Surely the parser could do without the empty lines, couldn't it?

pgcli/main.py
```diff
@@ -1,5 +1,5 @@
 from zoneinfo import ZoneInfoNotFoundError
-from configobj import ConfigObj, ParseError
+from configparser import ConfigParser
```
dbaty, line 2: I'd love to drop the `configobj` dependency. pgcli still uses it to parse its main configuration file, but we could probably replace that by `configparser` as well. There may be incompatibilities/differences between the two, they would have to be listed first (maybe we can ignore them, maybe we cant'). Are you willing to give it a try? (That should be in a separate pull request.)

DiegoDAF:
I hit this same divergence in my fork and had already measured it against libpq, so here is the data in case it is useful for this PR.

Method: a throwaway cluster, and `dbname` as the probe rather than `password`, because the server echoes it back verbatim in `database "..." does not exist`. That reads out exactly what libpq parsed, with no guessing.

| in the service file | libpq parses | ConfigObj parsed | this PR |
|---|---|---|---|
| `abc#def` | `abc#def` | `abc` | correct |
| `#abc` | `#abc` | `` (empty) | correct |
| `abc #def` | `abc #def` | `abc` | correct |
| `a,b` | `a,b` | `["a", "b"]` (a list) | correct |
| `'quoted'` | `'quoted'` | `quoted` | correct |
| `"quoted"` | `"quoted"` | `quoted` | correct |
| `a b` / `a%b` / `a=b` | unchanged | unchanged | correct |
| `abc   ` | `abc` | `abc` | correct |

So the change matches libpq on all ten cases, including the list one, which I had not expected: a value with a comma became a Python list, not a string.

Two notes from having done the same rewrite:

**On `itertools.repeat("\n", skipped_lines)`:** it is load-bearing, it keeps the reported line numbers aligned with the real file. Without it, a bad line at file line 6 is reported as line 3. The old code did the same thing differently, with `err.line_number += skipped_lines`. With the blank lines fed in, `ConfigParser` produces:

```
Source contains parsing errors: '/home/daf/.pg_service.conf'
	[line  6]: 'not a valid entry\n'
```

which is the correct line.

**On the unknown-service concern:** I could not reproduce a `KeyError`. The `if service not in service_file_config: return None, service_file` guard is still above it, and `ConfigParser.__contains__` returns False for an unknown section, so `parse_service_info("nope")` still returns `(None, path)`. I ran the exact code from this PR to check. Worth a regression test either way, since nothing currently pins that contract.

One real behaviour change that has not been mentioned: a malformed file now raises `configparser.ParsingError` where it used to raise `configobj.ParseError`. Nothing in pgcli catches either, so it only changes the traceback, but it is worth deciding deliberately rather than by accident.

Happy to contribute the parametrized version of that table as a test if you want it, it maps one-to-one onto the rows above.

pgcli/main.py
dbaty: As noted by Diego (or his LLM...), I missed the `if service not in service_file_config` block above. So using `service_file_config[service]` will not raise a `KeyError`.

pgcli/main.py
dbaty: As noted by Diego, the number `skipped_lines` must be kept to make sure that an error shows the correct line number.

PR #1633, opened 2026-09-08
Preserve macOS Keychain access choices

MelvinCERBA:
## Description

On macOS, Keychain trusts the executable that creates a credential by default. Because pgcli runs through Python, saving a password pre-authorizes Python to read it without confirmation. Rewriting an existing credential also resets access choices such as **Allow Once**.

This is a problem because, for example, you typically don't want your coding agent to be able to access DBs you store the credentials for in Keychain without your manual approval. 

This PR preserves the user's Keychain access choices:

- New credentials are created without any pre-authorized application, allowing the user to choose **Allow Once** or **Always Allow** on the next access.
- Existing credentials are updated in place so their access controls remain unchanged.
- Passwords loaded from the keyring are not rewritten after successful authentication.
- A replacement entered after a rejected stored password is still saved.
- Non-macOS platforms and custom keyring backends keep their existing behavior.

Implementation-wise, pgcli preserves keyring's normal backend selection, including chained backends. For native macOS Keychain, it reuses `keyring.backends.macOS.api`: existing items are updated with only `kSecValueData`, preserving their access controls, while new items receive a `SecAccess` created from an empty trusted-application list. The few Security.framework calls not exposed by keyring are bound through `ctypes`.

The implementation supports keyring 23.1 through 25.x. Regression tests cover native backend detection, chained and custom backends, ACL-preserving updates, and creation with no trusted applications.

## Checklist


UsmanGhias (commented):
Hi Melvin, thanks for putting this together! This is a fantastic security improvement that prevents python from unnecessarily pre-authorizing keychain credentials on macOS, protecting users from unintended background access.

Technically, I really like how cleanly you separated the macOS specific logic into `macos_keychain.py` while keeping the fallback path intact for other platforms. One minor thing to double check is the ctypes memory management and error handling across different macOS versions, but the unit tests look quite thorough and cover the core code paths well.

Overall, this is very neat work. Happy to see this land once CI passes.

Best,
Usman

MelvinCERBA:
Thanks for the review, @UsmanGhias! I double-checked the ctypes signatures, reference cleanup, and error paths, and pushed test-only commit 41090c34.

The added coverage checks cleanup after all allocation stages, error statuses from SecAccessCreate/SecItemUpdate/SecItemAdd, and exceptions from those calls. It also verifies cleanup after successful creation and that failed updates do not fall back to creating an item. All 49 targeted tests pass, along with Ruff lint and formatting checks. I found no implementation change necessary in the paths checked.

For compatibility, the declarations match the macOS SDKs I checked (13.3, 14.4, 14.5, and 15.2). Separate native-object checks used the macOS API modules from keyring 23.1.0, 23.13.1, 24.3.1, 25.0.0, and 25.7.0. Runtime validation was limited to macOS 15.0 on Apple Silicon: those checks exercised real CoreFoundation objects and SecAccessCreate, with credential storage calls stubbed. This does not establish end-to-end behaviour on other macOS releases.

The earlier [test run](https://github.com/dbcli/pgcli/actions/runs/34825507816) and [CodeQL run](https://github.com/dbcli/pgcli/actions/runs/34825507707) still show `action_required`; runs for the new commit are not visible yet. Could a maintainer approve the pending fork workflows? The earlier [Codex workflow](https://github.com/dbcli/pgcli/actions/runs/34828928244/job/103927398411) also stopped because my account has read-only access to this repository.

UsmanGhias (approved):

UsmanGhias (approved):

PR #1636, opened 2026-09-14
Fix COPY to STDOUT/from STDIN leaving the connection stuck

### LINKED ISSUE #1505: Use an incorrect copy command syntax copy instead of \copy and cannot quit app.
Use an incorrect copy command syntax copy instead of \copy and cannot quit app after that:

```
Server: PostgreSQL 14.15
Version: 4.3.0
Home: http://pgcli.com
simplifi-live> copy (select id from transcoding_formats order by id) to stdout csv;
COPY cannot be used with this method; use copy() instead
Time: 0.043s
simplifi-live> quit
A transaction is ongoing. Choose `c` to COMMIT, `r` to ROLLBACK, `a` to abort exit. [a]: r^M
```

anandghegde:
## Description

Running `copy ... to stdout` or `copy ... from stdin` (instead of `\copy`) broke the session. psycopg's `cursor.execute()` raises "COPY cannot be used with this method; use copy() instead", but only after the server has already started the COPY. The connection stays in the COPY state, so:

- every later query fails with "sending query failed: another command is already in progress"
- the transaction status stays `ACTIVE`, so `quit` asks about an ongoing transaction, and `c`/`r` fail for the same reason

Fix: in `execute_normal_sql`, when `execute()` raises a `ProgrammingError` and the connection is still `ACTIVE`, end the COPY on the connection. A `COPY FROM STDIN` is aborted with `put_copy_end`, and the rows of a `COPY TO STDOUT` are read and thrown away. Then the error is re-raised with a message that points to `\copy`:

```
postgres@localhost:postgres> copy (select 1) to stdout csv;
COPY to STDOUT or from STDIN is not supported, use \copy instead
Time: 0.004s
postgres@localhost:postgres> select 42;
+----------+
| ?column? |
|----------|
| 42       |
+----------+
SELECT 1
```

If the COPY runs inside `BEGIN`, a `COPY TO STDOUT` leaves the transaction open and usable. An aborted `COPY FROM STDIN` puts it in the failed state, the same as any other error.

How I tested it (Postgres 17 in Docker, psycopg 3.3.5):

- New `dbtest` tests in `tests/test_pgexecute.py` cover `COPY TO STDOUT` (small and 10k rows), `COPY FROM STDIN`, and `COPY TO STDOUT` inside an open transaction. They fail on `main` and pass with this change.
- `pytest tests`: 2783 passed, 1 xfailed, 1 xpassed.
- `behave tests/features`: 33 scenarios passed, 0 failed.
- `ruff check` and `ruff format --diff` are clean. `docutils --halt=warning changelog.rst` is clean.
- Ran pgcli itself with pexpect. On `main`, `copy (select 1) to stdout csv;` followed by `quit` hits the "A transaction is ongoing" prompt, and any query in between fails. With this change, later queries work, `copy ... from stdin` and `copy` inside `begin` behave the same way, `\copy` still works, and `quit` exits right away.

Fixes #1505

## Checklist


pgcli/pgexecute.py
```diff
@@ -471,7 +471,16 @@ def handle_notices(n):
             return title, None, None, res.command_status.decode()
 
         cur = self.conn.cursor()
-        cur.execute(split_sql)
+        try:
+            cur.execute(split_sql)
+        except psycopg.ProgrammingError as e:
```
j-bennet, line 476: Can `psycopg.ProgrammingError` be raised in situations other than `copy`, and if so, how do we handle that?

anandghegde: Yes, and those pass through unchanged. psycopg raises `ProgrammingError` in two broad cases here:

- Server errors with SQLSTATE class 42, which psycopg maps to subclasses such as `SyntaxError` and `UndefinedTable`. By the time these are raised the server has already ended the statement, so the transaction status is `IDLE` or `INERROR`, never `ACTIVE`.
- Client-side misuse (parameter count mismatches and similar). `execute_normal_sql` passes no parameters, and these are raised before anything is sent, so the status is not `ACTIVE` either.

The only result that leaves the connection `ACTIVE` is a `COPY_IN`/`COPY_OUT` status (`_raise_for_result` in psycopg's `_cursor_base.py`). That is why the handler checks `transaction_status` rather than the message, and re-raises the original exception otherwise.

To pin this down I added `test_other_programming_errors_are_not_rewritten` in ec70d50e: a syntax error and an undefined table still raise their own exception classes without the `\copy` message, and the connection stays usable. With the `ACTIVE` check removed, both cases fail. The full suite passes against Postgres 16 (2785 passed).

UsmanGhias (commented):
Hi Anand. This patch solves a tricky session-locking issue with psycopg when users accidentally run raw COPY statements instead of backslash commands. It is great to see robust test coverage added for both STDOUT and STDIN cases.

Technically, catching `psycopg.ProgrammingError` and explicitly checking the transaction status to drain or abort the stream is the right approach to keep the connection usable. One minor thing to double-check is ensuring that large volumes of data drained from stdout will not cause memory spikes, though typical usage here is interactive.

Overall, clean implementation and a great UX improvement for pgcli.

Best,
Usman

anandghegde:
Thanks for looking at it.

On the memory question: the drain doesn't accumulate. `_end_copy` discards each buffer as it arrives and keeps nothing:

```python
while pgconn.get_copy_data(0)[0] > 0:
    pass
```

`PGgetCopyData` hands back one row at a time and the bytes are dropped on the next iteration, so peak memory is one row plus libpq's own connection buffer, no matter how large the table is. Nothing is decoded into Python rows, and no list is built.

It only runs for a `COPY ... TO STDOUT` that psycopg has already rejected, and only to get the connection back out of `ACTIVE` so the next query works. The alternative would be closing and reopening the connection, which loses the session and any open transaction.

The cost is the time to stream the table over the wire, which is why the error still points at `\copy`.

PR #1638, opened 2026-09-17
Fix FunctionMetadata.fields() crash when arg_names is None

### LINKED ISSUE #1204: NoneType exception in get_completions()
## Description
With this SQL:
```sql
CREATE EXTENSION hstore;

CREATE OR REPLACE FUNCTION no_labels() RETURNS hstore AS $$
  SELECT hstore(ARRAY[]::text[])
$$ language sql immutable;

CREATE OR REPLACE FUNCTION labels() RETURNS hstore AS $$
  SELECT no_labels()
$$ language sql immutable;

CREATE OR REPLACE FUNCTION labels(variadic text[]) RETURNS hstore AS $$
  SELECT hstore($1)
$$ language sql immutable;
```

When editing this query:

    pgcli> select * from labels()
           --            ^ start inserting each( here

As soon as the `(` after `each` is inserted, this traceback appears:
```
Unhandled exception in event loop:
  File "/usr/lib/python3.8/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/usr/lib/python3.8/site-packages/prompt_toolkit/eventloop/async_generator.py", line 43, in runner
    for item in get_iterable():
  File "/usr/lib/python3.8/site-packages/prompt_toolkit/completion/base.py", line 226, in <lambda>
    lambda: self.completer.get_completions(document, complete_event)
  File "/usr/lib/python3.8/site-packages/prompt_toolkit/completion/base.py", line 262, in get_completions
    return completer.get_completions(document, complete_event)
  File "/home/ulidtko/.local/lib/python3.8/site-packages/pgcli/pgcompleter.py", line 485, in get_completions
    matches.extend(matcher(self, suggestion, word_before_cursor))
  File "/home/ulidtko/.local/lib/python3.8/site-packages/pgcli/pgcompleter.py", line 506, in get_column_matches
    scoped_cols = self.populate_scoped_cols(tables, suggestion.local_tables)
  File "/home/ulidtko/.local/lib/python3.8/site-packages/pgcli/pgcompleter.py", line 985, in populate_scoped_cols
    cols = func.fields()
  File "/home/ulidtko/.local/lib/python3.8/site-packages/pgcli/packages/parseutils/meta.py", line 168, in fields
    for name, typ, mode in zip(self.arg_names, self.arg_types, self.arg_modes)

Exception 'NoneType' object is not iterable
Press ENTER to continue...
```

## Your environment
  * Arch Linux
  * PostgreSQL 11 from DockerHub, `docker pull postgres:11` 
  * 3.0.0
  * is pretty long; I can provide specific dependency versions on request.

cristianchiriac:
Fixes #1204.

## Bug

Autocompleting a call to a table-returning or variadic function whose parameters have no names (e.g. an unnamed `variadic text[]` argument, as in the reported `hstore`/`labels(variadic text[])` example) crashes the completion thread:

```
File ".../pgcli/pgcompleter.py", line 985, in populate_scoped_cols
    cols = func.fields()
File ".../pgcli/packages/parseutils/meta.py", line 168, in fields
    for name, typ, mode in zip(self.arg_names, self.arg_types, self.arg_modes)
Exception 'NoneType' object is not iterable
```

`fields()` only guards against a missing `arg_modes` (`elif not self.arg_modes: ...`), then zips `arg_names`/`arg_types`/`arg_modes` unconditionally. `arg_modes` being truthy doesn't guarantee `arg_names` is populated too — a function can have modes but no names, in which case `arg_names` stays `None` and the `zip()` raises before it ever gets to filter by mode.

## Fix

Fall back to `None` placeholders for `arg_names`/`arg_types` the same way `args()` (a few lines up in the same class) already falls back for `arg_modes`, so the zip always gets iterables.

## Test plan

- Added `test_function_metadata_fields_with_variadic_and_no_arg_names`, directly reproducing the reported case via `FunctionMetadata`.
- Added `test_function_metadata_fields_table_mode_with_no_arg_names`, checking a TABLE-mode function with unnamed columns returns sensible `ColumnMetadata` (not just an empty list).
- Verified both fail with the exact reported `TypeError` against the unpatched code.
- `pytest tests/ --ignore=tests/features` — 2575 passed, 1 pre-existing failure unrelated to this change (a Windows-only tempfile permission issue in an unrelated alias-map test), 1 xfailed, 1 xpassed.
- `ruff check` / `ruff format` clean.


DiegoDAF:
Thanks for tracking this down. I hit the same crash in my fork a while back, so here is what I measured against your branch, in case it is useful.

The fix removes the `TypeError` in `fields()`, but the crash moves rather than disappears. With your branch applied, completing columns of a set-returning function declared with `TABLE(...)` and no argument names still fails, one layer further up:

```
upstream main   TypeError: 'NoneType' object is not iterable
                pgcli/packages/parseutils/meta.py:162, in fields

this PR         TypeError: 'NoneType' object is not iterable
                pgcli/pgcompleter.py:77, in generate_alias
```

The reason is that `fields()` now returns `ColumnMetadata` entries whose `name` is `None`:

```python
f = FunctionMetadata("s", "f", None, ["int4", "text"], ["t", "t"], "record",
                     False, False, True, False, None)
f.fields()   # [(None, 'int4'), (None, 'text')]
```

and `generate_alias()` iterates that name. The variadic case in your tests returns `[]`, so it never reaches that code and the tests pass.

What worked for me was to treat "no argument carries an output mode, or the names are missing" the same way the branch a few lines above already treats a function declared without output parameters: use the function name as the column name.

```python
fields = [
    ColumnMetadata(name, typ, [])
    for name, typ, mode in zip(self.arg_names or [], self.arg_types or [], self.arg_modes)
    if mode in ("o", "b", "t")
]
return fields or [ColumnMetadata(self.func_name, self.return_type, [])]
```

That gives `[('labels', 'hstore')]` for `labels(variadic text[]) returns hstore`, so `select labels from labels(...)` completes the column instead of offering nothing, and the same probe that crashes above returns suggestions.

Either way the crash is worth fixing, so this is a suggestion on top of your change rather than an objection to it.

Tatamis:
Thanks @DiegoDAF, you were right: with `TABLE(...)` and no argument names the crash just moved to `generate_alias()`. I reproduced it on the branch and pushed 2224ee06 with your suggestion: when no output column can be built from `arg_names`/`arg_modes`, `fields()` falls back to the function name, like the branch above it. The variadic case now returns `[('labels', 'hstore')]`, and I added tests for the unnamed `TABLE` case (checking `generate_alias()` gets a string) and for named `TABLE` columns.

PR #1639, opened 2026-09-17
ci: update the deprecated CodeQL action and checkout

DiegoDAF:
GitHub annotates every CodeQL run on this repository with two notices, visible on any recent run of the workflow:

> CodeQL Action major versions v1 and v2 have been deprecated. Please update all occurrences of the CodeQL Action in your workflow files to v3.

> Node.js 20 is deprecated. The following actions target Node.js 20 but are being forced to run on Node.js 24: `actions/checkout@v3`, `github/codeql-action/analyze@v2`, ...

The first is reported at failure level, so the workflow is already being flagged even though the analysis itself succeeds.

This moves `github/codeql-action` from v2 to v4 and `actions/checkout` from v3 to v5. Nothing else changes: same trigger, same `python` matrix and the same `+security-and-quality` query suite.

I ran the same change on my fork first and the CodeQL workflow comes back green with both annotations gone.


dbaty:
I did the same (and some more) in a separate pull request: https://github.com/dbcli/pgcli/pull/1641.

